PR #3, opened 2017-05-26
Added MD files for iedit, medit, redit

jalles76:
Not sure if these should be in plain text, rather.


luciensadi:
Markdown works for me, thanks for putting these together.

PR #6, opened 2017-06-02
Const char fixes

luciensadi:
I'm pleased to report that there are no more deprecated-const-char warnings displayed.


jalles76:
Cleanest compile of the AwakeMUD sources I've ever seen! Excellent job!

PR #8, opened 2017-06-02
Imported old spec_procs from 0.8.16B.

luciensadi:
This code isn't hooked into anything as far as I know, but it's there for later use if needed.


luciensadi:
This particular pull should not change any existing functionality. I almost feel like I should include the extra code in a text file instead of a compiled file since it's essentially dead code.

jalles76:
I agree. The procs refer to autoruns, rooms or mobs which are not included.

luciensadi:
Shifted code from spec_procs to a text file in the new snippets/ directory. Will merge.

PR #18, opened 2017-06-06
Spellset

luciensadi:


src/constants.cpp
```diff
@@ -1660,6 +1660,17 @@ const char *attributes[] =
     "Reaction"
   };
 
+const char *short_attributes[] =
```
luciensadi: Used to confirm which attribute the imm wants affected by the attribute spells in spellset.

src/interpreter.cpp
```diff
@@ -864,6 +865,7 @@ struct command_info cmd_info[] =
     { "spool"    , POS_DEAD    , do_spool    , 0, 0, FREE }, 
     { "speed"    , POS_RESTING , do_speed    , 0, 0 },
     { "spells"   , POS_SLEEPING, do_spells   , 0, 0 },
+    { "spellset" , POS_SLEEPING, do_spellset , LVL_DEVELOPER, 0 },
```
luciensadi: Should this require higher privs? I don't recall what 'developer' is numerically.

jalles76: skillset uses LVL_DEVELOPER so I think it's fine.

luciensadi:
Will merge this for now and file an issue to improve spellset.

PR #31, opened 2017-06-10
Bugfixes

luciensadi:
- During do_quit it was possible for your character to be save to the DB up to three times in a row. This has been reduced to one time.

- Fixed issue with out-of-order looks when restoring from copyover.


jalles76:
Tested and working. Nice work!

Issue #38, opened 2017-06-13
Modify new mem code to not mess with protos


luciensadi:
If an object was created directly from proto and has its pointers still pointing to the proto obj, don't delete the pointers. Same for all other things too.

luciensadi:
I haven't yet seen any issues from this while playtesting, so this issue is backburnered until we actually start seeing problems.

luciensadi:
Found the issues. This code can't be merged into the main branch until all references to prototype objects are found and made safe.

luciensadi:
Work on this issue is underway in the memory-management-redux branch.

luciensadi:
Also, I'd like to point out how much of a memory hog this game is:

> leaks Report Version:  2.0
> Process 34282: 23115 nodes malloced for 4588 KB

Truly horrifying.

luciensadi:
Closing out the old memory-management branch in favor of -redux. This resolves this issue.

PR #40, opened 2017-06-13
Memory management

luciensadi:
Many, many changes to the way new bits of memory are allocated. This will clean up a lot of memory leaks, but has potential buggy behavior linked to the deletion of prototype fields (issue #38). However, no buggy behavior has been seen yet, so I'm moving this code to the main branch for playtesting.


luciensadi:
Issue #38 must be resolved before this can be merged.

PR #44, opened 2017-06-15
Memory management redux

luciensadi:
I think we're finally ready to merge the memory management fixes. What I'd really like is some comprehensive playtesting-- you don't need to check for leaks in the process, just play around with commands and features and see if you can break anything. I'll merge this if we don't see any crashes in a week.


luciensadi:
Testing in progress, but I need to merge master->redux to do so.

Issue #61, opened 2017-06-19
Add more language choices to chargen


jalles76:
Trainer needs update

PR #73, opened 2017-06-22
Chargen languages

jalles76:
Resolves issue #61. Added all available languages in the game to chargen teacher,

### DIFF
--- a/src/spec_procs.cpp
+++ b/src/spec_procs.cpp
@@ -513,7 +513,7 @@ SPECIAL(teacher)
   for (i = 0; i < NUM_TEACHER_SKILLS; i++)
     if (skill_num == teachers[ind].s[i])
       break;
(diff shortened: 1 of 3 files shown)


src/spec_procs.cpp
```diff
@@ -513,7 +513,7 @@ SPECIAL(teacher)
   for (i = 0; i < NUM_TEACHER_SKILLS; i++)
     if (skill_num == teachers[ind].s[i])
       break;
-  if (i >= 8) {
+  if (i >= 20) {
```
luciensadi: Let's make this i >= NUM_TEACHER_SKILLS to avoid the magic-number nature of it.

luciensadi: Also good job, you fixed a latent bug in the actual game with this 😄

PR #84, opened 2017-06-23
First-pass cleanup of some of the in_room crash issues. Added garage …

luciensadi:
…to imm HQ.

### DIFF
--- a/src/act.comm.cpp
+++ b/src/act.comm.cpp
@@ -48,7 +48,7 @@ ACMD(do_say)
   skip_spaces(&argument);
   if (!*argument)
     send_to_char(ch, "Yes, but WHAT do you want to say?\r\n");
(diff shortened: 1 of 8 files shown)


src/act.comm.cpp
```diff
@@ -124,7 +124,7 @@ ACMD(do_say)
         act(buf, FALSE, ch, NULL, NULL, TO_ROOM);
       } else {
         success = success_test(GET_SKILL(ch, GET_LANGUAGE(ch)), 4);
-        for (tmp = world[ch->in_room].people; tmp; tmp = tmp->next_in_room)
+        for (tmp = world[get_absolute_in_room(ch)].people; tmp; tmp = tmp->next_in_room)
```
luciensadi: This actually makes it so that when you speak inside a vehicle it's heard outside normally. PR denied.

PR #97, opened 2017-07-17
Consolidated wizload and iload functions; they now both invoke perfor…

### LINKED ISSUE #45: Unable to iload credsticks, but you can load them.
I don't know why the iload restriction was put into place (my suspicion is that it's so an imm can't iload a credstick, put an arbitrary amount of nuyen on it, and hand it off to a mort without anyone tracking how much was given), but the restriction is not universal. It can be bypassed with LOAD OBJ <vnum>, or the imm could just buy a credstick from one of the vendors.

If the goal is to make it so imms Cannot Give Out Credsticks Ever, restrict them from doing so in give/drop.

Otherwise, just add logging so that the value of a credstick is displayed in the cheatlog and remove this restriction.

luciensadi:
…m_wizload_obj. This unifies the logic (no more arbitrary restrictions on one but not the other) and gets rid of a bug where an object created by the staff could be treated as a non-staff object. Also, why do we have both IMMLOAD and WIZLOAD flags? (Resolves #45)

### DIFF
--- a/src/act.wizard.cpp
+++ b/src/act.wizard.cpp
@@ -1618,11 +1618,72 @@ ACMD(do_return)
   }
 }
 
+void perform_wizload_object(struct char_data *ch, int vnum) {
+  int real_num, counter, i;
+  bool found = FALSE;
+  struct obj_data *obj = NULL;
+  
+  assert(ch != NULL);
+  
+  // Precondition: Number cannot be negative.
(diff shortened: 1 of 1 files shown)


jalles76:
Could be a Diku/Circle relic. Consolidating the two seems the prudnet thing to do.

PR #165, opened 2018-10-29
Fix for issue 164

### LINKED ISSUE #164: Changing types in IEDIT does not change all object values to 0
Values greater than index 9 are not changed.

luciensadi:
Change from unrolled loop to rolled loop to allow for loop variance. Fixes #164.


luciensadi:
Leaving as a PR until I have time to compile and test to see if this actually works. Currently working in another branch.

luciensadi:
Tested and confirmed functional.

PR #169, opened 2018-10-31
Probe Command

luciensadi:
This PR tracks the progress of the PROBE command. It is not in a mergeable state yet since this feature is still being tested.

Remaining todo item: Write a HELP file for this command.


luciensadi:
Some of the content from this PR is missing. Closing out.

src/interpreter.cpp
```diff
@@ -764,6 +764,7 @@ struct command_info cmd_info[] =
     { "practice" , POS_RESTING , do_practice , 1, 0 },
     { "prance"   , POS_STANDING, do_action   , 0, 0 },
     { "pray"     , POS_SITTING , do_action   , 0, 0 },
+    { "probe"    , POS_RESTING , do_examine  , 0, SCMD_PROBE },
```
lmclarke: Do we need to also change the line for examine to SCMD_EXAMINE?

luciensadi: SCMD_EXAMINE is already defined as 0, so there will be no mechanical difference, but that's a good idea for clarity's sake.

luciensadi: Fixed in b4f56da3cbb45e4d8f059eee2f662464495eb414.

luciensadi:
Known issues at the moment:
- Awake CE does not have a -4 TN for Smartlink IIs, while Awake does. Probe code currently mimics Awake behavior for compatibility reasons.
- Laser sights are not their own attachment type, so they show up as other things depending on what their designer made them as.
- Scopes do not identify their magnification level.
- No attachment calls out the flag that it provides (you must unattach, probe it, and reattach to see this).

src/fight.cpp
```diff
@@ -3191,9 +3191,6 @@ int calculate_vision_penalty(struct char_data *ch, struct char_data *victim) {
   return modifier;
 }
 
-// TODO: Macro this for compiler speed.
-#define IS_DAMTYPE_PHYSICAL(type) !((type) == TYPE_HIT || (type) == TYPE_BLUDGEON || (type) == TYPE_PUNCH || (type) == TYPE_TASER || (type) == TYPE_CRUSH || (type) == TYPE_POUND)
```
luciensadi: Moved to `utils.h`.

src/act.wizard.cpp
```diff
@@ -819,17 +818,6 @@ void do_stat_veh(struct char_data *ch, struct veh_data * k)
   send_to_char(buf, ch);
 }
 
-const char *workshops[] = {
```
luciensadi: Moved to `constants.cpp`.

luciensadi:
Latest push fixes all issues except the -4 TN for SL2. This is acceptable IMO.

luciensadi:
Added help file.

Question: Should the PROBE command be restricted to objects that you possess, or should we be allowing players to probe things like fountains and other random items in the area?

luciensadi:
Characters can no longer PROBE an item that they're not wearing/holding.

luciensadi:
@jalles76 , could I get your thoughts on this feature / PR?

luciensadi:
Master now includes libsodium branch, so imported that code here.

src/act.informative.cpp
```diff
@@ -55,6 +55,8 @@ extern char *colorize(struct descriptor_data *, char *);
 extern int mysql_wrapper(MYSQL *mysql, const char *query);
 extern MYSQL *mysql;
 
+extern int get_weapon_damage_type(struct obj_data* weapon);
```
luciensadi: From `fight.cpp`, reproduced in this comment:

```int get_weapon_damage_type(struct obj_data* weapon) {
  if (!weapon)
    return TYPE_HIT;
  
  int type;
  switch (GET_OBJ_VAL(weapon, 3)) {
    case WEAP_EDGED:
    case WEAP_POLEARM:
      type = TYPE_SLASH;
      break;
    case WEAP_WHIP:
    case WEAP_CLUB:
    case WEAP_GLOVE:
      type = TYPE_POUND;
      break;
    case WEAP_SMG:
    case WEAP_ASSAULT_RIFLE:
    case WEAP_LMG:
    case WEAP_MMG:
    case WEAP_HMG:
    case WEAP_MINIGUN:
      type = TYPE_MACHINE_GUN;
      break;
    case WEAP_SHOTGUN:
      type = TYPE_SHOTGUN;
      break;
    case WEAP_TASER:
      type = TYPE_TASER;
      break;
    case WEAP_CANNON:
      type = TYPE_CANNON;
      break;
    case WEAP_SPORT_RIFLE:
    case WEAP_SNIPER_RIFLE:
      type = TYPE_RIFLE;
      break;
    case WEAP_MISS_LAUNCHER:
    case WEAP_GREN_LAUNCHER:
      type = TYPE_ROCKET;
      break;
    default:
      type = TYPE_PISTOL;
      break;
  }
  
  return type;
}```

luciensadi:
Looks like it's as good as it's going to get. Merging into master.

Issue #189, opened 2018-11-08
IEDIT doesn't display edged weapon type as possibility when editing TYPE_WEAPON


luciensadi:
**Describe the bug**
g-menu does not have edged weapon type, starts at club

luciensadi:
Tentatively fixed per #197.

PR #197, opened 2019-02-10
Issue 189

luciensadi:
Fix for #189. Also includes some magic number removals.


src/iedit.cpp
```diff
@@ -182,10 +182,8 @@ void iedit_disp_skill_menu(struct descriptor_data *d)
 /* weapon type */
 void iedit_disp_weapon_menu(struct descriptor_data * d)
 {
-  int             counter;
-  
   CLS(CH);
-  for (counter = 1; counter < MAX_WEAP; counter += 2)
+  for (int counter = 0; counter < MAX_WEAP; counter += 2)
```
luciensadi: This is the actual fix for #189-- it was just an off-by-one error.

Issue #176, opened 2018-11-01
Player's vehicle won't save if inside a vehicle with no owner


luciensadi:
**Describe the bug**
If you put owned vehicle A inside unowned vehicle B (or if B is owned, but becomes unowned / a candidate for deletion while A is inside it), A will be lost on copyover or crash.

**Expected behavior**
Vehicle A should be moved to the room vehicle B was in, or the parking garage if the room is not a garage.

luciensadi:
This behavior is trickier to deal with than first glance would indicate. Non-owned player vehicles are 'deleted' by skipping their load when reading the vehicle save files. In order to make this change happen, all vehicles must first be loaded, then all vehicles owned by non-existent players must have their vehicular contents shifted to their current location and must be deleted. It is also important to not affect zoneloaded vehicles during this operation.

luciensadi:
Fixed in #203.

PR #203, opened 2019-04-04
WIP: Potential fix for #176. Needs testing before merge.

luciensadi:


luciensadi:
Fix does not currently work. The vehicle usually ends up at Dante's instead of being dropped into the appropriate room. There's also an issue where we're modifying a singly-linked list while reading from it.

luciensadi:
Fix tested. Nested vehicles now work properly when the character who owns them deletes-- they dump their vehicle contents into their container and destroy themselves. They specifically don't drop their items or mods since it can't be proven who owned these items.

PR #214, opened 2019-04-07
Overhauled enhanced articulation and implemented the missing skills.

luciensadi:


luciensadi:
Although it brings the game closer to canon, this also seems like it makes Enhanced Articulation a must-have for any character who's not a zero-physical-combat mage or a pure face character. The previous implementation where it only boosted QUI-linked attributes may have been set up for game balance. Should we go through with this change, or revert it?

luciensadi:
Will try this out for now and see how it works. We can always nerf later.

PR #276, opened 2019-05-29
Added cygwin support

vitalyo7:
This allows the MUD to be compiled and hosted on a windows server. Added support in the Makefile to support CYGWIN config as well as some header defines.

There was an issue where all db files in windows had CRLF line endings. This caused the tags to be read as "\r#1000" and impacted the parsing of the database. Which in turn caused the mud to crash at the start. I have added a fix to File.cpp to strip out the \r character from \r\n which seemed to resolve the issue.

I use cygwin mainly for testing locally, so I will keep testing if there are any issues.


luciensadi:
Looks good, thanks!

PR #438, opened 2019-07-27
Add perfmon to monitor pulse usage and trace command performance.

Odoth:


src/perfmon.cpp
```diff
@@ -0,0 +1,614 @@
...
+#include <vector>
+#include <ctime>
+#include <cfloat>
+#include <cstring>
+
+#include <sys/time.h>
+
+static double const INFINITY = DBL_MAX;
+
+#define PULSE_PER_SECOND 10
```
luciensadi: Should this reference awake.h's PASSES_PER_SEC so it scales if someone changes the pulse frequency later?

src/comm.cpp
```diff
@@ -605,6 +608,24 @@ void game_loop(int mother_desc)
...
+
+      {
+        long int total_usec = 1000000 * timespent.tv_sec + timespent.tv_usec;
+        double usage_pcnt = 100 * ((double)total_usec / OPT_USEC);
+        PERF_log_pulse(usage_pcnt);
+
+        if (usage_pcnt >= 100)
+        {
+          log("Pulse usage >= 100%%. Trace info: ");
+          char buf[MAX_STRING_LENGTH * 12];
```
luciensadi: M_S_L is currently defined as a rather beefy 32768 characters. This makes the buf declared here a 393kb array; is that required?

src/act.wizard.cpp
```diff
@@ -4868,3 +4869,68 @@ ACMD(do_zone) {
...
+            "perfmon summ            - Print summary,\r\n"
+            "perfmon prof            - Print profiling info.\r\n"
+            "perfmon sect <section>  - Print profiling info for section.\r\n",
+            ch );
+        return;
+    }
+
+    if (!str_cmp( arg1, "all"))
+    {
+        char buf[MAX_STRING_LENGTH * 12];
```
luciensadi: Same question as earlier about size of this buffer (and the similar declarations in this file).

src/interpreter.h
```diff
@@ -10,13 +10,24 @@
...
 #define _interpreter_h_
+
+#include "perfmon.h"
 /*
 #define ACMD(name)  \
    void (name)(struct char_data *ch, __attribute__ ((unused)) char *argument, __attribute__ ((unused)) int cmd, __attribute__ ((unused)) int subcmd)
 */
 
-#define ACMD(name)   \
+#define ACMD_DECLARE(name)   \
```
luciensadi: What is the purpose of changing ACMD() to ACMD_DECLARE() and adding instrumentation to a new version of ACMD()? Should all new commands be written with ACMD_DECLARE() now?

luciensadi: I understand the change now-- ACMD_DECLARE() for prototypes, ACMD() for actual commands.

src/perfmon.cpp
Odoth: My design preference has been to have it as a standalone module with no direct dependencies on the game code. Could make this configurable from outside instead of a define if you think that's worthwhile.

src/comm.cpp
Odoth: MSL * 12 was basically the max page string length from Aarchon. Will just use MSL here.

src/perfmon.cpp
luciensadi: This is fine for now, I'll revisit later if we consider changing the pass value.

PR #471, opened 2020-01-10
Fixes for incorrect encoding.

sticks2681:


luciensadi:
Also, it looks like a lot of the changes here were permission changes (644->755), is that accurate?

sticks2681:
I don't exactly know what happened there. I reverted that second commit that had deleted dos2unix.c and changed a lot of other files. The fixes for incorrect encoding should only touch the obj files 850, 852, 854, 856, 859.

PR #496, opened 2020-12-09
Bullet pants.

luciensadi:
The new ammo system design. This code is intended to implement the following system:

1) All rooms have a gun debris counter, which works like the blood system. Firing a weapon increments it, as does reloading a weapon. Starts at 'A few spent casings are scattered around here.' and ends with 'It looks like WWIII was fought here.'

2) All characters (PCs and NPCs) have abstract bullet storage that tracks ammo counters for each type of weapon and each type of ammunition. This storage is referred to as 'bullet pants' or 'BP' in this document. No clothes are required for bullet pants; your character is free to keister assault cannon rounds and otherwise '[PAIGE NO](https://i.imgur.com/1Loph0O.jpg)' itself. As part of this change, magazines are removed as objects, and exist solely as flavor text.

3) When reloading a weapon, the ammo is drawn directly from the bullet pants. For PCs, it uses the last specified type of ammo, and fails to reload if that specific ammotype is expended (for example, this lets you auto-stop shooting wage slaves when you run out of normal ammo).

4) NPC ammo counts are specified during editing/creation. Their ammo counts reset along with the zone. Their ammo cannot be looted. They only get one ammo type per weapon, because I don't want to write the logic for them to choose between ammo types.

5) PCs buy ammo in boxes, and can load their bullet pants from these boxes. Ammo still has weight, so there is a max carry limit. Ammo can be taken out of bullet pants, this puts them into new boxes that can then be combined with other boxes if desired.

6) Weapon ammotype for reloads is whatever is currently loaded into it. If no ammotype is specified (empty gun / first load) and you reload, it'll select Normal.

7) Staff can set character ammo counts with `set <target> pockets <weapon> <ammotype> <quantity>`.

Syntax for system:

`pocket[s]`: displays your current bullet pants contents
`pocket[s] put|store <quantity> <weapon> <type>`: add ammo from boxes to your bullet pants
`pocket[s] get|remove|retrieve <quantity> <weapon> <type>`: create a new ammo box from your bullet pants with the specified amount in it

`reload <weapon> [ammotype]`: reloads the selected weapon with type-X rounds
`reload <ammotype>`: changes your current weapon to type-X rounds
`reload`: reloads your weapon with the same type of rounds currently in it, fails if you don't have any more of that left


luciensadi:
Merging. It's not yet completed (see bullet_pants.cpp for list of todos at bottom of file), but it's testable.

PR #498, opened 2020-12-31
Menehune Points Chargen

Carejo:
Moves Menehune from 15 points in Chargen to 10 points, in the points Chargen system.

### DIFF
--- a/src/chargen.cpp
+++ b/src/chargen.cpp
@@ -805,13 +805,13 @@ void create_parse(struct descriptor_data *d, const char *arg)
           case RACE_KOBOROKURU:
           case RACE_TROLL:
           case RACE_ELF:
+          case RACE_MENEHUNE:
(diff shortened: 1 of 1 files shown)


UndeadNappist:
koborokuru should likely be moved to 5 unless there's some reason to have a race that is literally identical to baseline be 10. Do they have low-light or something?

Carejo:
> koborokuru should likely be moved to 5 unless there's some reason to have a race that is literally identical to baseline be 10. Do they have low-light or something?

From what I can find, they don't have a movement penalty like other dwarves (which I doubt is reflected in mechanics) and some documentation says they have extra die to soak edge weapons (also highly doubtful it's IG).

There seems to be a trend with dwarves and orks, that the base meta is one step down from all the submetas, though. So I'd much rather see a buff to Kobos then a point drop.

luciensadi:
Thanks!

PR #504, opened 2021-01-06
changes training/prac pricing

UndeadNappist:
Makes skills cheaper and attributes more expensive.

There are more skills in the game and a plethora of NERP skills, whereas you only have 6 attributes, and you can only raise them a limited amount.

Similarly, having higher attributes changes the price of skills.

### DIFF
--- a/src/spec_procs.cpp
+++ b/src/spec_procs.cpp
@@ -562,7 +562,7 @@ SPECIAL(nerp_skills_teacher) {
     return TRUE;
   }
   if (!PLR_FLAGGED(ch, PLR_NOT_YET_AUTHED)) {
-    int skill_nuyen_cost = MAX(1000, (GET_SKILL(ch, skill_num) * 5000));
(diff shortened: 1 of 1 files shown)


luciensadi:
What's the rationale for making this change?

UndeadNappist:
Originally I presume the low attribute cost was not to completely screw newbies, since a fast way to "fix" your character would be to quickly raise your stats. With the new archtype system and better explanations, this shouldn't be as much of a problem.

NERP skills were also a good sinkhole for costs and likely easier to change than going through and altering every piece of equipment in the game; so maybe this wasn't the best choice.

But it's mostly that there are more skills than attributes, every attribute does SOMETHING, whereas there are a lot of skills that do nothing which people would still want to buy.

UndeadNappist:
At any rate, it occurs to me that changing costs for things should be done before people start farming and such, and all the costs should be overpriced to avoid feeling like new players are being punished. Screwing around with the economy can easily cause hurt feelings.

luciensadi:
The current system seems to be playing out alright in the game. Thanks for the suggestion though!

PR #505, opened 2021-01-06
Adds mouth

UndeadNappist:
Damned whitespacing package.

Supposedly adds the mouth slot.

### DIFF
--- a/src/act.obj.cpp
+++ b/src/act.obj.cpp
@@ -2403,6 +2403,7 @@ void perform_wear(struct char_data * ch, struct obj_data * obj, int where)
                               "YOU SHOULD NEVER SEE THIS MESSAGE.  PLEASE REPORT.\r\n",
                               "You already have something on each of your ankles.\r\n",
                               "You are already wearing something on your feet.\r\n",
+                              "You're already wearing something on your feet.\r\n", // this message is repeated, not sure why
                               "You already have something in your mouth.\r\n",
                             };
 
--- a/src/act.obj.cpp
+++ b/src/act.obj.cpp
@@ -45,7 +45,7 @@ void calc_weight(struct char_data *ch);
 SPECIAL(weapon_dominator);
 
 int wear_bitvectors[] = {
-                          ITEM_WEAR_TAKE, ITEM_WEAR_HEAD, ITEM_WEAR_EYES, ITEM_WEAR_EAR,
+                          ITEM_WEAR_TAKE, ITEM_WEAR_HEAD, ITEM_WEAR_EYES, ITEM_WEAR_EAR, ITEM_WEAR_MOUTH,
                           ITEM_WEAR_EAR, ITEM_WEAR_FACE, ITEM_WEAR_NECK, ITEM_WEAR_NECK,
                           ITEM_WEAR_BACK, ITEM_WEAR_ABOUT, ITEM_WEAR_BODY, ITEM_WEAR_UNDER,
                           ITEM_WEAR_ARMS, ITEM_WEAR_ARM, ITEM_WEAR_ARM, ITEM_WEAR_WRIST,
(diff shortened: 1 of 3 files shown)


luciensadi:
This won't work as written. The new mouth slot must be at the end of all of the arrays-- end of wear_bits, end of the array in constants.cpp, etc. Remember that the code will look values up in the array by the values you provided in awake.h, so it'll return the 40th string in the list for wear location, etc.

UndeadNappist:
Ah, thanks. Figures I was doing something wrong.

src/awake.h
```diff
@@ -1122,6 +1123,7 @@ enum {
 #define ITEM_WEAR_FACE   23
 #define ITEM_WEAR_THIGH  24
 #define ITEM_WEAR_MAX    25
```
luciensadi, line 1125: X_MAX is used by the code to indicate 'the acceptable values for this are 0 up to this number', so having the mouth flag after wear_max means it won't be accounted for in a lot of the code. Swap these so mouth is 25 and max is 26.

src/act.obj.cpp
```diff
@@ -45,7 +45,7 @@ void calc_weight(struct char_data *ch);
 SPECIAL(weapon_dominator);
 
 int wear_bitvectors[] = {
-                          ITEM_WEAR_TAKE, ITEM_WEAR_HEAD, ITEM_WEAR_EYES, ITEM_WEAR_EAR,
+                          ITEM_WEAR_TAKE, ITEM_WEAR_HEAD, ITEM_WEAR_EYES, ITEM_WEAR_EAR, ITEM_WEAR_MOUTH,
```
luciensadi, line 48: This new value needs to be at the end of the wear_bitvectors[] array, after wear_feet. This ensures that we don't mess up any existing wear locations, and also that your string for wearing the item (which you set below) is in the right spot.

src/act.obj.cpp
```diff
@@ -2400,7 +2403,8 @@ void perform_wear(struct char_data * ch, struct obj_data * obj, int where)
                               "YOU SHOULD NEVER SEE THIS MESSAGE.  PLEASE REPORT.\r\n",
                               "You already have something on each of your ankles.\r\n",
                               "You are already wearing something on your feet.\r\n",
-                              "You're already wearing something on your feet.\r\n"
+                              "You're already wearing something on your feet.\r\n", // this message is repeated, not sure why
```
luciensadi, line 2406: Socks and shoes.

src/awake.h
```diff
@@ -263,8 +263,9 @@ enum {
 #define WEAR_RANKLE    35
 #define WEAR_SOCK      36
 #define WEAR_FEET      37
-#define WEAR_PATCH     38      
+#define WEAR_PATCH     38
 #define NUM_WEARS      39
```
luciensadi, line 267: Same comment as below-- NUM_WEARS must be last, so swap this with WEAR_MOUTH.

src/act.obj.cpp
UndeadNappist: I think I did it right this time?

src/act.obj.cpp
```diff
@@ -2400,7 +2404,7 @@ void perform_wear(struct char_data * ch, struct obj_data * obj, int where)
                               "YOU SHOULD NEVER SEE THIS MESSAGE.  PLEASE REPORT.\r\n",
                               "You already have something on each of your ankles.\r\n",
                               "You are already wearing something on your feet.\r\n",
-                              "You're already wearing something on your feet.\r\n"
+                              "You already have something in your mouth.\r\n",
```
luciensadi, line 2407: You've replaced the wear-feet message with the mouth one, so people with shoes on who try to wear more shoes will wonder why they're trying to eat their shoes. Recommend restoring `"You're already wearing something on your feet.\r\n",` before this.

luciensadi: Other than this, it looks good!

UndeadNappist: Fixed the dangling comma while I was at it.

luciensadi:
Thanks!

PR #515, opened 2021-02-07
Vile's Mildly Drunk Cyberlimbs

TheVilest:
Cyberlimbs? I may have gotten drunk halfway through, which means i'm doing something right.


luciensadi:
This is missing changes to awake.h that would define each cyberlimb / mod type and would also define the NUM_X values. Ex:

```
#define CYBERLIMB_MOD_RAILGUN              0
#define CYBERLIMB_MOD_SHEEP_SHEARER 1
#define NUM_CYBERLIMB_MODS                   2
```

src/db.cpp
```diff
@@ -5542,22 +5542,180 @@ void price_cyber(struct obj_data *obj)
       GET_OBJ_AVAILDAY(obj) = 2;
       break;
     case CYB_SKULL:
-      GET_OBJ_VAL(obj, 1) = 0;
-      GET_CYBERWARE_ESSENCE_COST(obj) = 75;
-      if (GET_OBJ_VAL(obj, 3))
+      GET_OBJ_AVAILDAY(obj) = GET_OBJ_AVAILTN(obj) = GET_CYBERWARE_ESSENCE_COST(obj) = GET_OBJ_COST(obj) = 0;
+      if (IS_SET(GET_OBJ_VAL(obj, 3), SKULL_OBVIOUS)) {
```
luciensadi, line 5546: This needs to match the awake.h name, so `SKULL_MOD_OBVIOUS`. Same for other items here.

src/db.cpp
```diff
@@ -5542,22 +5542,180 @@ void price_cyber(struct obj_data *obj)
...
+        GET_OBJ_AVAILTN(obj) = MAX(GET_OBJ_AVAILTN(obj), 8);
+        GET_OBJ_AVAILDAY(obj) = MAX(GET_OBJ_AVAILDAY(obj), 14);
+        GET_CYBERWARE_ESSENCE_COST(obj) += 0;
       break;
     case CYB_TORSO:
-      GET_OBJ_VAL(obj, 1) = 0;
-      GET_CYBERWARE_ESSENCE_COST(obj) = 150;
-      if (GET_OBJ_VAL(obj, 3))
+      GET_OBJ_AVAILDAY(obj) = GET_OBJ_AVAILTN(obj) = GET_CYBERWARE_ESSENCE_COST(obj) = GET_OBJ_COST(obj) = 0;
+      if (IS_SET(GET_OBJ_VAL(obj, 3), TORSO_OBVIOUS)) {
```
luciensadi, line 5566: If you think you might add more bits later, you can save a bit here by having it be `if (!IS_SET(... TORSO_SYNTHETIC)`. Since the cybermod is always either obvious or synthetic, the negation of one means the other is true.

src/db.cpp
```diff
@@ -5542,22 +5542,180 @@ void price_cyber(struct obj_data *obj)
...
+        GET_OBJ_COST(obj) += 6500;
+        GET_OBJ_AVAILTN(obj) = MAX(GET_OBJ_AVAILTN(obj), 8);
+        GET_OBJ_AVAILDAY(obj) = MAX(GET_OBJ_AVAILDAY(obj), 14);
+        GET_CYBERWARE_ESSENCE_COST(obj) += 0;
+      }
+      if (IS_SET(GET_OBJ_VAL(obj, 3), TORSO_ARMORMOD2)) {
+        GET_OBJ_COST(obj) += 13000;
+        GET_OBJ_AVAILTN(obj) = MAX(GET_OBJ_AVAILTN(obj), 8);
+        GET_OBJ_AVAILDAY(obj) = MAX(GET_OBJ_AVAILDAY(obj), 14);
+        GET_CYBERWARE_ESSENCE_COST(obj) += 0;
```
luciensadi, line 5588: If none of these mods increase the essence cost, then I would remove this, as it's a non-operation. If you're trying to make it clear that it doesn't increase the armor cost, just add a comment to that effect.

src/db.cpp
luciensadi: This also works neatly around the edge case of a builder forgetting to set either one of these, making the item extremely cheap and zero essence.

src/iedit.cpp
```diff
@@ -213,6 +213,42 @@ void iedit_disp_cybereyes_menu(struct descriptor_data *d)
   send_to_char(CH, "Set Options: ^c%s^n\r\nEnter options (0 to quit): ", buf1);
 }
 
+void iedit_disp_cyberarms_menu(struct descriptor_data *d)
+{
+  CLS(CH);
+  for (int y = 0; y < NUM_ARMSMODS; y += 2)
```
luciensadi, line 219: Value name needs to match awake.h, I think it was `NUM_ARM_MODS` or something?

src/iedit.cpp
```diff
@@ -2370,6 +2414,26 @@ void iedit_parse(struct descriptor_data * d, const char *arg)
             case CYB_REFLEXTRIGGER:
             case CYB_SKULL:
             case CYB_TORSO:
+            case CYB_STRENGTHMOD:
+            case CYB_QUICKNESSMOD:
+            case CYB_ARMORMOD:
+            case CYB_CYBERARMGYROMOUNT:
```
luciensadi, line 2420: Does this make sense? This implies that there is cyberware of the type 'strengthmod' that can be installed, and that it should use the cyberarms table (same for qui, armo, gyro). Likely need to remove these.

src/iedit.cpp
```diff
@@ -213,6 +213,42 @@ void iedit_disp_cybereyes_menu(struct descriptor_data *d)
   send_to_char(CH, "Set Options: ^c%s^n\r\nEnter options (0 to quit): ", buf1);
 }
 
+void iedit_disp_cyberarms_menu(struct descriptor_data *d)
+{
+  CLS(CH);
+  for (int y = 0; y < NUM_ARMSMODS; y += 2)
+    send_to_char(CH, "%2d) %-20s %2d) %-20s\r\n", y+1, armsmods[y], y+2, y+1 < NUM_ARMSMODS ? armsmods[y+1] : "");
+  sprintbit(GET_OBJ_VAL(OBJ, 3), armsmods, buf1, sizeof(buf1));
```
luciensadi, line 221: This is going to add more magic numbers to the code, with magic numbers being defined as 'numbers that aren't explained, but have an effect on the code if changed'. Recommend swapping out the `GET_OBJ_VAL(OBJ, 3)` calls with something from utils.h like `GET_CYBERWARE_FLAGS(OBJ)`.

src/iedit.cpp
```diff
@@ -2370,6 +2414,26 @@ void iedit_parse(struct descriptor_data * d, const char *arg)
...
             case CYB_SKULL:
             case CYB_TORSO:
+            case CYB_STRENGTHMOD:
+            case CYB_QUICKNESSMOD:
+            case CYB_ARMORMOD:
+            case CYB_CYBERARMGYROMOUNT:
+            case CYB_CYBERARMS:
+              if (number < 0 || number > NUM_ARMSMODS+1) {
+                send_to_char("Invalid Input! Enter options (0 to quit): ", CH);
+                return;
```
luciensadi, line 2424: Recommend using the logic below for CYB_HANDRAZOR so you set the bits like you're intending to do. Currently, this will take the exact value put in and use it for the value instead of setting the bit like you want it to.

src/utils.cpp
```diff
@@ -1610,6 +1610,42 @@ bool biocyber_compatibility(struct obj_data *obj1, struct obj_data *obj2, struct
           return FALSE;
         }
         break;
+      case CYB_CYBERARMS:
+        if (GET_OBJ_VAL(bio1, 0) == BIO_MUSCLEAUG || GET_OBJ_VAL(bio1, 0) == BIO_MUSCLETONER
+         || GET_OBJ_VAL(bio1, 0) == BIO_ORTHOSKIN || GET_OBJ_VAL(cyber2, 0) == CYB_DERMALPLATING
+         || GET_OBJ_VAL(cyber2, 0) == CYB_DERMALSHEATHING || GET_OBJ_VAL(cyber2, 0) == CYB_BONELACING
+         || GET_OBJ_VAL(cyber2, 0) == CYB_CYBERARMS) {
+          send_to_char("Cyber Arms are incompatible with these augmentations.\r\n", ch);
```
luciensadi, line 1618: As much as it's gonna suck, you should write out the augmentations that it's incompatible with, if only to save yourself the next decade of `newbie which augmentations are incompatible with cyber arms?`

luciensadi: also, same thing with magic numbers, use `GET_CYBERWARE_TYPE()`. There's no GET_BIOWARE_TYPE yet, but you can add it in utils.h (search for the cyberware one then stick it in the ITEM_BIOWARE section above it).

src/db.cpp
TheVilest: Matched em all up (I think)

src/db.cpp
TheVilest: Yeh, still trying to figure out how to add the other bits to expand upon it, just wanted to nail down the basic foundation to start working off of. Added if (!IS_SET to all of the obvious/synthetic options

src/db.cpp
TheVilest: Some of the extra stuff I wanted to potentially add down the line will cost an essence increase, I just included it to be safe, I can remove em for any that don't require additional essence costs though.

src/iedit.cpp
TheVilest: Should be matched up now

src/iedit.cpp
TheVilest: Yeh, think this was an earlier attempt I forgot to remove (I got a little drunk halfway through, as the title of this push implies) before I realized they were gonna be added as mods instead of standalone cyberware pieces. Went ahead and removed em.

src/utils.cpp
TheVilest: The builder in me saw the obvious solution being a helpfile explaining the incompatibilities.... But that would require people to read. This is an uncertainty, at best, i'm aware.

src/constants.cpp
```diff
@@ -2215,15 +2261,17 @@ const char *decap_cyber_types[] = {
...
   "hydraulic jack",
   "move by wire",
   "climbing claws",
   "smartlink",
-  "muscle replacement"
+  "muscle replacement",
+  "cyberarms modifications",
+  "cyberlegs modifications",
+  "cyberskull modifications",
+  "cybertorso modifications"
```
luciensadi, line 2274: If cyberarm gyromount is a completely separate piece of 'ware that's not a mod, then it should be included here; otherwise it should be removed from the above cyber_types table.

src/iedit.cpp
```diff
@@ -2368,8 +2412,78 @@ void iedit_parse(struct descriptor_data * d, const char *arg)
...
+              if (number < 0 || number > 2) {
+                send_to_char("Invalid Input! Enter options (0 to quit): ", CH);
+                return;
+              }
+              if (number == 0) {
+                iedit_disp_menu(d);
+                return;
+              }
+              number--;
+              TOGGLE_BIT(GET_CYBERWARE_FLAGS(OBJ), 10 << number);
```
luciensadi, line 2425: `10 << number` should be `1 << number` unless you're positive you want the input values to be 20 and 40

src/iedit.cpp
```diff
@@ -2368,8 +2412,78 @@ void iedit_parse(struct descriptor_data * d, const char *arg)
             case CYB_HANDBLADE:
             case CYB_HANDSPUR:
             case CYB_REFLEXTRIGGER:
-            case CYB_SKULL:
-            case CYB_TORSO:
+            case CYB_CYBERARMS:
+              if (number < 0 || number > 2) {
```
luciensadi, line 2416: It looks like there are 10 mods available, recommend `if (number < 0 || number > NUM_ARMS_MODS)` so that this is flexible in case future mods are added

luciensadi: The same change should be made to the other inputs as well-- we want to remove magic numbers (numbers without explanation) wherever possible for better code readability.

src/newmagic.cpp
```diff
@@ -1803,7 +1803,7 @@ void cast_manipulation_spell(struct char_data *ch, int spell, int force, char *a
       send_to_char("You easily dodge it!\r\n", vict);
       act("$n dodges out of the way of a cloud of steam.", TRUE, vict, 0, ch, TO_NOTVICT);
     } else {
-      success -= success_test(GET_BOD(vict) + GET_BODY(vict), force - GET_IMPACT(vict));
+      success -= success_test(GET_BOD(vict) + GET_BODY(vict), force - (GET_IMPACT(vict) / 2));
```
luciensadi, line 1806: What's the source for this?

src/utils.cpp
```diff
@@ -1610,6 +1610,174 @@ bool biocyber_compatibility(struct obj_data *obj1, struct obj_data *obj2, struct
           return FALSE;
         }
         break;
+      case CYB_CYBERARMS:
```
luciensadi, line 1613: This won't work as written. Think of a switch statement like a chain of if-elseif-else. What you have is:

```
if the type is cyberarms, then do this
otherwise, if the type is cyberarms, then do this
otherwise...```

Instead, remove all the extraneous CYB_CYBERARMS lines and breaks, and just stick all your if statements together.

File: src/constants.cpp (hunk 9 above)
Comment (TheVilest) [reply src/constants.cpp]: Completely seperate (and hopefully eventually functional) ware

File: src/iedit.cpp (hunk 10 above)
Comment (TheVilest) [reply src/iedit.cpp:2425]: Whoops, definitely should be a 1

File: src/iedit.cpp (hunk 11 above)
Comment (TheVilest) [reply src/iedit.cpp:2416]: Good call, changed the reference for all of em to streamline any additional references down the line

File: src/newmagic.cpp (hunk 12 above)
Comment (TheVilest) [reply src/newmagic.cpp:1806]: 3rd Ed Core, page 196: "Impact Armor protects against damage from Elemental Manipulations, but at only half its normal rating (round down)."

File: src/utils.cpp (hunk 13 above)
Comment (TheVilest) [reply src/utils.cpp:1613]: Hopefully this is what you meant

File: src/newmagic.cpp (hunk 12 above)
Comment (TheVilest) [reply src/newmagic.cpp:1806]: Lightning Bolt, Acid Stream, Flamethrower, Laser, Steam, Thunderbolt and Waterbolt should all halve impact armour. Clout doesn't, because it's a Telekinetic Manipulation, not Elemental Manipulation

File: src/constants.cpp [hunk 14]
```diff
@@ -2104,6 +2104,49 @@ const char *eyemods[] = {
...
+  "Obvious",
+  "Synthetic",
+  "Armor Mod1",
+  "Strength Mod1",
+  "Strength Mod2",
+  "Strength Mod3",
+  "Quickness Mod1",
+  "Quickness Mod2",
+  "Quickness Mod3",
+  "Cyberarm Gyromount"
```
luciensadi, line 2117: Missing terminal comma means this will be merged with the next line item-- but you also don't need the \n at the end since the code you used doesn't look for that. Your call as to whether you add a comma here, or delete the \n in the rest and remove the commas.

src/utils.cpp
luciensadi: Yep! However, it won't compile since there are a bunch of opening braces without closing braces. You gotta either take off the ending braces of `if (x) {` or add closing braces to the lines that use them, like `} if (x) {`. I personally prefer the latter style, but either way is fine.

luciensadi (commented):
Reviewing this, I'm not sure where these limbs actually have an effect-- ex: gyro limb should be included in newfight's gyro detection code, strength-granting bits should be in... not sure actually, that might just need to be an affect on the item.

TheVilest:
> Reviewing this, I'm not sure where these limbs actually have an effect-- ex: gyro limb should be included in newfight's gyro detection code, strength-granting bits should be in... not sure actually, that might just need to be an affect on the item.

Still tryin to figure out the gyro part, as far as stat bits go I was just gonna handle those through affects, the mods serve as references for the associated values

luciensadi:
How do you want to handle this PR? If the code's not fully implemented yet (ex: gyros), I'm not sure it makes sense for me to review and give feedback. I can help you with the design if you want that?

src/fight.cpp
```diff
@@ -2895,6 +2895,18 @@ int check_recoil(struct char_data *ch, struct obj_data *gun)
...
+      case WEAP_HEAVY_PISTOL:
+      case WEAP_MACHINE_PISTOL:
+      case WEAP_SMG:
+      case WEAP_LMG:
+      case WEAP_SPORT_RIFLE:
+      case WEAP_SNIPER_RIFLE:
+      case WEAP_SHOTGUN:
+      case WEAP_ASSAULT_RIFLE:
+    if (GET_OBJ_VAL(obj, 0) == ARMS_MOD_GYROMOUNT && !GET_OBJ_VAL(obj, 9))
+      comp+= 3;
```
TheVilest, line 2909: So, cyberarm gyromount gives 3 points of recoil comp on every firearm up to LMG, but doesn't give it to MMG, HMG and Assault Cannons. I have a feeling this is a terrible attempt to reflect that fact or even functions remotely as intended.

src/constants.cpp
```diff
@@ -2165,15 +2208,18 @@ const char *cyber_types[] = {
   "Balance Tail",
   "Body Compartment",
   "CyberFin",
-  "CyberSkull",
-  "CyberTorso",
```
luciensadi, line 2169: Removing these line items actually breaks all cyberware. Remember that the key for this table is in awake.h, and removing things here does not remove it there. The effect of this change is to rename everything to what it is two spaces down-- cyber skulls will now appear as dermal sheathing, torsos as foot anchors, etc. I'll fix it on my side.

PR #531, opened 2021-07-16
Proposed magic changes, plus some random commands

luciensadi:
Attempting to bring magic more in line with the expected TNs etc... wrote this late at night, so I need a fresh pair of eyes on it to make sure it all makes sense.

### DIFF
--- a/src/newmagic.cpp
+++ b/src/newmagic.cpp
@@ -2954,10 +2954,10 @@ ACMD(do_conjure)
         break;
       }
       
-    // Modify the TN for wound penalties.
-    target += damage_modifier(ch, buf3, sizeof(buf3));
-    snprintf(buf2, sizeof(buf2), "Wound modifiers: %s", buf3);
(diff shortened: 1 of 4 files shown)


src/newmagic.cpp
```diff
@@ -1164,20 +1167,28 @@ void cast_health_spell(struct char_data *ch, int spell, int sub, int force, char
...
+        base_target = drug_types[GET_DRUG_AFFECT(vict)].power;
+      if (!base_target) {
         send_to_char("They aren't affected by any drugs.\r\n", ch);
         return;
       }
-      success = success_test(skill, target);
+      success = success_test(skill, base_target + target_modifiers);
       if (success > 0 && !AFF_FLAGGED(vict, AFF_DETOX)) {
         create_sustained(ch, vict, spell, force, 0, success, spells[SPELL_STABILIZE].draindamage);
         send_to_char("You notice the affects of the drugs suddenly wear off.\r\n", vict);
```
tuxkamen, line 1194: affects -> effects

PR #559, opened 2021-08-07
Update constants.cpp

UndeadNappist:
Switches racial modifiers for gnomes from +1 qui to +1 str. Maximums are still correct.


luciensadi:
Is the current stat set wrong?

UndeadNappist:
Yes. Gnomes should be +1 body, +1 str, and +2 willpower. Their racial maximums are set correctly, but it's giving them +1 quickness instead of +1 str at character creation.

PR #560, opened 2021-08-08
Rigger messaging tweaks

rhicora:
- Clarify messaging on rigging while working
- Improves messaging for trying drone type upgrades on a regular vehicle
- Improves messaging for trying standard upgrades on a drone.
- Improves take wheel message to make it clear you are taking control instead of already in control.

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -293,7 +293,10 @@ ACMD(do_rig)
     send_to_char(ch, "Try removing your helmet first.\r\n");
     return;
   }
-
+  if (ch->in_veh) {
+    send_to_char(ch, "You can't control a vehicle from inside one.\r\n");
+    return;
+  }
   if (GET_SKILL(ch, SKILL_PILOT_CAR) == 0 && GET_SKILL(ch, SKILL_PILOT_BIKE) == 0 &&
       GET_SKILL(ch, SKILL_PILOT_TRUCK) == 0) {
     send_to_char("You have no idea how to do that.\r\n", ch);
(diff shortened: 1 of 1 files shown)


src/act.drive.cpp
```diff
@@ -802,7 +805,7 @@ ACMD(do_control)
     else if (GET_OBJ_VAL(cyber, 0) == CYB_DATAJACK || (GET_OBJ_VAL(cyber, 0) == CYB_EYES && IS_SET(GET_OBJ_VAL(cyber, 3), EYE_DATAJACK)))
       jack = cyber;
 
-  if (IS_WORKING(ch) || AFF_FLAGGED(ch, AFF_PILOT)) {
```
luciensadi, line 805: IS_WORKING needs to be preserved somewhere, otherwise people will be building ammo while they rig. If the message didn't mesh for you, it can have its own message.

src/act.drive.cpp
```diff
@@ -821,10 +824,6 @@ ACMD(do_control)
     return;
   }
 
-  if (ch->in_veh) {
```
luciensadi, line 824: This needs to be restored unless you've tested it exhaustively, it's a safety mechanism to prevent crashes. The code has limited support for actions taken inside vehicles, and riggers have their bodies referenced for about 50% of the actions they take for some reason.

luciensadi (changes requested):

src/act.drive.cpp
rhicora: Good call, it should be a custom message but we can do that!

src/act.drive.cpp
rhicora: Gotcha, worth testing first. I think it makes sense for riggers to be able to work out of vehicles they aren't driving, but I'll set up a test environment and make sure it is functional before changing this part.

luciensadi (approved):

PR #563, opened 2021-08-13
Update config.h

UndeadNappist:
Drops minimums to all be the same.

Deckers are breaking the economy. Something's gotta give.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -39,10 +39,10 @@ extern const char *CHARACTER_DELETED_NAME_FOR_SQL;
 #define HOST_SECURITY_BLACK_MARKET_MAXIMUM              10000
 
 #define HOST_SECURITY_BLUE_MARKET_MINIMUM               500
-#define HOST_SECURITY_GREEN_MARKET_MINIMUM              1000
-#define HOST_SECURITY_ORANGE_MARKET_MINIMUM             1500
-#define HOST_SECURITY_RED_MARKET_MINIMUM                2000
-#define HOST_SECURITY_BLACK_MARKET_MINIMUM              2500
+#define HOST_SECURITY_GREEN_MARKET_MINIMUM              500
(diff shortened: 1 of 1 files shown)


tuxkamen (commented):
I don't think this is the right short-term approach.  (Long-term fixes involve making the Matrix more persistent and aware of its inhabitants' actions so that kamikaze runs aren't possible.)

The short-term problem seems to me to be that people are cycling through every mall (b/g) at low levels or every Seattle corp (o/r) at higher levels (of course there's no such thing as black or UV, that's an old wives' tale...), and the 30x speed of the mud makes it so that:

1) All sites recover their data about once per GAME day, which is too easy to farm;
2) The paydata market regenerates every GAME hour at an average of 125/tick/color, which means that all colors regain (again, on average) 3000 nuyen per game day. 

Dropping the minimums on b/g will not probably not effectively solve this problem, though it will have slight and growing impact on o/r/(mythical color) hosts.  I can say that I rarely, if ever, notice any hit taken to the market when I sell my (mostly r/mythical) data.

Short-term solutions that would slow nuyen gain more directly would involve:
- Slowing down the rate of payrate recovery and paydata generation (this is the big one)
- Dropping the *maximums* below their previous levels
- Code for 'saturation'--that is, have the fence refuse to buy after a while
- Adding consequences for data-scumming - spawn NPC hunters, maybe?

Medium-term solutions involve making a Matrix run more meaningful instead of having a complete reset of the world's security posture whenever a logoff occurs:
- Attach tallies to grids and hosts and persist them while in a host
- Host tallies and alert statuses decrease following game rules
- Rearchitect hosts to have fewer LTG/PLTG node types (for reasons)

UndeadNappist:
No longer relevant, and the time-sensitive thing to do would have been to just slash all decking maximums several months ago. Too late now.

PR #567, opened 2021-09-03
fix #566: Backtick rank fields in SQL statements.

### LINKED ISSUE #566: MySQL 8.x field support for 'rank'.
**Describe the bug**
In order to support the use of MySQL 8.x, SQL statements make in CE need to backtick any references to 'rank' fields.  Ideally everything should be backticked, but there aren't too many 'rank' references, so...

**To Reproduce**
Steps to reproduce the behavior (example below):
1. Try to run the game using MySQL 8.x.  SQL generation and load will work, but creating characters will throw SQL errors when doing any insert or update that involves rank.

**Expected behavior**
No SQL errors when performing DB events involving rank.

**Current Output**
N/A (typical syntax errors from MySQL about an issue near the word Rank).

tuxkamen:
NOTE: The total chargen.cpp is due to dos2unix (chargen.cpp is in DOS format for some reason).  A diff -w will show that the rank field is the only concern, but if there is a reason for the format being like that, I'll change it back and reapply.


tuxkamen:
Oof.  Fixing the whitespace further is a real drag considering the other changes.  I'm going to put it back to the original format and repush.

PR #568, opened 2021-09-04
New ammo messages

UndeadNappist:
Attempts to add funny ammo failure messages.

Pretty sure I screwed up line 602.

### DIFF
--- a/src/pro_create.cpp
+++ b/src/pro_create.cpp
@@ -599,37 +599,37 @@ void update_buildrepair(void)
         STOP_WORKING(CH);
       } else if (AFF_FLAGGED(CH, AFF_AMMOBUILD) && --GET_AMMOBOX_TIME_TO_COMPLETION(PROG) < 1) {
         if (GET_AMMOBOX_TIME_TO_COMPLETION(PROG) <= -2) // --(-1) = -2; prevents penalizing people who ace the test.
-          (i = rand(1,9)) {
+          (i = number(1,16)) {
             switch(i) {
               case 1:
-                send_to_char("You seem to have messed up the batch of ammo.\r\n", CH);
-                break;
-              case 2:
                 send_to_char("You swear loudly as a gunshot rings out because you accidentally fired the round.\r\n", CH);
                 break;
+              case 2:
+                send_to_char("The poor-quality metal shavings you're trying to turn into a bullet disintegrate and you curse the entire lineage of the fixer who gave you these materials.\r\n", CH);
+                break;
               case 3:
-                send_to_char("The poor-quality metal shavings you're trying to turn into a bullet disintegrate. you curse the entire lineage of the fixer who gave you these materials\r\n", CH);
+                send_to_char("Somehow, an ancient unknown language comes unbidden to your lips. You think you said something unspeakable about a dog - regardless, the bullet is now tainted. You discard it.\r\n", CH);
                 break;
               case 4:
-                send_to_char("Somehow, Chinese comes unbidden to your lips. You think you said something unspeakable about a dog - regardless, the bullet is now tainted. You discard it.\r\n", CH);
-                break;
-              case 5:
                 send_to_char("You grind and grind and grind and grind and grind and grind and then you realize you have nothing left to grind.\r\n", CH);
                 break;
-              case 6:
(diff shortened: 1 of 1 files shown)


src/pro_create.cpp
```diff
@@ -599,7 +599,37 @@ void update_buildrepair(void)
...
       } else if (AFF_FLAGGED(CH, AFF_AMMOBUILD) && --GET_AMMOBOX_TIME_TO_COMPLETION(PROG) < 1) {
         if (GET_AMMOBOX_TIME_TO_COMPLETION(PROG) <= -2) // --(-1) = -2; prevents penalizing people who ace the test.
-          send_to_char("You seem to have messed up the batch of ammo.\r\n", CH);
+          (i = rand(1,9)) {
+            switch(i) {
+              case 1:
+                send_to_char("You seem to have messed up the batch of ammo.\r\n", CH);
+                break;
+              case 2:
+                send_to_char("You swear loudly as a gunshot rings out because you accidentally fired the round.\r\n", CH);
```
tuxkamen, line 608: Something about this one sounds a little off, syntactically.

src/pro_create.cpp
```diff
@@ -599,7 +599,37 @@ void update_buildrepair(void)
...
+                break;
+              case 3:
+                send_to_char("The poor-quality metal shavings you're trying to turn into a bullet disintegrate. you curse the entire lineage of the fixer who gave you these materials\r\n", CH);
```
tuxkamen, line 611: 'you curse'->'You curse', but also no ending period.

src/pro_create.cpp
```diff
@@ -599,7 +599,37 @@ void update_buildrepair(void)
...
+                break;
+              case 4:
+                send_to_char("Somehow, Chinese comes unbidden to your lips. You think you said something unspeakable about a dog - regardless, the bullet is now tainted. You discard it.\r\n", CH);
```
tuxkamen, line 614: I hate to be that guy, but what do the people who have skill in Chinese get for this one? :)

src/pro_create.cpp
```diff
@@ -599,7 +599,37 @@ void update_buildrepair(void)
...
+                break;
+              case 5:
+                send_to_char("You grind and grind and grind and grind and grind and grind and then you realize you have nothing left to grind.\r\n", CH);
+                break;
+              case 6:
+                send_to_char("You drop a tool, causing a spark to light from your gunpowder to the casing you're trying to pack. there's a disappointing 'pop' noise as it bursts into flames.\r\n", CH);
```
tuxkamen, line 620: 'there's' -> 'There's'.

src/pro_create.cpp
```diff
@@ -599,7 +599,37 @@ void update_buildrepair(void)
         STOP_WORKING(CH);
       } else if (AFF_FLAGGED(CH, AFF_AMMOBUILD) && --GET_AMMOBOX_TIME_TO_COMPLETION(PROG) < 1) {
         if (GET_AMMOBOX_TIME_TO_COMPLETION(PROG) <= -2) // --(-1) = -2; prevents penalizing people who ace the test.
-          send_to_char("You seem to have messed up the batch of ammo.\r\n", CH);
+          (i = rand(1,9)) {
```
tuxkamen, line 602: I think you need to use number(1,9) instead.

src/pro_create.cpp
```diff
@@ -599,7 +599,37 @@ void update_buildrepair(void)
...
+                break;
+              case 7:
+                send_to_char("You're pretty sure the bullet started talking to you. After several hours, things have started to blur together - better be safe, though. It might be a free spirit. You throw it away.\r\n", CH);
+                break;
+              case 8:
+                send_to_char("You were trying to craft APDS, weren't you? Poor bastard.\r\n", CH);
+                break;
+              case 9:
+                send_to_char("You stumble slightly, jamming your thumb in the vice. The process of ripping your hand out obliterates the shell casing.\r\n", CH);
+                break;
```
tuxkamen, line 630: My suggestion to you would be to *always* use a default case  (default:), especially if you're unsure if others will add on messages.  The default failure message (which is currently on 1) can be moved there, and then you can make the number() function have some configurable amount of 'default' failures by bumping the number higher than the bespoke messages (example: number(1,16) would give you half default, half bespoke).

UndeadNappist: Thanks for the help!

PR #569, opened 2021-09-16
Update utils.cpp

UndeadNappist:
Updates the names and verbage of skills as per HDcase's suggestions.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -2881,11 +2881,11 @@ const char *skill_rank_name(int rank, bool knowledge) {
     return "uh oh! you have a negative skill, please report!";
   
   RANK_MESSAGE(0, "not learned", "not learned");
-  RANK_MESSAGE(1||2, "Novice", "Scream-sheet level");
-  RANK_MESSAGE(3||4, "Competent", "Interested");
-  RANK_MESSAGE(5||6, "Proficient", "Dedicated");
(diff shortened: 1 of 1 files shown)


src/utils.cpp
```diff
@@ -2881,13 +2881,17 @@ const char *skill_rank_name(int rank, bool knowledge) {
...
+  RANK_MESSAGE(2, "novice", "interested");  
+  RANK_MESSAGE(3, "competent", "interested");
+  RANK_MESSAGE(4, "competent", "dedicated");
+  RANK_MESSAGE(5, "proficient", "well-rounded");
+  RANK_MESSAGE(6, "proficient", "educated");
+  RANK_MESSAGE(7, "skilled", "educated");
+  RANK_MESSAGE(8, "skilled", "intuitive");
+  RANK_MESSAGE(9, "professional", "expert");
+  RANK_MESSAGE(10, "professional", "expert");
+  RANK_MESSAGE(11, "expert", "specialist");
```
luciensadi, line 2894: We should probably keep 'expert' consistent as the top skill level, so maybe make the prior knowledge ones 'specialist' and this one 'expert'?

UndeadNappist: Done and done.

luciensadi:
Looks good, thanks!

PR #570, opened 2021-09-16
Mass find and replace of parentheses in skills and abilities.

UndeadNappist:
Uses mass find + mass replace to strip out parenths and slashes from skills and abilities and spells them out more clearly.

May require someone else to actually implement due to affecting helpfiles and a number of other things.


luciensadi:
This kind of thing can't be done with a mass find/replace due to the amount of potential issues an automated replace does (ex: this touched the SQL files, which I'm 99% sure it shouldn't have).

UndeadNappist:
> This kind of thing can't be done with a mass find/replace due to the amount of potential issues an automated replace does (ex: this touched the SQL files, which I'm 99% sure it shouldn't have).

Moved it back to not touch the SQL files and some other stuff. Not certain why github touched all the leave/enter messages, however. LeaveMsg:	ØD¢0£ seems a little strange to have in the codebase.

UndeadNappist:
Outdated.

PR #575, opened 2021-11-22
NERP metamagics get moved and removed from the MUD.

UndeadNappist:
What is says on the tin. NERP metamagics get moved and removed.

TODO: Add a trainer that only delivers NERP magical abilities.


UndeadNappist:
turns out moving 10010 has too many issues. Edit to that trainer reverted.

PR #576, opened 2021-11-24
Update spec_assign.cpp

UndeadNappist:
Gives a new spell trainer all the appropriate specprops for NERP metamagic.
Wanted to also give him NERP adept abilities, but couldn't understand how that array works.


luciensadi:
Thanks! This will throw errors for public-repo users though. Since this is a CE-only trainer, it should be enclosed in the private CE world ifdef:
```C++
#ifdef USE_PRIVATE_CE_WORLD
<your code here>
#endif
```

UndeadNappist:
Ah, good to know. Will keep that in mind going forwards. It also has to be moved to the sombero, but I don't have access there.

Issue #509, opened 2021-01-16
Unable to compile on Ubuntu 20·04


celtichawk:
As the subject says. Ubuntu 20.04 LTS on my system errors out.

Log below
```
scott@scott-System-Product-Name:~$ cd AwakeMUD/
scott@scott-System-Product-Name:~/AwakeMUD$ cd src/
scott@scott-System-Product-Name:~/AwakeMUD/src$ ./make
bash: ./make: No such file or directory
scott@scott-System-Product-Name:~/AwakeMUD/src$ make -s
/usr/bin/ld: newdb.o: in function `idle_delete()':
/home/scott/AwakeMUD/src/newdb.cpp:1968: undefined reference to `mysql_host'
/usr/bin/ld: /home/scott/AwakeMUD/src/newdb.cpp:1968: undefined reference to `mysql_user'
/usr/bin/ld: /home/scott/AwakeMUD/src/newdb.cpp:1968: undefined reference to `mysql_password'
/usr/bin/ld: /home/scott/AwakeMUD/src/newdb.cpp:1968: undefined reference to `mysql_db'
/usr/bin/ld: comm.o: in function `game_loop(int)':
/home/scott/AwakeMUD/src/comm.cpp:605: undefined reference to `gettimeofday(timeval*, timezone*)'
/usr/bin/ld: /home/scott/AwakeMUD/src/comm.cpp:631: undefined reference to `gettimeofday(timeval*, timezone*)'
/usr/bin/ld: /home/scott/AwakeMUD/src/comm.cpp:664: undefined reference to `gettimeofday(timeval*, timezone*)'
/usr/bin/ld: /home/scott/AwakeMUD/src/comm.cpp:683: undefined reference to `gettimeofday(timeval*, timezone*)'
/usr/bin/ld: /home/scott/AwakeMUD/src/comm.cpp:700: undefined reference to `gettimeofday(timeval*, timezone*)'
/usr/bin/ld: db.o: in function `initialize_and_connect_to_mysql()':
/home/scott/AwakeMUD/src/db.cpp:271: undefined reference to `mysql_host'
/usr/bin/ld: /home/scott/AwakeMUD/src/db.cpp:271: undefined reference to `mysql_user'
/usr/bin/ld: /home/scott/AwakeMUD/src/db.cpp:271: undefined reference to `mysql_password'
/usr/bin/ld: /home/scott/AwakeMUD/src/db.cpp:271: undefined reference to `mysql_db'
[...]

PR #577, opened 2021-11-25
Fixed gettimeofday() related linking issues on cutting edge distros…

Nodens-:
… by making the function a wrapper around std::chrono::systemclock::now() which is available on Linux, OSX and Windows

Fixed typos introduced by PRs 573 and 575

Second commit:
Pretty sure AwakeMUD won't compile, no matter what, on the following:
NeXT, Ultrix, MIPS OS, Sequent, Apollo (Domain/OS)

As they have been deprecated since the 90s and compiling C++11 on
them... So may as well cleanup the code and deprecate them
officially. By the Makefile it seems that only Linux, OSX and Cygwin are
currently officially supported but it will probably compile fine on
FreeBSD, AIX and HP-UX as well, albeit with a bit of tweaking on the
Makefile.

I plan to do a bit more cleaning up and also fix compilation with
`CXX = g++` as well.

### DIFF
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -22,6 +22,7 @@
 #include <new>
 #include <time.h>
 #include <iostream>
+#include <chrono>
 using namespace std;
 
 #if defined(WIN32) && !defined(__CYGWIN__)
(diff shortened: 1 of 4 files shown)


luciensadi:
Looks good to me, thanks!

Nodens-:
@luciensadi No problem :) You may also want to close https://github.com/luciensadi/AwakeMUD/issues/509 with this.

PR #578, opened 2021-11-27
Add OSX installation instructions to readme

bigfanguy:
I followed your instructions to get it working on my mac. I wanted to add some non-straightforward steps to make it easier for the next person.  Feel free to edit it however you like.


luciensadi:
Thanks! I think I just used `brew link` for the linking stage, but the specific `ln` commands might be the better choice. We'll go with this for now :)

PR #579, opened 2021-12-02
New util function get_string_after_color_code_removal()

Nodens-:
allows matching keywords from named/restrung objects no matter
how they're colored. Implemented on object selector and
deck/part/spell building functions along with
get_string_length_after_color_code_removal() for limiting
string lengths on color code stripped strings.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -3744,17 +3744,24 @@ int get_string_length_after_color_code_removal(const char *str, struct char_data
 
 // Returns a string stripped of color for keyword matching or possibly other uses as well.
 // We don't need to check for color code validity because we call get_string_legth_after_color_code_removal() when the strings are initially written.
-char* get_string_after_color_code_removal(const char *str, struct char_data *ch_to_notify_of_failure_reason) {
+char* get_string_after_color_code_removal(const char *str, struct char_data *ch) {
   if (!str) {
-    mudlog("SYSERR: Null string received to get_string_after_color_code_removal().", ch_to_notify_of_failure_reason, LOG_SYSLOG, TRUE);
+    mudlog("SYSERR: Null string received to get_string_after_color_code_removal().", ch, LOG_SYSLOG, TRUE);
     return NULL;
   }
     
   const char *ptr = str;
   static char clearstr [MAX_STRING_LENGTH];
+  memset(clearstr, 0, sizeof(clearstr));
   int pos = 0;
   
   while (*ptr) {
+    // Buffer overflow failsafe.
+    if (pos == MAX_STRING_LENGTH) {
+      clearstr[pos] = '\0';
+      return clearstr;
+    }
+    
     if (*ptr == '^') {
       // Parse a single ^ character.
       if (*(ptr+1) == '^') {
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -3742,6 +3742,47 @@ int get_string_length_after_color_code_removal(const char *str, struct char_data
   return len;
 }
 
+// Returns a string stripped of color for keyword matching or possibly other uses as well.
+// We don't need to check for color code validity because we call get_string_legth_after_color_code_removal() when the strings are initially written.
+char* get_string_after_color_code_removal(const char *str, struct char_data *ch_to_notify_of_failure_reason) {
+  if (!str) {
+    mudlog("SYSERR: Null string received to get_string_after_color_code_removal().", ch_to_notify_of_failure_reason, LOG_SYSLOG, TRUE);
+    return NULL;
+  }
+    
+  const char *ptr = str;
+  static char clearstr [MAX_STRING_LENGTH];
+  int pos = 0;
+  
+  while (*ptr) {
+    if (*ptr == '^') {
+      // Parse a single ^ character.
+      if (*(ptr+1) == '^') {
+        clearstr[pos] = *(ptr+1);
+        pos++;
+        ptr += 2;
+        continue;
+      }
+      // Skip color codes.
+      // There are two types of color: Two-character tags (^g) and xterm tags (^[F123]). We must account for both.
+      // 7 for xterm tags 2 for regular tags
+      else if (*(ptr+1) == '[') {
+          ptr  += 7;
+      }
+      else {
+        ptr += 2;
+      }
+    }
+    //Clear character, save it.
+    else {
+      clearstr[pos] = *ptr;
+      pos++;
+      ptr += 1;
+    }
+  }
+  return  clearstr;
+}
+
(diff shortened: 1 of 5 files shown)


luciensadi:
Hey, thanks for this! Looked it over, and I have a few recommendations:
- Zero out clearstr when the function starts, which guarantees that the function returns a null-terminated string (unless the input string is greater than or equal to the length of clearstr's array); currently, it appears to return a non-terminated string, which will cause unexpected behavior
- Add length checking such that if clearstr is about to overflow, execution stops and the array is returned in its existing state, along with an error being logged to console for followup investigation

src/utils.cpp
```diff
@@ -3742,6 +3742,47 @@ int get_string_length_after_color_code_removal(const char *str, struct char_data
   return len;
 }
 
+// Returns a string stripped of color for keyword matching or possibly other uses as well.
+// We don't need to check for color code validity because we call get_string_legth_after_color_code_removal() when the strings are initially written.
+char* get_string_after_color_code_removal(const char *str, struct char_data *ch_to_notify_of_failure_reason) {
+  if (!str) {
+    mudlog("SYSERR: Null string received to get_string_after_color_code_removal().", ch_to_notify_of_failure_reason, LOG_SYSLOG, TRUE);
```
luciensadi, line 3749: This doesn't actually notify the character, it just writes a syslog that has their name in it. You'll want to either send_to_char() them a message, or have the code that calls this send the error message.

Nodens-: Hi!

- The string is already nulll terminated because we copy char after char from the original null terminated string hence '\0' is being copied at the end as well. The function is just a copier, if the original string is null terminated then the resulting string will also be null terminated. I have already checked for this and the strings used in obj->restring are null terminated and my resulting return strings were also null terminated.
I could check at the end of the copy loop if `clearstr[pos] != '\0' and insert null termination on `clearstr[pos+1]` if not at max length or overwrite `clearstr[pos]` with null termination if it is (it's cheaper in cycles than zeroing clearstr). But I feel that this would just limit the function's usability (won't be able to return non-null terminated strings if that was required) and it's really the wrong place to ensure null termination as that should happen when writing the original strings to begin with (and it happens there already). This is also why I'm not doing any color code format testing as well because it is a waste of cycles when we have already done that in `get_string_length_after_color_code_removal()` when length checking, prior to writing the original strings.
Let me know if you want me to do this and limit the functionality to null terminated strings only. If we keep it as is, I can add a comment that the function will return the same type of string as what is passed to it (null terminated or not) for clarity.

- I didn't do this because the function is expecting proper existing strings up to `MAX_STRING_LENGTH` to convert to plaintext. I added the normal string length checks to the functions that were writing the original strings to begin with using `get_string_length_after_color_code_removal()` like it's done everywhere else. But I could add this for redundancy/fail safe if you want me to.

- Regading the mudlog() call, I know it doesn't send anything to the character heh. All errors that the character should be aware of are already handled in the caller functions and in `get_string_length_after_color_code_removal()`. I just added this syslog entry for consistency because `get_string_length_after_color_code_removal()` also did the same. I have a habit of conforming to debugging practices that are already present in the code :)

Let me know what you want me to do about 1 and 2!

Nodens-:
Hi!

- The string is already nulll terminated because we copy char after char from the original null terminated string hence '\0' is being copied at the end as well. The function is just a copier, if the original string is null terminated then the resulting string will also be null terminated. I have already checked for this and the strings used in obj->restring are null terminated and my resulting return strings were also null terminated.
I could check at the end of the copy loop if `clearstr[pos] != '\0' and insert null termination on `clearstr[pos+1]` if not at max length or overwrite `clearstr[pos]` with null termination if it is (it's cheaper in cycles than zeroing clearstr). But I feel that this would just limit the function's usability (won't be able to return non-null terminated strings if that was required) and it's really the wrong place to ensure null termination as that should happen when writing the original strings to begin with (and it happens there already). This is also why I'm not doing any color code format testing as well because it is a waste of cycles when we have already done that in `get_string_length_after_color_code_removal()` when length checking, prior to writing the original strings.
Let me know if you want me to do this and limit the functionality to null terminated strings only. If we keep it as is, I can add a comment that the function will return the same type of string as what is passed to it (null terminated or not) for clarity.

- I didn't do this because the function is expecting proper existing strings up to `MAX_STRING_LENGTH` to convert to plaintext. I added the normal string length checks to the functions that were writing the original strings to begin with using `get_string_length_after_color_code_removal()` like it's done everywhere else. But I could add this for redundancy/fail safe if you want me to.

- Regading the mudlog() call, I know it doesn't send anything to the character heh. All errors that the character should be aware of are already handled in the caller functions and in `get_string_length_after_color_code_removal()`. I just added this syslog entry for consistency because `get_string_length_after_color_code_removal()` also did the same. I have a habit of conforming to debugging practices that are already present in the code :)

Let me know what you want me to do about 1 and 2!

luciensadi:
I'm not actually sure that the string does come through null-terminated. Take the example of `str` pointing to an empty string, as seen in `get_string_after_color_code_removal("", NULL);`.

1. This passes `if (!str)` since the pointer is not null.
2. `ptr` is initialized to point at `str`, `clearstr` is not modified, and `pos` is set to 0.
3. The initial evaluation of `while (*ptr)` fails, because the value at `*ptr` is `'\0'` aka `0`, which is not a truthy value.
4. `clearstr` is returned without modification.

You can test this by throwing a quick debug print in db.cpp and modifying your util function slightly to fill clearstr with a value. With the following code:

```C++
void boot_world(void)
{
  log("Validating get_string_after_color_code_removal() with a nulled string...");
  log_vfprintf("Result: %s", get_string_after_color_code_removal("", NULL));
  log("Completed.");
  exit(0);
  <...>
```
and
```C++
char* get_string_after_color_code_removal(const char *str, struct char_data *ch_to_notify_of_failure_reason) {
  <...>
  static char clearstr [MAX_STRING_LENGTH];
  memset(clearstr, '1', sizeof(clearstr));
  <...>
```

I get this output:
```
Dec  4 10:28:36 :: Validating get_string_after_color_code_removal() with a nulled string...
Dec  4 10:28:36 :: Result: 11111<big ol' snip>11111
Dec  4 10:28:36 :: Completed.
```
Replacing the memset with a 0 instead of a '1' will fix the null-termination issues for any string that fits within MAX_STRING_LENGTH. We can't guarantee that the caller will always pass us a conformant string, though, so we should also do bounds checking as a failsafe.

No worries about the mudlog statement, I called that out based on the variable name indicating that the character was expected to be notified of failure. If that's not expected from this function, then changing the signature to a standard `struct char_data *ch` might serve well there.

Nodens-:
Ah right! I did not consider the case of passing an empty string because I assumed all sanity checks are happening when we initially write these strings from user input and we just need to convert existing valid strings for lookup purposes. Also good catch on reusing the buffer as the char array is static! I'll ensure null termination and clearing of the buffer. Commit coming up shortly.

luciensadi:
Excellent, thanks for this!

PR #581, opened 2021-12-08
Fix for autoconnecting clients (eg Mushclient) failing to complete

Nodens-:
telnet protocol options negotiation on reconnecting states.

### DIFF
--- a/src/interpreter.cpp
+++ b/src/interpreter.cpp
@@ -2324,6 +2324,12 @@ int perform_dupe_check(struct descriptor_data *d)
     break;
   }
   
+  //This is required for autoconnecting clients like MUSHClient
+  //in order to do telnet protocol options renegotiation on
+  //reconnection states. Otherwise things like MXP and 256
+  //color support break. It doesn't affect other clients.
+  d->pProtocol->bRenegotiate = TRUE;
+  
   // KaVir's protocol snippet.
   MXPSendTag( d, "<VERSION>" );
 


Nodens-:
To further explain this fix (since I was half asleep yesterday when I pushed the PR):
Due to how Kavir's implementation of telnet protocol options negotiation works, the negotiation breaks when the client sends non-negotiation data early before the negotiation process is complete. So when clients which support autoconnecting, like Mushclient, send name and password immediately on connect, they fail to complete options negotiation on log on.
This has been pointed out by Tijer (great guy, I know him heh) years ago and Kavir fixed it by implementing `bRenegotiate` boolean flag to trigger renegotiation at a later stage (after login). Frankly the proper solution would be to move options negotiation entirely after login but certain clients were already depending on the current behavior because Kavir's snippet was the one everyone used to implement these telnet protocol options (last I checked Mudlet did as well which is notorious for not following the telnet RFC--that was years ago but I doubt something changed there).
In this case, due to how descriptors are handled on reconnecting to a character that is currently logged in (linkdead etc), this renegotiation of telnet protocol options never gets triggered and that leaves options in a broken state with the only workaround being to disable autologin on the client (something that is inconvenient and most players would have no idea about anyway).
By setting `d->pProtocol->bRenegotiate = TRUE;` before calling `MXPSendTag();` we force this renegotiation to trigger.

 We could check for `TTYPE` and only trigger the renegotiation for clients who need it BUT that means having to maintain a list of clients and is prone to breakage. Specially when clients like MUSHClient allow custom setting of `TTYPE` and I've seen a lot of users putting custom strings there, not knowing what it is (which is reasonable as players don't go reading RFCs :P) or following some troubleshooting guide for a MUD with non-standard negotiation that instructs to use another `TTYPE` in order to support their client and then creating new client settings files by copy/pasting an existing file and editing them for the new mud which retains all such settings.
 So triggering the renegotiation for all clients reconnecting to logged on characters is the better solution as it doesn't break any client I've tested it on. It just makes some of them complete a second negotiation although they didn't require it.

luciensadi:
If this also fixes that bug that happens when you send non-negotiation data first and the MUD responds by spamming you with endless telnet control codes until you reconnect, you win all the prizes. That thing's been haunting us all year.

luciensadi:
Merged, thanks!

Nodens-:
Have not seen this yet but if it happened on autologin clients reconnecting to logged on characters, this will fix it. If you've seen it with Mudlet only though, I'm currently under the educated guess that they have messed up something again in their negotiation handling. Since like 6 months back (maybe more) and affecting all their versions since then including current. I have seen Mudlet users describing something like that, which can only be negotiation related and freezes mud I/O client side (spam of telnet control codes can certainly cause it). It was on another MUD I play, which I'm sure it follows the telnet specification, as I know the code maintainer well but I don't have access to code/packet sniffing to debug it there and he's too busy to bother with the MUD apart from serious reproducible bugs. I didn't bother to check Mudlet commits either because last I pointed out to them that they're not handling negotiation according to the RFC (again years ago) it got ignored and I had to implement special handling for it (It was MCCP related causing broken compression streams after copyover for Mudlet only due to their broken nego). My last interaction with Mudlet devs was me trying to explain thread safety and what mutex locks are to them in a github issue about implementation of async profile saving.. If you have wireshark dumps of the issue I'd like to see them though.

PR #588, opened 2021-12-18
increases quest_timer from 15 to 25

UndeadNappist:
Should encourage people to go find more runs and that sort of thing, especially with the faster johnson system. If they hit a snag in the 800-1k range, will write more content.

### DIFF
--- a/src/awake.h
+++ b/src/awake.h
@@ -2101,7 +2101,7 @@ enum {
 #define SERIOUS         3
 #define DEADLY          4
 
-#define QUEST_TIMER  15
+#define QUEST_TIMER  25
 
 #define OPT_USEC                  100000  /* 10 passes per second */
 #define PASSES_PER_SEC            (1000000 / OPT_USEC)


Nodens-:
Not sure if there are metrics already but just a heads up, that there could be situations where there are no 26 different runs available at some specific reputation point as johnsons get outrepped and others are locked in higher rep as well (including cases where the run is not feasible due to other reasons such as matrix runs on players that don't deck). In such a situation people would not be able to do anything at all. Perhaps a staff command is needed that simulates and prints out how many available runs there are for any rep range also showing how many of them are matrix runs etc. This tool would also help to identify reputation ranges that may need more content.

UndeadNappist:
Something to keep in mind from the builder side of thing is that KEY or ACCESS jobruns are almost always set to 10k rep nuyen as the cap. While most of them grant peanuts, they're also always available. So it's actually pretty hard to be completely locked out, even during the 800-1000 range where you have the least amount of runs.

UndeadNappist:
To address the point directly; a tool like that would be super-helpful and there are no metrics currently available.

Nodens-:
Roger that. The question is if there are enough of them. Are they more than 26? If yes there's no problem. As there will always be something available. Regarding the metrics tool, I noted it down in my todo list.

luciensadi:
Good feedback, I'll consider this one blocked pending metrics or other tooling to help us find what an acceptable threshold is.

UndeadNappist:
With the intro of TU and a few others, there might be johnsons that are inaccessible based on your species. Something to keep in mind.

luciensadi:
Things seem to be working fine with the current rotation cycle, I'm thinking we leave this one as-is unless there are a lot of players asking for longer cycles. Thanks for the suggestion though!

PR #591, opened 2021-12-29
A number of fixes.

Nodens-:
-Program names on create program were not factoring color codes.
-Shop lists (including OLC) did not format properly with colored items.
-Cyberware shops selling cyberlimbs did not format properly due to name length of cyberlimbs.
-Fingertip compartment is now a proper sheath for monowhips. Only usable as one if it's actually empty and cyber command now displays it's actual status `(F)` for Full/having something inside `(W)` for whip, `(W) (R)` if it's actually ready to draw.
-Conjuration drain did not factor related bioware at all. Now it does.
-Surprise attacks did not work at all with multiple points of failure. Now they do. I would like to see melee surprise attacks skipping the closing distance (you surprise the mob,  you obviously do it up close instead of start charging to it from the other side of the room) but I tested this and I think it is OP because ranged mobs stay in clash and don't try to break back into ranged. Perhaps an extra stealth check there at high TN (modified by Int?) would make sense for skipping the closing in.

### DIFF
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -3710,8 +3710,24 @@ ACMD(do_cyberware)
 
   send_to_char("You have the following cyberware:\r\n", ch);
   for (obj = ch->cyberware; obj != NULL; obj = obj->next_content) {
-    snprintf(buf, sizeof(buf), "%-40s Essence: %0.2f\r\n",
-            GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+    if (GET_OBJ_VAL(obj, 0) == CYB_FINGERTIP) {
+      if (obj->contains && GET_OBJ_TYPE(obj->contains) == ITEM_WEAPON && GET_HOLSTER_READY_STATUS(obj) ) {
+        snprintf(buf, sizeof(buf), "%-32s ^Y(W) (R)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+        send_to_char(buf, ch);
+        continue;
+      }
+      else if (obj->contains && GET_OBJ_TYPE(obj->contains) == ITEM_WEAPON && !GET_HOLSTER_READY_STATUS(obj)) {
+        snprintf(buf, sizeof(buf), "%-36s ^Y(W)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+        send_to_char(buf, ch);
+        continue;
+      }
+      else if (obj->contains) {
+        snprintf(buf, sizeof(buf), "%-36s ^Y(F)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+        send_to_char(buf, ch);
+        continue;
+      }
+    }
+    snprintf(buf, sizeof(buf), "%-40s Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
     send_to_char(buf, ch);
   }
 }
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -3710,24 +3710,24 @@ ACMD(do_cyberware)
 
   send_to_char("You have the following cyberware:\r\n", ch);
   for (obj = ch->cyberware; obj != NULL; obj = obj->next_content) {
-    if (GET_OBJ_VAL(obj, 0) == CYB_FINGERTIP) {
+    if (GET_CYBERWARE_TYPE(obj) == CYB_FINGERTIP) {
       if (obj->contains && GET_OBJ_TYPE(obj->contains) == ITEM_WEAPON && GET_HOLSTER_READY_STATUS(obj) ) {
-        snprintf(buf, sizeof(buf), "%-32s ^Y(W) (R)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+        snprintf(buf, sizeof(buf), "%-32s ^Y(W) (R)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_CYBERWARE_ESSENCE_COST(obj) / 100));
         send_to_char(buf, ch);
         continue;
       }
       else if (obj->contains && GET_OBJ_TYPE(obj->contains) == ITEM_WEAPON && !GET_HOLSTER_READY_STATUS(obj)) {
-        snprintf(buf, sizeof(buf), "%-36s ^Y(W)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+        snprintf(buf, sizeof(buf), "%-36s ^Y(W)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_CYBERWARE_ESSENCE_COST(obj) / 100));
         send_to_char(buf, ch);
         continue;
       }
       else if (obj->contains) {
-        snprintf(buf, sizeof(buf), "%-36s ^Y(F)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+        snprintf(buf, sizeof(buf), "%-36s ^Y(F)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_CYBERWARE_ESSENCE_COST(obj) / 100));
         send_to_char(buf, ch);
         continue;
       }
     }
-    snprintf(buf, sizeof(buf), "%-40s Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+    snprintf(buf, sizeof(buf), "%-40s Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_CYBERWARE_ESSENCE_COST(obj) / 100));
     send_to_char(buf, ch);
   }
 }
@@ -3745,7 +3745,7 @@ ACMD(do_bioware)
   for (obj = ch->bioware; obj != NULL; obj = obj->next_content) {
     snprintf(buf, sizeof(buf), "%-40s Rating: %-2d     Bioware Index: %0.2f\r\n",
             GET_OBJ_NAME(obj),
-            GET_OBJ_VAL(obj, 1), ((float)GET_OBJ_VAL(obj, 4) / 100));
+            GET_BIOWARE_RATING(obj), ((float)GET_BIOWARE_ESSENCE_COST(obj) / 100));
     send_to_char(buf, ch);
   }
 }
--- a/src/act.obj.cpp
+++ b/src/act.obj.cpp
@@ -3599,7 +3599,7 @@ ACMD(do_holster)
     send_to_char(ch, "%s is not a holsterable weapon.\r\n", capitalize(GET_OBJ_NAME(obj)));
     return;
   }
-  if (GET_OBJ_TYPE(cont) != ITEM_HOLSTER && (GET_OBJ_TYPE(cont) == ITEM_CYBERWARE && GET_OBJ_VAL(cont, 0) != CYB_FINGERTIP)) {
+  if (GET_OBJ_TYPE(cont) != ITEM_HOLSTER && (GET_OBJ_TYPE(cont) != ITEM_CYBERWARE || GET_CYBERWARE_TYPE(cont) != CYB_FINGERTIP)) {
     send_to_char(ch, "%s is not a holster.\r\n", capitalize(GET_OBJ_NAME(cont)));
     return;
   }
@@ -3610,11 +3610,8 @@ ACMD(do_holster)
 
   const char *madefor = "<error, report to staff>";
   
-  if (GET_OBJ_TYPE(cont) == ITEM_CYBERWARE && GET_OBJ_VAL(cont, 0) == CYB_FINGERTIP)
-    madefor = "monowhips";
-  else {
-    if (!holster_can_fit(cont, obj))
-      dontfit++;
+  if (GET_OBJ_TYPE(cont) == ITEM_HOLSTER && !holster_can_fit(cont, obj)) {
+    dontfit++;
     
     switch (GET_HOLSTER_TYPE(cont)) {
       case HOLSTER_TYPE_SMALL_GUNS:
@@ -3632,6 +3629,7 @@ ACMD(do_holster)
       return;
     }
   }
+
   if (GET_OBJ_SPEC(obj) == weapon_dominator) {
     dominator_mode_switch(ch, obj, DOMINATOR_MODE_PARALYZER);
   }
@@ -3669,7 +3667,7 @@ ACMD(do_ready)
       return;
     }
   }
-  if (GET_OBJ_TYPE(obj) != ITEM_HOLSTER && (GET_OBJ_TYPE(obj) == ITEM_CYBERWARE && GET_OBJ_VAL(obj, 0) != CYB_FINGERTIP)) {
+  if (GET_OBJ_TYPE(obj) != ITEM_HOLSTER && (GET_OBJ_TYPE(obj) != ITEM_CYBERWARE || GET_CYBERWARE_TYPE(obj) != CYB_FINGERTIP)) {
     send_to_char(ch, "%s is not a weapons holster.\r\n", capitalize(GET_OBJ_NAME(obj)));
     return;
   }
--- a/src/act.obj.cpp
+++ b/src/act.obj.cpp
@@ -48,6 +48,8 @@ void calc_weight(struct char_data *ch);
 
 SPECIAL(weapon_dominator);
 
+extern WSPEC(monowhip);
+
 int wear_bitvectors[] = {
                           ITEM_WEAR_TAKE, ITEM_WEAR_HEAD, ITEM_WEAR_EYES, ITEM_WEAR_EAR,
                           ITEM_WEAR_EAR, ITEM_WEAR_FACE, ITEM_WEAR_NECK, ITEM_WEAR_NECK,
@@ -3448,6 +3450,12 @@ int draw_weapon(struct char_data *ch)
 {
   struct obj_data *potential_holster, *obj;
   int i = 0;
+  
+  //Look in fingertip first to get it out of the way.
+  //At some point we need to write a mechanism to select what is drawn automatically -- Nodens
+  for (potential_holster = ch->cyberware; potential_holster; potential_holster = potential_holster->next_content)
+    if (GET_OBJ_VAL(potential_holster, 0) == CYB_FINGERTIP && GET_HOLSTER_READY_STATUS(potential_holster))
+      i += draw_from_readied_holster(ch, potential_holster);
 
   // Go through all the wearslots, provided that the character is not already wielding & holding things.
   for (int x = 0; x < NUM_WEARS && (!GET_EQ(ch, WEAR_WIELD) || !GET_EQ(ch, WEAR_HOLD)); x++) {
@@ -3496,6 +3504,13 @@ bool holster_can_fit(struct obj_data *holster, struct obj_data *weapon) {
 }
 
 struct obj_data *find_holster_that_fits_weapon(struct char_data *ch, struct obj_data *weapon) {
+  //If the weapon is a monowhip see if we have a fingertip compartment and it's not currently full
+  if (obj_index[GET_OBJ_RNUM(weapon)].wfunc == monowhip) {
+    struct obj_data *cont;
+    for (cont = ch->cyberware; cont; cont = cont->next_content)
+      if (GET_OBJ_VAL(cont, 0) == CYB_FINGERTIP && !cont->contains)
+        return cont;
+  }
   // Look at their worn items. We exclude inventory here.
   for (int x = 0; x < NUM_WEARS; x++) {
     // Is it a holster?
@@ -3584,7 +3599,7 @@ ACMD(do_holster)
     send_to_char(ch, "%s is not a holsterable weapon.\r\n", capitalize(GET_OBJ_NAME(obj)));
     return;
   }
-  if (GET_OBJ_TYPE(cont) != ITEM_HOLSTER) {
+  if (GET_OBJ_TYPE(cont) != ITEM_HOLSTER && (GET_OBJ_TYPE(cont) == ITEM_CYBERWARE && GET_OBJ_VAL(cont, 0) != CYB_FINGERTIP)) {
     send_to_char(ch, "%s is not a holster.\r\n", capitalize(GET_OBJ_NAME(cont)));
     return;
   }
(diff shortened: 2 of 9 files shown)


src/newfight.cpp
```diff
@@ -1153,11 +1181,27 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
...
+              if (IS_NPC(def->ch)) {
+                GET_MOBALERT(def->ch) = MALERT_CALM;
+                GET_MOBALERTTIME(def->ch) = 0;
+              }
               return;
+            }
           }
         }
       }
+      //Handle suprise attack/alertness here -- defender didn't die.
```
luciensadi, line 1198: If I'm reading this right, this is after the `if (!defender_died)` block-- doesn't that mean that the defender did actually die here? If so, this will crash.

src/newfight.cpp
```diff
@@ -1188,8 +1232,17 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
     act( rbuf, 1, att->ch, NULL, NULL, TO_ROLLS );
     
     // If the attacker dies from recoil, bail out.
-    if (damage(att->ch, att->ch, convert_damage(staged_dam), TYPE_HIT, FALSE))
+    if (damage(att->ch, att->ch, convert_damage(staged_dam), TYPE_HIT, FALSE)) {
```
luciensadi, line 1235: Same issue re: attacker dying and extracting the defending quest target.

src/newmagic.cpp
```diff
@@ -663,11 +663,38 @@ bool conjuring_drain(struct char_data *ch, int force)
   else
     drain = SERIOUS;
   drain = convert_damage(stage(-success_test(GET_CHA(ch), force), drain));
-  if (force > GET_MAG(ch) / 100)
+  
+  // Physical drain.
+  if (force > GET_MAG(ch) / 100) {
+    // Iterate through bioware to find relevant bioware.
+    for (struct obj_data *bio = ch->bioware; bio && drain > 0; bio = bio->next_content) {
+      if (GET_OBJ_VAL(bio, 0) == BIO_PLATELETFACTORY && drain >= 3)
```
luciensadi, line 671: This will produce inconsistent results depending on the ordering of the player's bioware list. If they're about to take 3 drain, and the trauma damper comes before the platelet factory, they won't get the effects of both. Maybe make a new `delta` variable to tally up the total expected changes, then apply them all at once after the loop?

Also, `GET_BIOWARE_TYPE()` here please!

src/newshop.cpp
```diff
@@ -1359,11 +1361,19 @@ void shop_list(char *arg, struct char_data *ch, struct char_data *keeper, vnum_t
       else strcpy(buf2, "-");
 
       if (IS_OBJ_STAT(obj, ITEM_NERPS)) {
-        snprintf(ENDOF(buf), sizeof(buf) - strlen(buf), "^Y(N)^n %-29s^n %-6s%2s   %0.2f%c  %9d\r\n", GET_OBJ_NAME(obj),
+        //Format string: "^Y(N)^n %-58s^n %-6s%2s   %0.2f%c  %9d\r\n"
+        //We apply padding for color codes here.
+        snprintf(paddingnumberstr, sizeof(paddingnumberstr), "%d", 58 + count_color_codes_in_string(GET_OBJ_NAME(obj)));
```
luciensadi, line 1366: 58 is an extremely magical number here, and I can't tell what it is or where it came from. If it's derived from the length of some spacer string, recommend breaking that spacer string out into its own component and replacing the 58 here with a `strlen(that_string)`.

src/newshop.cpp
```diff
@@ -1317,8 +1319,8 @@ void shop_list(char *arg, struct char_data *ch, struct char_data *keeper, vnum_t
 
 
   if (shop_table[shop_nr].flags.IsSet(SHOP_DOCTOR)) {
-    strlcpy(buf, " **   Availability     Item                              Rating  Ess/Index     Price\r\n"
-                 "------------------------------------------------------------------------------------\r\n", sizeof(buf));
+    strlcpy(buf, " **   Availability     Item                                                           Rating  Ess/Index     Price\r\n"
+                 "-----------------------------------------------------------------------------------------------------------------\r\n", sizeof(buf));
```
luciensadi, line 1323: The longer we make hardcoded strings like this, the more likely someone will have wrapping issues. Is it necessary to add this much spacing, or can we get away with making this shorter and condensing information elsewhere to fit?

src/newshop.cpp
```diff
@@ -1359,11 +1361,19 @@ void shop_list(char *arg, struct char_data *ch, struct char_data *keeper, vnum_t
...
+        snprintf(paddingnumberstr, sizeof(paddingnumberstr), "%d", 58 + count_color_codes_in_string(GET_OBJ_NAME(obj)));
+        snprintf(formatstr, sizeof(formatstr), "%s%s%s", "^Y(N)^n %-", paddingnumberstr, "s^n %-6s%2s   %0.2f%c  %9d\r\n");
+        snprintf(ENDOF(buf), sizeof(buf) - strlen(buf), formatstr, GET_OBJ_NAME(obj),
                 GET_OBJ_TYPE(obj) == ITEM_CYBERWARE ? "Cyber" : "Bio", buf2, ((float)GET_OBJ_VAL(obj, 4) / 100),
                 GET_OBJ_TYPE(obj) == ITEM_CYBERWARE ? 'E' : 'I', buy_price(obj, shop_nr));
       } else {
-        snprintf(ENDOF(buf), sizeof(buf) - strlen(buf), "%-33s^n %-6s%2s   %0.2f%c  %9d\r\n", GET_OBJ_NAME(obj),
+        //Format string: "%-62s^n %-6s%2s   %0.2f%c  %9d\r\n"
+        //We apply padding for color codes here.
+        snprintf(paddingnumberstr, sizeof(paddingnumberstr), "%d", 62 + count_color_codes_in_string(GET_OBJ_NAME(obj)));
```
luciensadi, line 1374: 62 is also magical! As are 46, 50, 30, 50, and 50 below.

luciensadi: Reviewing more, I see that you're pulling the numbers from the old format strings... it looks like you get the unenviable task of figuring out what the formula was for generating those numbers in the first place and then figuring out how to make that readable here. I suspect it was derived from the length of the dash padding in the tables, but I'm not certain.

src/utils.cpp
```diff
@@ -922,9 +922,6 @@ void mudlog(const char *str, struct char_data *ch, int log, bool file)
         case LOG_ECONLOG:
           check_log = PRF_ECONLOG;
           break;
-        case LOG_RADLOG:
```
luciensadi, line 925: Please add back in the radlog here

src/utils.h
```diff
@@ -106,6 +106,7 @@ struct  obj_data *get_obj_proto_for_vnum(vnum_t vnum);
 void    set_natural_vision_for_race(struct char_data *ch);
 int     get_string_length_after_color_code_removal(const char *str, struct char_data *ch_to_notify_of_failure_reason);
 char *  get_string_after_color_code_removal(const char *str, struct char_data *ch);
+int count_color_codes_in_string(const char *str) ;
```
luciensadi, line 109: Nit: please indent the function name to match the rest here.

luciensadi (changes requested):

src/act.informative.cpp
```diff
@@ -3710,8 +3710,24 @@ ACMD(do_cyberware)
 
   send_to_char("You have the following cyberware:\r\n", ch);
   for (obj = ch->cyberware; obj != NULL; obj = obj->next_content) {
-    snprintf(buf, sizeof(buf), "%-40s Essence: %0.2f\r\n",
-            GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+    if (GET_OBJ_VAL(obj, 0) == CYB_FINGERTIP) {
```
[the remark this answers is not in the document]
Nodens-: Got it!

src/act.obj.cpp
```diff
@@ -3584,7 +3599,7 @@ ACMD(do_holster)
     send_to_char(ch, "%s is not a holsterable weapon.\r\n", capitalize(GET_OBJ_NAME(obj)));
     return;
   }
-  if (GET_OBJ_TYPE(cont) != ITEM_HOLSTER) {
+  if (GET_OBJ_TYPE(cont) != ITEM_HOLSTER && (GET_OBJ_TYPE(cont) == ITEM_CYBERWARE && GET_OBJ_VAL(cont, 0) != CYB_FINGERTIP)) {
```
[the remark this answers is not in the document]
Nodens-: Right!

src/act.obj.cpp
```diff
@@ -3595,25 +3610,28 @@ ACMD(do_holster)
 
   const char *madefor = "<error, report to staff>";
   
-  if (!holster_can_fit(cont, obj))
-    dontfit++;
+  if (GET_OBJ_TYPE(cont) == ITEM_CYBERWARE && GET_OBJ_VAL(cont, 0) == CYB_FINGERTIP)
+    madefor = "monowhips";
```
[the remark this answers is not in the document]
Nodens-: I'm not sure what I was thinking here, it's been a while since I wrote this. Probably set it for the logging message that follows. with the intent to handle it there but forgot to remove it when I figured that I don't need to handle gating in `holster_can_fit()` at all.
The actual gating happens in `find_holster_that_fits_weapon()` where it will only return a fingertip if the weapon is a monowhip otherwise it goes for regular holsters (including if there's no figertip and weapon is monowhip). There was no need for redundant checking.

src/act.obj.cpp
```diff
@@ -3641,12 +3659,17 @@ ACMD(do_ready)
...
+  if ((finger = get_obj_in_list_vis(ch, buf, ch->cyberware)) && GET_OBJ_VAL(finger, 0) == CYB_FINGERTIP)
+    obj = finger;
+  else {
+    if (!(generic_find(buf, FIND_OBJ_EQUIP, ch, &tmp_char, &obj))) {
+      send_to_char(ch, "You don't seem to be using %s %s.\r\n", AN(argument), argument);
+      return;
+    }
   }
-  if (GET_OBJ_TYPE(obj) != ITEM_HOLSTER) {
+  if (GET_OBJ_TYPE(obj) != ITEM_HOLSTER && (GET_OBJ_TYPE(obj) == ITEM_CYBERWARE && GET_OBJ_VAL(obj, 0) != CYB_FINGERTIP)) {
```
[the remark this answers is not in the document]
Nodens-: Roger

src/act.obj.cpp
```diff
@@ -3659,6 +3682,11 @@ ACMD(do_ready)
     GET_HOLSTER_READY_STATUS(obj) = 0;
     return;
   } else {
+    //Check if we have any fingertip compartments that are readied
+    for (struct obj_data *cyber = ch->cyberware; cyber; cyber = cyber->next_content)
+      if (GET_OBJ_VAL(cyber, 0) == CYB_FINGERTIP && GET_HOLSTER_READY_STATUS(cyber))
```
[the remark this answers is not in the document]
Nodens-: It does. I made sure that the value is not used for anything for the fingertip compartment before going with this. There was no point for another macro and extra code since the value was free.

src/newfight.cpp
```diff
@@ -785,6 +785,13 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
       
       combat_message(att->ch, def->ch, att->weapon, -1, att->ranged->burst_count);
       damage(att->ch, def->ch, -1, att->ranged->dam_type, 0);
+      
+      //Handle suprise attack/alertness here -- ranged attack failed.
+      if (AFF_FLAGGED(def->ch, AFF_SURPRISE))
```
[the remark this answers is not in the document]
Nodens-: Right! I was sure this needed better examination and it's why the PR is in draft mode to begin with. Also I was not sure on the intended ALARM statuses and was expecting feedback before going further into it. Hence I had switched branch and worked on other stuff. ;)

src/newfight.cpp
```diff
@@ -834,6 +841,13 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
     if (IS_SPIRIT(def->ch) || IS_ELEMENTAL(def->ch)) {
       if (att->ranged->power <= GET_LEVEL(def->ch) * 2) {
         damage(att->ch, def->ch, 0, att->ranged->dam_type, att->ranged->is_physical);
+        
+        //Handle suprise attack/alertness here -- spirits ranged.
+         if (AFF_FLAGGED(def->ch, AFF_SURPRISE))
```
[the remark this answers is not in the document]
Nodens-: Roger

src/newfight.cpp
```diff
@@ -1036,6 +1050,13 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
         
         att->melee->power -= GET_IMPACT(def->ch) / 2;
       }
+      // Because we swap att and def pointers if defender wins the clash we need to make sure attacker gets proper values
+      // if they're using a ranged weapon in clash instead of setting their melee power to the damage code of the ranged
+      // weapon as it was happening.
+      else if (IS_RANGED(att->weapon)) {
+        att->melee->power = GET_STR(att->ch);
+        att->melee->damage_level = MODERATE;
```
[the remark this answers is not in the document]
Nodens-: The main problem here was that the code went from this:
`if (GET_OBJ_RNUM(att->weapon) >= 0 && obj_index[GET_OBJ_RNUM(att->weapon)].wfunc == monowhip) {`

to this else block:
```
else {
        att->melee->power = GET_WEAPON_STR_BONUS(att->weapon) + GET_STR(att->ch);
        att->melee->power -= GET_IMPACT(def->ch);
      }```
      
This obviously overwrites those initially set values with the values of the current weapon. And this would not normally be a problem because we're attacking with a melee weapon BUT, a few lines above, we swap the pointers around if the defender wins the clash (which is clever in itself as it saves us rewriting a lot of code), which means that in that case, we're now doing this for the defender without any check if the weapon they're actually holding is a melee weapon or a gun. This results in defenders winning a clash with a ranged weapon using its damage code in the melee clash and a mob holding eg an AR in FA mode murdering you with dam code it shouldn't have (an adept with counter holding a canon yay).

I could just check if the weapon is ranged and if not leave the values as they are but thought this was cleaner to begin with.

File: src/newfight.cpp [hunk 17]
```diff
@@ -1061,6 +1082,13 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
     if (IS_SPIRIT(def->ch) || IS_ELEMENTAL(def->ch)) {
       if (att->melee->power <= GET_LEVEL(def->ch) * 2) {
         damage(att->ch, def->ch, 0, att->melee->dam_type, att->melee->is_physical);
+        
+        //Handle suprise attack/alertness here -- spirits melee.
+         if (AFF_FLAGGED(def->ch, AFF_SURPRISE))
+           AFF_FLAGS(def->ch).RemoveBit(AFF_SURPRISE);
+
+        GET_MOBALERT(def->ch) = MALERT_ALERT;
```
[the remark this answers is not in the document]
Nodens-: Roger

src/newfight.cpp
```diff
@@ -1153,11 +1181,27 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
             int dam_total = convert_damage(stage(-1 * success_test(GET_BOD(attacker) + (successes == 0 ? GET_DEFENSE(attacker) : 0), 10), SERIOUS));
             
             // If the attacker dies from backlash, bail out.
-            if (damage(attacker, attacker, dam_total, TYPE_RECOIL, PHYSICAL))
+            if (damage(attacker, attacker, dam_total, TYPE_RECOIL, PHYSICAL)) {
+              
+              //Handle suprise attack/alertness here -- attacker died.
```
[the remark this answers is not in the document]
Nodens-: ROFL

src/newfight.cpp
Nodens-: No it's at the end of that block, the defender is alive. The bracket above is closing this block: `if (successes_for_use_in_monowhip_test_check <= 0) {`

`if (!defender_died)` block is closing at line 1205 so this one's ok, the defender is definitely alive.

src/newfight.cpp
Nodens-: Roger

src/newmagic.cpp
Nodens-: Hmm. You're right, I didn't think of that. You know this is exactly how it's implemented in spell_drain as well. So both need fixing there.

src/newshop.cpp
Nodens-: This is using the space that the currently longest string we have takes. And the width problem is already there, meaning the string output of the actual items are already that long and they just mess up the table. So basically this is fixing the padding so the tables print fine. It is 1 character longer than what the current longest string requires which is this: `a pair of Saeder-Krupp "Panzerfaust" obvious cyberarms -beta` (It's not even a huge string heh it's that we're printing a lot of columns on shop tables).

Now can we condense stuff? We probably can. We can shorten the special order output be removing the "Sp." part, we can change "Cyber" to "Cyb", little stuff like that to save 2 characters here, 2 characters there. The real question though is, is it worth it? We're at 2022 and there are really no 80 char wide terminals anymore. Several command outputs are already beyond the `LINE_LENGTH-1` because simple they don't fit in that no matter what. Monitors and all mud clients, including plain telnet, do not have that issue anymore, it's not the 80s, heh.
I have not checked for it but since this is the Kavir protocol options code, NAWS is probably enabled as well (unless it was never implemented entirely here but I've seen the nego code talking abou t it) which means all modern (as in more than a decade old :P ) clients can negotiate their wrapping width too. Even BlowTorch on android does not have an 80 char width limitation. MUD output on my monitor, and I'm sure on yours too, takes like 1/4th of available space with the rest being blank. Is it really worth beating this dead horse? :P

src/newshop.cpp
Nodens-: I am using the existing formatting numbers, which were probably arbitrarily derived at some point to a certain width fitting current-at-the-time strings. I am just appending the space that needs appending so that the tables will print properly for the existing strings and factoring in color codes space so columns align properly.

Nodens-: I have added the original format strings as comments both for clarity and for reference as I formulated the modifications. There was also 1 (if I remember right, could be 2) char error when it came down to aligning the header with the item prices for non-colored items as well so that's fixed as well.

src/newshop.cpp
Nodens-: As I said above, indeed I am deriving from the existing numbers. I am pretty sure the original numbers were not calculated algorithmically. They look like they were arbitrary chosen to fit current-at-the-time strings/dashes/formatting and were possibly extended over time for longer strings. Evidence on that is that tables between different types of stores do not match. Width per store type is as required for strings to fit. There's really nothing that can be done about this except, as I said above to start saving letters with abbreviations etc and that will only save us so much. But the question remains, is it worth conforming to a limitation that doesn't really apply nowadays, specially on strings that are nowhere else limited in the code (eg not passed to arrays with fixed short lengths etc)?

As a slightly related sidenote, I want to add more variables to prompt (a lot of them requested, eg progress) and largely enlarge the limit as, at least in the prompt's case, it's a string that only you can see and you should be able to set it as long as you want it to, even making it multiline with a new line variable. Could make nice color coded prompts with lots of information on them that currently don't fit and you're not forcing it on anyone else so I don't see why not.

src/utils.h
Nodens-: Aye aye!

src/utils.cpp
Nodens-: Wow. I have no idea how that got removed in the first place as I didn't touch that function at all. Nor had I any reason to... Either I was coding half asleep or the cat jumped on ctrl-x or something..rofl. Really this is crazy.
https://static1.srcdn.com/wordpress/wp-content/uploads/2020/11/tim-and-eric-mind-blown.jpg

src/newfight.cpp
luciensadi: Perfect, thanks!

src/newshop.cpp
luciensadi: The magic numbers bit isn't about shortening lines, it's just about making the code maintainable going forward. The solution I'm advocating for is a two-parter: first, pick the longest line in the output (usually the padding dashes at the top) and split it out into its own static string that you concatenate into the display, then replace the magic numbers with `strlen(the_new_string)`. That makes it clear what the padding is based off of, and also automatically updates the padding if we extend the string later.

Now, if that number was wholly arbitrary to begin with, or was based on something dumb like subtracting the total possible length of other fields from the length of the dash-line or something... eh. It should still be done, but the effort to do it becomes more than I'm willing to put in right now.

Prompts can be extended, it's just currently limited by the SQL field's length, so you'd need a migration for that and some code in db.cpp to check the migration was applied. Also, there's already a newline variable in prompts, so you wouldn't need to add that.

src/newshop.cpp
luciensadi: Nah, we can leave it be then and see how it goes over.

src/newfight.cpp
```diff
@@ -784,7 +784,15 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
...
       combat_message(att->ch, def->ch, att->weapon, -1, att->ranged->burst_count);
-      damage(att->ch, def->ch, -1, att->ranged->dam_type, 0);
+      bool target_died = 0;
+      target_died = damage(att->ch, def->ch, -1, att->ranged->dam_type, 0);
+      
+      //Handle suprise attack/alertness here -- ranged attack failed.
+      if (!target_died && AFF_FLAGGED(def->ch, AFF_SURPRISE))
+        AFF_FLAGS(def->ch).RemoveBit(AFF_SURPRISE);
+           
+      GET_MOBALERT(def->ch) = MALERT_ALARM;
```
luciensadi, line 794: These should go in a !target_died block as well, otherwise there's a chance they'll operate on invalid characters. Also, they probably should be guarded by `if (IS_NPC(def->ch))` to avoid setting these values on player characters.

src/newfight.cpp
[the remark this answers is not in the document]
luciensadi: Ahh, got it. Works for me then, thanks!

src/newshop.cpp
Nodens-: Ahhh gotcha. Yeah we can do that. It would probably mean splitting the top string into per column parts if we want to strlen() it to calculate the paddings for each section. I wonder if I should just replace those calls entirely with std::stringstream instead of using C-stryle formatting strings there. If I'm gonna rewrite it, might as well bring it to C++ and this century in the process.

I had not checked existing variables yet but yes but I've seen the SQL type for prompt is set to varchar(79) while working on the obj order saving stuff.

src/newfight.cpp
Nodens-: Of course. Did the updates half-asleep so I missed it heh.

src/newshop.cpp
Nodens-: I will rewrite this part in a later PR. As currently working on the one changing object ordering is making this branch a hassle to test as I have to dump and restore dbs continuously to change branches.

src/act.informative.cpp
```diff
@@ -3710,8 +3710,24 @@ ACMD(do_cyberware)
 
   send_to_char("You have the following cyberware:\r\n", ch);
   for (obj = ch->cyberware; obj != NULL; obj = obj->next_content) {
-    snprintf(buf, sizeof(buf), "%-40s Essence: %0.2f\r\n",
-            GET_OBJ_NAME(obj), ((float)GET_OBJ_VAL(obj, 4) / 100));
+    if (GET_CYBERWARE_TYPE(obj) == CYB_FINGERTIP) {
+      if (obj->contains && GET_OBJ_TYPE(obj->contains) == ITEM_WEAPON && GET_HOLSTER_READY_STATUS(obj) ) {
+        snprintf(buf, sizeof(buf), "%-32s ^Y(W) (R)^n Essence: %0.2f\r\n", GET_OBJ_NAME(obj), ((float)GET_CYBERWARE_ESSENCE_COST(obj) / 100));
```
luciensadi, line 3715: It just occurred to me that this isn't screenreader-friendly; we can merge it as is and I can fix it to avoid you having to drop DB to switch branches again, but generally speaking, anything that includes abbreviations like this should have a separate display mode for anyone flagged PRF_SCREENREADER. Something like `<item name> (Contains Weapon) (Readied)  Essence: ...`

src/newfight.cpp
```diff
@@ -833,7 +843,17 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
...
-        damage(att->ch, def->ch, 0, att->ranged->dam_type, att->ranged->is_physical);
+        bool target_died = 0;
+        target_died = damage(att->ch, def->ch, 0, att->ranged->dam_type, att->ranged->is_physical);
+        
+        //Handle suprise attack/alertness here -- spirits ranged.
+        if (!target_died && IS_NPC(def->ch)) {
+          if (AFF_FLAGGED(def->ch, AFF_SURPRISE))
+            AFF_FLAGS(def->ch).RemoveBit(AFF_SURPRISE);
+           
+          GET_MOBALERT(def->ch) = MALERT_ALERT;
```
luciensadi, line 854: This one does need to be MALERT_ALARM as well. MALERT_ALERT only really comes into play when they hear something fishy like gunfire. I know death screams also put them in an alert state, but I'm not sure if it's MALERT_ALERT or _ALARM. Regardless, being in active combat alarms them.

src/newfight.cpp
```diff
@@ -1152,12 +1189,27 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
...
               return;
           }
         }
       }
+      //Handle suprise attack/alertness here -- defender didn't die.
+      if (IS_NPC(def->ch)) {
+        if (AFF_FLAGGED(def->ch, AFF_SURPRISE))
+          AFF_FLAGS(def->ch).RemoveBit(AFF_SURPRISE);
+
+        GET_MOBALERT(def->ch) = MALERT_ALERT;
```
luciensadi, line 1210: ALARM / 30 here as well please.

src/newfight.cpp
luciensadi: I can also just fix this myself after merge if you don't want to switch branches again.

Nodens-:
Yeah let's do that, merge it, push the edits so I can pull and merge on my end as I'm sure I already have conflicts on the current branch.

PR #592, opened 2021-12-30
Reshuffling adept powers

UndeadNappist:
Moves spell shroud to the NERP trainer. While it is a functionally coded power, as far as I can tell detection spells are self-buffs only, making them useless in code.

The other part is adding a second adept trainer that moves the - quite frankly, really terrible abilities that are overpriced or completely outclassed by abilities that cost exactly the same (astral perception versus blind fighting) - and putting a warning sign on them in an adjacent room.

It's not a no, but it's definitely a "Please be experienced before walking into this archetype." I'm still waffling on how necessary it is.


UndeadNappist:
Decided splitting the unarmed was unnecessary, and it's probably better to tweak Nerve Strike, Blind Fighting, and Boosted (attribute) to make them more usable instead of fiddling around with trainers.

PR #593, opened 2022-01-10
Order of objects and first part of copyover related changes.

Nodens-:


src/comm.cpp
```diff
@@ -341,12 +341,28 @@ void copyover_recover()
...
-      else
-        load_room = real_room(GET_LOADROOM(d->character));
-      char_to_room(d->character, &world[load_room]);
-      //look_at_room(d->character, 0, 0);
+      // Since we're coming from copyover if the char is in a vehicle and is the owner throw him at the front seat
+      // otherwise throw him at the back. Because we don't save frontness and we'll assume the owner is driving
+      // anyway.
+      if (d->character->player_specials->saved.last_veh) {
+        for (struct veh_data *veh = veh_list; veh; veh = veh->next)
+          if (veh->idnum == d->character->player_specials->saved.last_veh && veh->damage < VEH_DAM_THRESHOLD_DESTROYED) {
```
luciensadi, line 349: We should abort scanning regardless when we hit a vehicle with idnum == last_veh. Bump the damage check to a nested check, maybe?

src/db.cpp
```diff
@@ -5069,28 +5143,76 @@ void load_consist(void)
...
-            if (inside > 0) {
-              if (inside == last_in)
-                last_obj = last_obj->in_obj;
-              else if (inside < last_in)
-                while (inside <= last_in) {
-                  if (!last_obj) {
-                    snprintf(buf2, sizeof(buf2), "Load error: Nested-item save failed for %s. Disgorging to room.", GET_OBJ_NAME(obj));
-                    mudlog(buf2, NULL, LOG_SYSLOG, TRUE);
-                    break;
+            if (house_version == VERSION_HOUSE_FILE) {
```
luciensadi, line 5146: Wouldn't this be the kind of place where you want to hardcode the version rather than put it in a global like this? My concern is that if the version is changed to 2, the maintainer will update VERSION_HOUSE_FILE to reflect that, and then this v1 code will no longer work for it. Might be better to do `if (house_version == VERSION_HOUSE_FILE_V1)` or something.

src/newdb.cpp
```diff
@@ -740,27 +745,57 @@ bool load_char(const char *name, char_data *ch, bool logon)
...
+      snprintf(buf2, sizeof(buf2), "Load error:  Objects in ware containers found with invalid containers for Char ID: %ld. Dumped in inventory.", GET_IDNUM(ch));
+      mudlog(buf2, NULL, LOG_SYSLOG, TRUE);
+    }
     mysql_free_result(res);
   }
   {
     struct obj_data *obj;
     int vnum = 0;
-    snprintf(buf, sizeof(buf), "SELECT * FROM pfiles_bioware WHERE idnum=%ld;", GET_IDNUM(ch));
+    snprintf(buf, sizeof(buf), "SELECT * FROM pfiles_bioware WHERE idnum=%ld ORDER BY Vnum;", GET_IDNUM(ch));
```
luciensadi, line 798: Why order by vnum?

src/newdb.cpp
```diff
@@ -1134,8 +1233,10 @@ static bool save_char(char_data *player, DBIndex::vnum_t loadroom)
     }
   }
 
-  /* Figure out their vehicle-- they can only load in it if they own it. */
-  if (player->in_veh && player->in_veh->owner == GET_IDNUM(player))
+  /* Figure out their vehicle-- they can only load in it if they own it.  Unless we're calling from copyover.*/
+  if (player->in_veh && fromCopyover)
```
luciensadi, line 1237: This is probably the better way to format this check for readability:
```C
if (player->in_veh && (fromCopyover || player->in_veh->owner == GET_IDNUM(player)))
  inveh = player->in_veh->idnum;
```

luciensadi (changes requested):
Added a few notes about nits. I haven't reviewed the actual saving logic yet, it's pretty dense; I'm also not sure I need to as long as you've tested it though.

src/db.cpp
Nodens-: The intent is to document format changes in the header. `VERSION_HOUSE_FILE` should always point to current and with any updates to the format old versions that need to be handled can be added to the header (as eg `VERSION_HOUSE_FILE_V1` and obviously the logic modified as it will need handling the extra changes anyway--which means replacing macros where needed). The intended design would be more obvious if I had a previous version available to list at the header but currently it's just 1 and none (with defaults to 0).

I did not hardcode them on the spot because if changes grow in number it'll be hard to document it as there would be version numbers there won the spot without much explanation (or multiple copies of comments--like I have now so someone can read what's going on without having to guess there's a comment at a different instance of the algo--that multiplied heh).
I plan to document this entirely in the header to keep everything consistent. By reading the comment there someone looking at the code for the first time should be able to figure out how to implement format changes and how to handle old ones. If someone doesn't understand how to handle that part properly objects could go poof.
 I wanted you to look at it though before I start documenting it.

src/newdb.cpp
Nodens-: Temporary trick to make bioware output consistent, since there's no `posi` field there to order by.

src/comm.cpp
```diff
@@ -341,12 +341,28 @@ void copyover_recover()
...
-        load_room = real_room(GET_LAST_IN(d->character));
-      else
-        load_room = real_room(GET_LOADROOM(d->character));
-      char_to_room(d->character, &world[load_room]);
-      //look_at_room(d->character, 0, 0);
+      // Since we're coming from copyover if the char is in a vehicle and is the owner throw him at the front seat
+      // otherwise throw him at the back. Because we don't save frontness and we'll assume the owner is driving
+      // anyway.
+      if (d->character->player_specials->saved.last_veh) {
+        for (struct veh_data *veh = veh_list; veh; veh = veh->next)
```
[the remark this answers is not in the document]
Nodens-: We don't actually have a room at that point but I grabbed it from last saved.

src/comm.cpp
Nodens-: Done

src/newdb.cpp
Nodens-: Done

Nodens-:
> Added a few notes about nits. I haven't reviewed the actual saving logic yet, it's pretty dense; I'm also not sure I need to as long as you've tested it though.

I've been testing it for about a full week in any possible way I could imagine. Obviously taking a backup of db and house/storage/veh is a good idea anyhow.

src/db.cpp
luciensadi: I'm not 100% sure I'm understanding this, let me try to rephrase my point and see if it makes sense.

Under imaginary logic v1, you have a field that's an int, and that works fine-- VERSION_HOUSE_FILE is 1, and this is a version-1 file, so it's all kosher.

Once you upgrade to logic v2, which changes that same field to a string, the parsing for v2 won't work on a v1 file. Thus, the coder potentially needs to remember to move all the logic from this current block to a new block with `if (house_version == 1)` (because this is the logic that's known good for version 1), then write the new logic in the VERSION_HOUSE_FILE block that's compatible with v2.

If instead this block was always named `if (house_version == v1)`, then there's no chance of forgetting that, and the coder can write a new `== v2` block above it with their new logic.

Nodens-: Yeah this works but there's no central documentation location in code that way. It's meant to look like this:

```
/* Current versions are defined in VERSION_HOUSE_FILE and VERSION_VEH_FILE
 * Making changes to the formats you need to change the current version macro
 * to the new version, which is what is written to file,  add a new macro
 * for the previous version, document the format changes here and handle them 
 * accordingly in code.
 */
// Current version adds X
#define VERSION_HOUSE_FILE 3
// Current version adds Y
#define VERSION_VEH_FILE 2

// Adds Blahblah required field.
#define VERSION_HOUSE_FILEV2
// Saves the objects in reverse with the container last. Then reads objects in turn, saves
// anything with a higher nesting level than 0 in a vector, then once a container is found
// places all objects with nesting level +1 into it.
#define VERSION_HOUSE_FILEV1
// Original logic, saves containers last and reads them first, then places subsequent objects
// with nesting level +1 into them. No version number written in file so this defaults to 0.
#define VERSION_HOUSE_FILEV0

// Saves the objects in reverse with the container last. Then reads objects in turn, saves
// anything with a higher nesting level than 0 in a vector, then once a container is found
// places all objects with nesting level +1 into it.
#define VERSION_VEH_FILEV1
// Original logic, saves containers last and reads them first, then places subsequent objects
// with nesting level +1 into them. No version number written in file so this defaults to 0.
#define VERSION_VEH_FILEV0```

So basically when you read the code and hit upon an macro, you go to its definition and got complete documentation for everything. Current version is what's written to the files, additional versions that need special handling are as macros with documentation below. Then change version handling accordingly using the defined macros as something may be something like an just an if block inside the current handling. Makes sense?

Nodens-: Do note that I absolutely loathe using macros in general :P I do hate our getter/setter ones with passion haha. But it is one of those cases where it makes sense. Still, if you don't like this design let me know and I'll change it.

luciensadi:
Looks good, thanks!

PR #596, opened 2022-01-29
Vehicles, drugs and adrenal pump.

Nodens-:
-Fixed vehicle speed which was stuck at 55 for everything on autonav. Speed now is actually meaningful.
-Complete revamp of load rules, bringing them to canon.
-Implemented canon overload rules and speed penalties.
-Implemented canon speed penalties on vehicular damage.
-Implemented canon pull tests for oveloaded vehicles depending on how overloaded they are. (These will need to be factored in other rolls as well but it's WIP).
-Vehicles towed or pushed into vehicles now have their used load factored in container vehicles.
-Implemented canon weight rules. (Need to also add Cargo Factor).
-Implemented canon stress damage (for overloading/pull tests at this point).
-Vehicles moving indoors speed penalty was changed to canon rough terrain rule of /2.
-Autonav does not do crash tests now. Previously it would not crash because the speed was always too low to make a crash test. 
-Added the missing header documentation for house/storage/veh file formats.
-Tagged the rest of the wrongly detected unused variables with UNUSED() macro.
-Removed an actual unused variable.
-Fixed compiler warning regarding the use of braces in macro. We currently have no compiler warnings at all.
-Fixed adrenal pump problem of having it's duration reduced in a way that made it end way too fast.
-Increased duration of pump.
-Slightly reduced duration of adrenal sack refill to 2-3minutes (was 160-200 seconds).
-Allowed reactivation of the pump before the sack is full with half the duration as per canon rule.
-Fixed drugs. They were entirely broken. State logic was not working at all drugs changed state every 2 seconds ignoring duration entirely.
-Added drug duration to status/affect command output.
-Added adrenal pump status to bioware command output.
-Cleaned up errors and warning pertaining to compiling with g++ which also resulted in a few bug and potential bug fixes.
-We now compile cleanly on both clang and g++.


Nodens-:
Slight correction to https://github.com/luciensadi/AwakeMUD/pull/596/commits/463702cb51c72f12b57125c4243d2fded3fadc23 commit message. The autonav speed modifier is actually `/10` not `/5` as I changed it after further testing but forgot to edit the commit message.

luciensadi:
Thanks for the PR! Would you mind writing up a summary of what changes are being made in the PR description?

Nodens-:
> Thanks for the PR! Would you mind writing up a summary of what changes are being made in the PR description?

Oops sorry, was 28 hours awake heh. Fixed.

src/act.movement.cpp
```diff
@@ -751,8 +751,9 @@ void move_vehicle(struct char_data *ch, int dir)
     ch->in_room = was_in;
   else
     ch->in_room = NULL;
-
-  if ((get_speed(veh) > 80 && SECT(veh->in_room) == SPIRIT_CITY) || veh->in_room->icesheet[0])
+    
+  // Don't do crash tests on autonav.
```
luciensadi, line 755: I'd argue that if autonav is moving at maximum speed, it should be doing crash tests against some standardized skill rating. A car going 400+ kph over a sheet of ice is going to have a bad day regardless of who's at the wheel.

src/act.wizard.cpp
```diff
@@ -5745,7 +5747,7 @@ ACMD(do_tail)
     strlcat(buf,"\r", sizeof(buf));
     send_to_char( buf, ch );
   }
-  fclose( out );
+  pclose( out );
```
luciensadi, line 5750: Why was this change made? Was fclose not the right call here?

src/chargen.cpp
```diff
@@ -149,7 +149,8 @@ void archetype_selection_parse(struct descriptor_data *d, const char *arg) {
     char helpbuf[100];
     char non_const_arg[MAX_INPUT_LENGTH];
     strncpy(non_const_arg, arg, sizeof(non_const_arg) - 1);
-    char *line = any_one_arg(non_const_arg, helpbuf);
+    char line[MAX_INPUT_LENGTH];
+    strncpy(line, any_one_arg(non_const_arg, helpbuf), MAX_INPUT_LENGTH- 1);
```
luciensadi, line 153: Since strncpy is not guaranteed to be zero-terminated, I've been going through replacing strncpys with strlcpy wherever I can. Please do the same when you see these!

src/db.cpp
```diff
@@ -5178,13 +5162,8 @@ void load_consist(void)
 
             snprintf(buf, sizeof(buf), "%s/Inside", sect_name);
             inside = data.GetInt(buf, 0);
+            // Lookup macro definition
             if (house_version == VERSION_HOUSE_FILE) {
-              // Since we're now saved the obj linked lists  in reverse order, in order to fix the stupid reordering on
```
luciensadi, line 5182: As long as these comments are accurate, they should be kept, as they're a good reference to what the code is doing. I'm always in favor of more verbose commenting.

src/fight.cpp
```diff
@@ -3030,6 +3046,7 @@ int check_recoil(struct char_data *ch, struct obj_data *gun)
   struct obj_data *obj;
   int rnum, comp = 0;
   bool gasvent = FALSE;
```
luciensadi, line 3048: We can probably just drop this variable, I don't know why it's here.

src/limits.cpp
```diff
@@ -1428,8 +1478,7 @@ void misc_update(void)
     }
 
     if (!IS_NPC(ch)) {
-      if (!CH_IN_COMBAT(ch))
-        check_adrenaline(ch, 0);
+      check_adrenaline(ch, 0);
```
luciensadi, line 1481: Why was this taken out of its in-combat guard?

src/newfight.cpp
```diff
@@ -1115,6 +1115,7 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
   int bod_success = 0;
   int bod = GET_BOD(def->ch) + (def->too_tall ? 0 : GET_BODY(def->ch));
   int bod_dice = 0;
+  UNUSED(bod_dice);
```
luciensadi, line 1118: Is bod_dice actually used in calculations? These unused-variable warnings are usually pretty reliable from what I've seen.

src/structs.h
```diff
@@ -233,20 +233,31 @@ struct room_data
...
+      canary(0),
+#endif
+      watching(NULL)
   {
     for (int i = 0; i < NUM_OF_DIRS; i++) {
       dir_option[i] = NULL;
       temporary_stored_exit[i] = NULL;
     }
     
+    memset(vision, 0, 3 * sizeof(int));
```
luciensadi, line 253: It might be better to do `memset(vision, 0, sizeof(vision) * sizeof(vision[0]));` (provided that sizeof works like I think it does and gives the number of elements in the array when given an array like that). This will prevent us from having to come back and change the memset if we change the variable type later.

luciensadi:
I'm reviewing the gridguide changes now, and I'm not sure uncapping the vehicle speed to maximum is the right move here. We're going to have souped-up Dynamits going at 400+ kph through residential streets in places, which GridGuide would not allow for safety reasons; making that kind of change would have to be part of a larger overhaul where we set up hacking of GridGuide's red/hard hosts as something deckers could do to remove speed limits for specific cars. 

It also looks like you've also removed the teleportation limiter, which would lead to vehicles traveling 40+ rooms per tick; this limiter was added previously due to teleportation-like driving that made the game world feel small.

luciensadi (commented):
Mostly just reviewed the vehicle code. This is the first pass of several reviews since this PR is too large for me to hold in my head at once.

src/act.drive.cpp
```diff
@@ -1815,27 +1816,87 @@ void process_autonav(void)
       if (!ch) {
         veh->dest = NULL;
         veh->cspeed = SPEED_OFF;
-        return;
+        continue;
       }
 
       int dir = 0;
-      for (int x = MIN(MAX((int)get_speed(veh) / 10, 1), MAX_GRIDGUIDE_ROOMS_PER_PULSE); x && dir >= 0 && veh->dest; x--) {
```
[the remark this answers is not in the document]
Nodens-: I removed it because it made speed pointless due to how autonav works..

The initial problem is that there was no reason whatsover to own a fast vehicle. Speed was capped at 55 for all of them, which means that even the slowest of vehicles with an immense max load was moving at exactly the same speed, or EVEN faster than a fast vehicle which could crawl due to the load rule that was non-canon to begin with. But even removing the artificial 55 cap it still made almost no difference at all because of the max rooms per 3 seconds setting being that low.

The design intend is to remove the artificial cap and scale real speed according to vehicle speed. So that limiter had to go and a universal modifier applied vs vehicle speed directly. I performed tests (they're documented in the comment in config.h) but I was not aware that you can reach 400 speed (how? what mods? I thought you can only get like +20-30 speed via mods). This means the modifier needs to become heavier and/or perhaps apply an upper limit as well but closer to 25, not 10 because again we'd have all vehicles moving at the same speed or we'd have to make slow very vehicles extremely slow with a applying a heavier modifier.

The plan basically is to move autonav to worker thread so we can do this properly and apply a wait state *per room moved* depending on vehicle speed but did not go directly there because the vehicle/rigging system is getting overhauled and I can't plan ahead on mutex locks and thread safety before having the concrete design (or at least the basic skeleton).

src/utils.h
```diff
@@ -868,6 +863,20 @@ bool WEAPON_FOCUS_USABLE_BY(struct obj_data *focus, struct char_data *ch);
 
 // ITEM_DRUG convenience defines
 
+#define GET_DRUG_TYPE(part)                        (GET_OBJ_VAL((part), 0))
+
+#define GET_DRUG_AFFECT(ch)                        (ch->player_specials->drug_affect[0])
```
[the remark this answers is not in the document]
Nodens-: Moved them because that was the ITEM_DRUG section and I added a new one there. Thought we were grouping those per item type for easy reference but if it's only for GET_OBJ_VALs I can move them back.

src/act.drive.cpp
```diff
@@ -1815,27 +1816,87 @@ void process_autonav(void)
...
-        return;
+        continue;
       }
       if (veh->in_room == veh->dest) {
         send_to_veh("Having reached its destination, the autonav shuts off.\r\n", veh, 0, TRUE);
         veh->cspeed = SPEED_OFF;
         veh->dest = NULL;
+        //Pull test for vehicles between * 1.0 and * 1.5 load for stress damage happens upon reaching destination and every 59 MUD minutes in save_vehicles().
+        // Rigger 3 p.64-65 & Shadowrun 3 p. 145-147
+        if (veh->damage < VEH_DAM_THRESHOLD_DESTROYED && GET_VEH_ISOVERLOADED(veh) == LOAD_HEAVY) {
```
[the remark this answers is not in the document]
Nodens-: That's intended. LOAD_HEAVY vehicles incur the test only when they arrive at their destination on autonav. LOAD_MAX vehicles will incur it on ever room moved (the test happens in the autopathing loop above at the proper TN).
This is my interpretation of the canon rules which have the tests done per minute for LOAD_MAX and per hour for LOAD_HEAVY. The intent of the canon rules is that you should be able to move a bit while being very overloaded (although on high TN tests) and you should be able to able to move a lot longer while being a bit overloaded (always risking stress damage). Due to the difference in time abstractions beyond TT and the Awake, this design made sense. I follow this logic on everywhere these tests are applied.

src/act.drive.cpp
```diff
@@ -1815,27 +1816,87 @@ void process_autonav(void)
...
+    if (veh_collapsed) {
+      veh->cspeed = SPEED_OFF;
+      veh->dest = NULL;
+      struct char_data *tch = NULL, *next = NULL;
+      for (tch = veh->people; tch; tch = next) {
+        next = tch->next_in_veh;
+        char_from_room(tch);
+        char_to_room(tch, veh->in_room);
+        AFF_FLAGS(tch).RemoveBits(AFF_PILOT, AFF_RIG, ENDBIT);
+        send_to_char("Your vehicle collapses under load strain and you get out!\r\n", tch);
```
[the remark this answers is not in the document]
Nodens-: Yeah I was thinking about that as well. But wanted to bounce it off you, Not sure if collapsing due to extreme weight is actually dangerous for the passengers or not. We're technically not crashing..it's more like the wheel axons (sorry not sure on English terminology on that stuff so not sure if I make sense heh) break and the chassis hits the ground, probably halting shortly with sparks flying off the asphalt... What do you think?

src/utils.h
```diff
@@ -338,14 +338,20 @@ extern bool PLR_TOG_CHK(char_data *ch, dword offset);
...
+/* veh utils  ************************************************************/
 
 #define GET_VEH_NAME(veh) (decapitalize_a_an((veh)->restring ? (veh)->restring : ((veh)->short_description ? (veh)->short_description : "an ERRONEOUS VEHICLE")))
 #define GET_VEH_DESC(veh) ((veh)->restring_long ? (veh)->restring_long : (veh)->long_description)
 #define GET_VEH_RNUM(veh) ((veh)->veh_number)
 #define GET_VEH_VNUM(veh) (GET_VEH_RNUM(veh) >= 0 ? veh_index[GET_VEH_RNUM(veh)].vnum : -1)
+// Shadowrun 3 p.133 & Rigger 3 p.63 (Will fix details such as bench seats later on as the system fleshes out -- Nodens)
+#define GET_VEH_MAXLOAD(veh) ((veh)->load + MAX(0,((veh)->seating[0] + (veh)->seating[1])) * 100)
+#define GET_VEH_MAXOVERLOAD(veh) ((veh)->load * 2 + MAX(0, ((veh)->seating[0] + (veh)->seating[1])) * 100)
+#define GET_VEH_ISOVERLOADED(veh) ((veh)->usedload > GET_VEH_MAXLOAD((veh)) * 1.5 ? 2 : (veh)->usedload > GET_VEH_MAXLOAD((veh)) ? 1 : 0)
```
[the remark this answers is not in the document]
Nodens-: Good call!

src/db.cpp
Nodens-: They were repetitive saying the same thing so I thought directing the documentation to the header was cleaner. Can add them back.

src/fight.cpp
Nodens-: You're right, I saw it was set below, didn't see that the value was not consumed anywhere.

src/act.drive.cpp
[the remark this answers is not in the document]
Nodens-: I want to add that "teleporting" can not be currently (as in as long as we run this every 3 seconds) avoided for very small distances and initiating autonav close to the next execution of process_autonav() but I don't think that it's inherently bad. You can *walk* 10 rooms in less time than it takes to go with a vehicle heh.

src/limits.cpp
Nodens-: Because the in-combat call was removed entirely and this now controls the pump entirely. The previous system with two calls running in different intervals/conditions made controlling pump duration too erratic (It was way to short finishing immediately most of the time). This allows a finer control on the pump duration. Again time abstraction issues here. It now last a lot longer but not sure if the duration I set it to is ok (It's a random roll of 1d6 per level of the pump.. (so max 2d6) I multiplied the duration by 10, it's more usable now but the possible duration range is large).

src/newfight.cpp
Nodens-: Hmm you're right. I saw the awake check and the spirit defense part and I wrongly assumed it's used heh. It's actually not. Remove it? Or use it?

src/structs.h
Nodens-: No sizeof(vision[0]) will return the size of the pointer to the value (so a fixed size depending on platform). All arrays used as parameters to function calls as pointers or by value, decay into pointers in C. What is actually passed to the function is a pointer to the value. Look up "array decay in C++" for detailed info.

src/act.drive.cpp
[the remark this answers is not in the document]
Nodens-: Now that I think about it, an upper limit of 25 will just make those speed mods that can reach 400 speed irrelevant. Can you give me a configuration of mods that reaches 400 so I can make the fastest vehicle and test it with stock vehicles etc to find an acceptable modifier that doesn't make speed stuff useless until I can redesign this with wait states per room?

Nodens-:
See the reply on the code comment. I'm not allowing them to go max speed. I'm removing the flat caps and then applying a global reduction modifier uniformly against vehicle speed directly. The design goal is to reduce speed enough for not teleporting but to retain meaningful speed difference between vehicle types at all times. The old system made speed 100% irrelevant. Everything moved at the same speed and fast cars when loaded with a few items were actually going slower than a Bison with 500 used load. :P

I was not aware that you can mod a vehicle to 400 speed if I was I would have tested against that. If you or someone else can give me such a configuration to build the fastest speed possible for testing I would adjust the modifier further. My testing was done using the fastest available vehicle (Saab) vs a commonly used high load vehicle (the Bison which you can get in chargen too). In my tests with the current modifier a seattle center<->west tacoma ride takes 8-9 seconds on Saab, 13-14s on Bison. That's not teleporting and I think it's quite acceptable, no but if speeds such as 400 are indeed attainable I need to build such a vehicle and adjust for the upper limit.

Also I DO plan on implementing everything related to the vehicle/rigging system including grid hacks. ;) I just started with the speed/load stuff because I was looking at fixing the "speed is irrelevant" issue.

src/act.drive.cpp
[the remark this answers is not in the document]
Nodens-: I'm sure you've seen it already but both of these overloaded conditions are also checked in intervals every mud hour at the saving function. I wanted to separate them in different interval locations but couldn't find a better solution as using a faster check for the LOAD_MAX condition would destroy a vehicle in seconds due to the very high TN tests and most vehicles having low body values. The design as it is will allow you to overload a lot a vehicle for a few MUD ticks or a few rooms before it goes poof. It's a risky endeavor meant to be used in extreme need (Which will probably become more meaningful as I add more of the vehicle/rigging stuff in). LOAD_HEAVY will allow you to temporarily overload for a quick run (eg moving your stuff from your apartment to another one), risking some damage in the process.

src/act.drive.cpp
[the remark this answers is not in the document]
luciensadi: You don't need mods to make a vehicle hit 400, as staff you can just edit the vehicle itself. It's what I do when I need test vehicles. I also don't have an issue with making this tick faster (every second? half-second?) and dividing the rooms traveled to match. Not sure what that would raise the effective speed minimum to though.

src/utils.h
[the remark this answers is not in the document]
luciensadi: Yeah, please do, this will confuse me later otherwise. Thanks!

src/newfight.cpp
Nodens-: I also need to go back and recheck the rest I have tagged. Some of them were used in ifdefs (which was why I added the macro in the first place as there's no way around that) but I may have hastily tagged a few that I saw being set without checking if they're actually consumed.

src/act.drive.cpp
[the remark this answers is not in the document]
luciensadi: At maximum speed, having your axles / shocks collapse will do a whole lot of damage and also make the car uncontrollable. It'd be worse than yanking the handbrake at highway speeds, because maximum speed here is far higher than highway speed, and the handbrake also doesn't drop the chassis into contact with the road and risk sending broken parts / fire into the cabin. This is a pretty traumatic experience for a car!

src/db.cpp
luciensadi: Yes please. I think the ideal end state would be shifting all of the reproduced saving logic out into a centralized function that we can reuse in multiple places, but for now duplicating the documentation across the separate functions will help.

src/newfight.cpp
luciensadi: I don't know what it was for in the first place, but if it's not being used now then it's probably not needed. Might be worth running the function's logic by a SR3 expert to make sure we're not missing something here, I definitely don't qualify there since everything I know about the system I learned from the codebase.

src/act.drive.cpp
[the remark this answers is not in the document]
Nodens-: Jack gave me a list of all the speed mod parts a few minutes ago so I'll build the fastest Saab possible and test with that.

src/act.drive.cpp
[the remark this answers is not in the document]
Nodens-: Roger that. Will apply crash damage to passengers then like we do on the crash tests.

src/db.cpp
Nodens-: Roger that!

src/newfight.cpp
Nodens-: From what I see it is meant to be used for 2 purposes. To eliminate your target's body cpool dice (soaking) in case they're sleeping and to allow a player character to use their skill level in conjuring as cpool body dice. This was meant to make the dice 0 in these cases but frankly we can implement both of these cases without using an extra variable at all. So I'll remove it and ask if someone can tell me about the sleeping and conjuring rules and add those directly if they actually apply.

src/act.drive.cpp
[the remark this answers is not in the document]
Nodens-: Oh one sidenote, regarding making this tick faster. That alone doesn't help us solve the issue because we can't apply wait states. Any wait states applied will block the main thread for x duration per room per vehicle moving on autonav lagging everything to hell and back. What it would do, without wait states, is it would just allow us to get more accurate measurements and predictable results when someone initiates autonav BUT at the "cost" of removing the up to 3 second delay per autopathing cycle, essentially making vehicles move WAY faster.

In fact depending on how my tests go, I may move process_autonav() to a slower interval, artificially increasing the minimum time required for autonav. That is if I can't settle on a speed modifier that works for the upper in-game possible limit and also doesn't make slow vehicles need 5 minutes to move. :P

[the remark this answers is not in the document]
Nodens-: Or faster interval with even slower modifier resulting in less rooms per cycle.. Hmm I need to test to see what works but now that I have the means to get the highest possible speed in game, I can run proper tests and tune this.

luciensadi:
The adrenal pump and drug changes have been merged, as well as the vehicle-itself load size functions etc. I haven't brought in the autonav or vehicle overload changes though.

With that done, this PR has gone pretty stale, so I'm going to go ahead and close it. Thanks for submitting this!

PR #597, opened 2022-01-29
Update act.drive.cpp

UndeadNappist:
It is now on players to handle punctuation.

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -356,12 +356,12 @@ ACMD(do_vemote)
     return;
   }
   RIG_VEH(ch, veh)
-  snprintf(buf, sizeof(buf), "%s%s.\r\n", capitalize(GET_VEH_NAME(veh)), argument);
+  snprintf(buf, sizeof(buf), "%s%s\r\n", capitalize(GET_VEH_NAME(veh)), argument);
   if (veh->in_room)
     send_to_room(buf, veh->in_room);
   else
     send_to_veh(buf, veh->in_veh, ch, TRUE);
-  snprintf(buf, sizeof(buf), "Your vehicle%s.\r\n", argument);
+  snprintf(buf, sizeof(buf), "Your vehicle%s\r\n", argument);
   send_to_veh(buf, veh, NULL, TRUE);
   return;
 }
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -356,12 +356,12 @@ ACMD(do_vemote)
     return;
   }
   RIG_VEH(ch, veh)
-  snprintf(buf, sizeof(buf), "%s%s\r\n", capitalize(GET_VEH_NAME(veh)), argument);
+  snprintf(buf, sizeof(buf), "%s%s%s^n\r\n", capitalize(GET_VEH_NAME(veh)), argument, ispunct(get_final_character_from_string(argument)) ? "" : ".");
   if (veh->in_room)
     send_to_room(buf, veh->in_room);
   else
     send_to_veh(buf, veh->in_veh, ch, TRUE);
-  snprintf(buf, sizeof(buf), "Your vehicle%s\r\n", argument);
+  snprintf(buf, sizeof(buf), "Your vehicle%s%s^n\r\n", argument, ispunct(get_final_character_from_string(argument)) ? "" : ".");
   send_to_veh(buf, veh, NULL, TRUE);
   return;
 }
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -4592,7 +4592,7 @@ ACMD(do_syspoints) {
 
   // Require all arguments.
   if (!*arg || !*target || !*amt || !*reason || k <= 0 ) {
-    send_to_char(ch, "Syntax: syspoints <award|penalize> <player> <%samount> <Reason for award>.\r\n", k < 0 ? "POSITIVE " : "");
+    send_to_char(ch, "Syntax: syspoints <award|penalize> <player> <%samount> [for] <Reason for award>.\r\n", k < 0 ? "POSITIVE " : "");
     return;
   }
 
@@ -4604,7 +4604,7 @@ ACMD(do_syspoints) {
     award_mode = FALSE;
     k *= -1;
   } else {
-    send_to_char("Syntax: syspoints <award|penalize> <player> <amount> <Reason for award>.\r\n", ch);
+    send_to_char("Syntax: syspoints <award|penalize> <player> <amount> [for] <Reason for award>\r\n", ch);
     return;
   }
 
@@ -4637,14 +4637,14 @@ ACMD(do_syspoints) {
     }
 
     // Mail the victim.
-    snprintf(buf, sizeof(buf), "You have been %s %d system points for %s.\r\n",
+    snprintf(buf, sizeof(buf), "You have been %s %d system points for%s\r\n",
             (award_mode ? "awarded" : "penalized"),
             k,
             reason);
     store_mail(idnum, ch, buf);
 
     // Notify the actor.
-    send_to_char(ch, "You %s %d system points %s %s for %s.\r\n",
+    send_to_char(ch, "You %s %d system points %s %s for %s\r\n",
                 (award_mode ? "awarded" : "penalized"),
                 k,
                 (award_mode ? "to" : "from"),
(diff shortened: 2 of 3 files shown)


src/act.drive.cpp
```diff
@@ -356,12 +356,12 @@ ACMD(do_vemote)
     return;
   }
   RIG_VEH(ch, veh)
-  snprintf(buf, sizeof(buf), "%s%s.\r\n", capitalize(GET_VEH_NAME(veh)), argument);
+  snprintf(buf, sizeof(buf), "%s%s\r\n", capitalize(GET_VEH_NAME(veh)), argument);
```
luciensadi, line 359: Since we're already using a snprintf here, we may as well take advantage of an inline conditional to provide punctuation. There's a good example of this with `ispunct()` in newecho.cpp, where it's used to punctuate emotes that don't have terminal punctuation already.

src/act.other.cpp
```diff
@@ -4592,7 +4592,7 @@ ACMD(do_syspoints) {
 
   // Require all arguments.
   if (!*arg || !*target || !*amt || !*reason || k <= 0 ) {
-    send_to_char(ch, "Syntax: syspoints <award|penalize> <player> <%samount> <Reason for award>.\r\n", k < 0 ? "POSITIVE " : "");
+    send_to_char(ch, "Syntax: syspoints <award|penalize> <player> <%samount> [for] <Reason for award>.\r\n", k < 0 ? "POSITIVE " : "");
```
luciensadi, line 4595: You've added a `[for]` here but not added logic to strip this out, which will cause `syspoints award player X for this reason` to output 'You have been awarded X system points for for this reason'.

src/act.other.cpp
```diff
@@ -4637,22 +4637,22 @@ ACMD(do_syspoints) {
     }
 
     // Mail the victim.
-    snprintf(buf, sizeof(buf), "You have been %s %d system points for %s.\r\n",
+    snprintf(buf, sizeof(buf), "You have been %s %d system points for%s\r\n",
```
luciensadi, line 4640: Why was the space removed? Is it outputting with two spaces currently?

src/act.other.cpp
```diff
@@ -4637,22 +4637,22 @@ ACMD(do_syspoints) {
...
             (award_mode ? "awarded" : "penalized"),
             k,
             reason);
     store_mail(idnum, ch, buf);
 
     // Notify the actor.
-    send_to_char(ch, "You %s %d system points %s %s for %s.\r\n",
+    send_to_char(ch, "You %s %d system points %s %s for %s\r\n",
```
luciensadi, line 4647: Same change re: terminal punctuation, it should be auto-supplied if the player has not supplied it already.

src/act.other.cpp
```diff
@@ -4637,22 +4637,22 @@ ACMD(do_syspoints) {
...
                 (award_mode ? "awarded" : "penalized"),
                 k,
                 (award_mode ? "to" : "from"),
                 capitalize(target),
                 reason);
 
     // Log it.
-    snprintf(buf, sizeof(buf), "%s %s %d system points %s %s for %s (%d to %d).",
+    snprintf(buf, sizeof(buf), "%s %s %d system points %s %s for %s (%d to %d)",
```
luciensadi, line 4655: This is a log message that the player doesn't see, and the staff member cannot alter the ending punctuation here. Please restore the period.

src/act.wizard.cpp
```diff
@@ -2511,7 +2511,7 @@ ACMD(do_award)
   k = atoi(amt);
 
   if (!*arg || !*amt || !*reason || k <= 0 ) {
-    send_to_char("Syntax: award <player> <karma x 100> <Reason for award>.\r\n", ch);
+    send_to_char("Syntax: award <player> <karma x 100> [for] <Reason for award>\r\n", ch);
```
luciensadi, line 2514: Same note re: `[for]` not being removed.

src/act.wizard.cpp
```diff
@@ -2535,11 +2535,11 @@ ACMD(do_award)
...
     gain_karma(vict, k, TRUE, FALSE, FALSE);
 
-  send_to_char(vict, "You have been awarded %0.2f karma for %s.\r\n", (float)k*0.01, reason);
+  send_to_char(vict, "You have been awarded %0.2f karma for %s\r\n", (float)k*0.01, reason);
 
-  send_to_char(ch, "You awarded %0.2f karma to %s for %s.\r\n", (float)k*0.01, GET_CHAR_NAME(vict), reason);
+  send_to_char(ch, "You awarded %0.2f karma to %s for %s\r\n", (float)k*0.01, GET_CHAR_NAME(vict), reason);
 
-  snprintf(buf2, sizeof(buf2), "%s awarded %0.2f karma to %s for %s (%d to %d).",
+  snprintf(buf2, sizeof(buf2), "%s awarded %0.2f karma to %s for %s (%d to %d)",
```
luciensadi, line 2542: Log message, please revert.

src/act.wizard.cpp
```diff
@@ -2535,11 +2535,11 @@ ACMD(do_award)
   else
     gain_karma(vict, k, TRUE, FALSE, FALSE);
 
-  send_to_char(vict, "You have been awarded %0.2f karma for %s.\r\n", (float)k*0.01, reason);
+  send_to_char(vict, "You have been awarded %0.2f karma for %s\r\n", (float)k*0.01, reason);
 
-  send_to_char(ch, "You awarded %0.2f karma to %s for %s.\r\n", (float)k*0.01, GET_CHAR_NAME(vict), reason);
+  send_to_char(ch, "You awarded %0.2f karma to %s for %s\r\n", (float)k*0.01, GET_CHAR_NAME(vict), reason);
```
luciensadi, line 2540: These two send_to_chars can also accept the inline punctuation checker. Something like `"You awarded %0.2f karma to %s for %s%s\r\n", [...], ispunct(reason[strlen(reason) - 2]) ? "" : ".");`

luciensadi (changes requested):
Thanks for the PR!

src/act.other.cpp
UndeadNappist: This is the syntax feedback message that's given when you type syspoints with incorrect/no args, not the actual message the logs, player, or staff member receives upon successful action.

It's also why the punctuation wasn't changed here.

It's there to remind staff that "for" gets inserted whether you include it or not (and I often created variations of that exact message you listed by accident multiple times).

I'm open to less awkward solutions or ideas.

src/act.other.cpp
UndeadNappist: Nope, that was a mistake on my part. Corrected.

src/act.other.cpp
UndeadNappist: Will lead to messages of "staff awarded x syspoints to player for thus and such !!!!!."

Thought I was helping future readability.

src/act.wizard.cpp
UndeadNappist: Syntax reminder for staff, does not actually affect the logging or input line.

src/act.other.cpp
luciensadi: Oh, I see what you wanted to do here. Things in brackets like this traditionally mean optional arguments, so the meaning of this would be "you can write `for` here or not, whichever you prefer". Unless Epi / Vile have strong opinions, I think we should just revert this and make sure staff know that 'for' is automatically included.

UndeadNappist: Epiphany has requested including for, sans brackets, and that is what I have done.

luciensadi:
The punctuation changes look good, thanks! Let's take out the `for` changes though, changing that syntax is part of a larger rewrite to auto-add that word if it's not present and I don't think either of us wants to tackle that before the next patch.

UndeadNappist:
It already autoadds the word regardless of whether it is present or not.

luciensadi:
Right, but that's known behavior. Until we can write it to intelligently add or not add 'for' to the output, leaving it as-is is fine. The syntax string you've written hints that the user should write `sysp award X Y for Z` which will output `Staffmember has awarded you Y syspoints for for Z`. Removing the auto-added 'for' isn't preferable right now since I'm already used to the existing syntax.

GarbageJet:
Leave the "for." Make sure there's something in there to remind me not to type it a second time. That's my opinion.

Oh my god do I hate the automatic periods in that and vemote though.

UndeadNappist:
Stripped the for out, left the syntax and auto-insertion as is. The hubbub conversation should be enough to remind everyone anyhow, so job done either way.

PR #599, opened 2022-02-05
Update comm.cpp

UndeadNappist:
Theoretically should throw current vision penalties into the prompt, although it ignores such things as defender's invisibility, stealth, etcetrta.

### DIFF
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -1410,6 +1410,9 @@ int make_prompt(struct descriptor_data * d)
               else
                 strcpy(str, "@v");
               break;
+            case 'V':       // Vision penalties in theory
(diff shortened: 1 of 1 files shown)


src/comm.cpp
```diff
@@ -1410,6 +1410,9 @@ int make_prompt(struct descriptor_data * d)
               else
                 strcpy(str, "@v");
               break;
+            case 'V':       // Vision penalties in theory
+              sprintf(str, sizeof(str), "%d", calculate_vision_penalty(d->character));
```
luciensadi, line 1414: This should be an `snprintf` (sprintf is not memory-safe, and I'm working to remove it from the game where I see it). Other than that, yeah, this should work.

PR #600, opened 2022-02-06
Implements houserule for flee attempts

UndeadNappist:
Koborokuru have a normal x3 multiplier for speed, so that's been corrected in fight.cpp as well.

I'm not convinced I wrote the logic right for flee tests.

### DIFF
--- a/src/act.offensive.cpp
+++ b/src/act.offensive.cpp
@@ -597,19 +597,18 @@ bool passed_flee_success_check(struct char_data *ch) {
 
   if (!can_hurt(ch, FIGHTING(ch), TRUE, 0))
     return TRUE;
-
-  if success_test(GET_QUI(ch), GET_QUI(FIGHTING(ch)) + racial_flee_modifier) > 0 {
-    switch (GET_RACE(ch)) {
-      case RACE_SATYR:
-       racial_flee_modifier--;
-       break;
-      case RACE_DWARF:
-      case RACE_GNOME:
-      case RACE_MENEHUNE:
-        racial_flee_modifier++;
-        break;
-        }
-    return TRUE;
+  
+  switch (GET_RACE(ch)) {
+    case RACE_SATYR:
+      racial_flee_modifier--;
+      break;
+    case RACE_DWARF:
+    case RACE_GNOME:
+    case RACE_MENEHUNE:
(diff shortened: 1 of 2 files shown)


luciensadi:
Thanks for the PR! This code won't actually have the intended effect, it removes the penalty from the attacker but not the defender, and the change to act.offensive.cpp happens after the return statement, so the code is never executed.

I think it'd help if we got you set up to compile the game so you could see the compiler warnings that are generated. What operating system are you running?

UndeadNappist:
I figured it was something like that.

Win10, 64 bit.

src/act.offensive.cpp
```diff
@@ -598,7 +598,19 @@ bool passed_flee_success_check(struct char_data *ch) {
   if (!can_hurt(ch, FIGHTING(ch), TRUE, 0))
     return TRUE;
 
-  return success_test(GET_QUI(ch), GET_QUI(FIGHTING(ch))) > 0;
+  if success_test(GET_QUI(ch), GET_QUI(FIGHTING(ch)) + racial_flee_modifier) > 0
```
luciensadi, line 601: If-statements need to have their conditions enclosed in parens like `if (x)`, and almost always should be written as block statements:
```c++
if (x) {
  your code here
}```

File: src/act.offensive.cpp [hunk 2]
```diff
@@ -598,7 +598,19 @@ bool passed_flee_success_check(struct char_data *ch) {
...
+    switch (GET_RACE(ch)) {
+      case RACE_SATYR:
+       racial_flee_modifier--;
```
luciensadi, line 604: You're changing racial_flee_modifier after it's used in the success test, so this code won't do anything. The switch statement will need to come before the success_test call.

src/act.offensive.cpp
```diff
@@ -598,7 +598,19 @@ bool passed_flee_success_check(struct char_data *ch) {
...
+       break;
+      case RACE_DWARF:
+      case RACE_GNOME:
+      case RACE_MENEHUNE:
+        racial_flee_modifier++;
+        break;
+        }
+    return TRUE;
```
luciensadi, line 612: This statement has fallen outside of the if-check due to not using a block statement, so the effect of this code is to always return true and guarantee flee success!

PR #602, opened 2022-02-13
Update newshop.cpp

UndeadNappist:
Sets packaged up 'ware to be automatically kept due to how expensive it is.

Would do the same for other bought items, but not exactly sure how to do it more eloquently than copy-pasting the same line in fifteen different places.


luciensadi:
Thanks for the PR!

PR #604, opened 2022-02-14
Update act.informative.cpp

UndeadNappist:
Different descriptions based on whether or not you're in a vehicle.

### DIFF
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -1713,8 +1713,15 @@ void look_at_room(struct char_data * ch, int ignore_brief, int is_quicklook)
       }
     } else {
       if (weather_info.sky >= SKY_RAINING) {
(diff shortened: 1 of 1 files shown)


src/act.informative.cpp
```diff
@@ -1713,16 +1713,16 @@ void look_at_room(struct char_data * ch, int ignore_brief, int is_quicklook)
       }
     } else {
       if (weather_info.sky >= SKY_RAINING) {
-          if(ch->in_room->vehicle == !VEH_BIKE) {
+          if(ch->in_veh && ch->in_veh->type == !VEH_BIKE) {
```
luciensadi, line 1716: Almost! This needs to be `!=` instead of `== !`. The former is 'not equal' and the latter is 'equal to not'. VEH_BIKE is defined as a number, and I don't know what the negation of a number is, but it'd probably be a bug.

PR #606, opened 2022-02-23
Vile's Fingerless Hobo Glove Spells Update

TheVilest:
New Spells, coded terribly, by none other than yours truly.

Love,
Vile


src/spec_assign.cpp
```diff
@@ -344,6 +344,10 @@ struct spell_trainer spelltrainers[] = {
 //{60602, SPELL_DETECTOBJECT, 0, 6},
   {60602, SPELL_COMBATSENSE, 0, 6},
   {60602, SPELL_MINDLINK, 0, 6},
+  {60602, SPELL_MINDLINK, 0, 6},
```
luciensadi, line 347: Mindlink has been duplicated here.

src/newmagic.cpp
```diff
@@ -1708,6 +1708,39 @@ void cast_health_spell(struct char_data *ch, int spell, int sub, int force, char
...
+        send_to_char("Your eyes tingle as the shadows around you become clearer.\r\n", vict);
+        act("You successfully sustain that spell on $N.", FALSE, ch, 0, vict, TO_CHAR);
+      } else
+        send_to_char(FAILED_CAST, ch);
+      spell_drain(ch, spell, force, 0);
+      break;
+    case SPELL_INFRAVISION:
+      WAIT_STATE(ch, (int) (SPELL_WAIT_STATE_TIME));
+      success = success_test(skill, 6 + target_modifiers);
+      if (success > 0 && !AFF_FLAGGED(ch, AFF_INFRAVISION)) {
```
luciensadi, line 1725: IIRC you'll have to write in logic to apply / remove the aff_infra flag when the spell is cast and removed, I don't actually see anything that adds the flag (unless I just missed it?)

src/act.movement.cpp
```diff
@@ -72,7 +72,7 @@ int can_move(struct char_data *ch, int dir, int extra)
   if (IS_SET(extra, CHECK_SPECIAL) && special(ch, convert_dir[dir], &empty_argument))
     return 0;
 
-  if (ch->in_room && ch->in_room->icesheet[0] && !IS_ASTRAL(ch) && !IS_PROJECT(ch)) {
+  if (ch->in_room && ch->in_room->icesheet[0] && !IS_ASTRAL(ch) && !IS_PROJECT(ch) && !IS_AFFECTED(ch, AFF_LEVITATE)) {
```
luciensadi, line 75: You've used AFF_LEVITATE a lot, but I don't think it's been defined. It goes in awake.h with the other affs, and ~~you'll also need to add entries to constants.cpp to describe it when someone is statted etc~~ looks like you already did that part.

src/awake.h
```diff
@@ -698,7 +698,11 @@ enum {
 #define SPELL_THUNDERCLAP    63
 #define SPELL_WATERBOLT    64
 #define SPELL_SPLASH    65
-#define MAX_SPELLS    66
+#define SPELL_SPLASH    65
```
luciensadi, line 701: SPELL_SPLASH has been duplicated here.

src/constants.cpp
```diff
@@ -1909,6 +1910,8 @@ struct spell_types spells[] =
     { "Detect Magic", FALSE, DETECTION, AREA, -1, SUSTAINED, 0, LIGHT },
     { "Detect Object", TRUE, DETECTION, AREA, -1, SUSTAINED, 1, MODERATE },
     { "Mindlink", FALSE, DETECTION, SINGLE, -1, SUSTAINED, 0, SERIOUS },
+    { "Nightvision", TRUE, DETECTION, SINGLE, -1, SUSTAINED, 1, MODERATE },
```
luciensadi, line 1913: These have to be in the exact order they're written in awake.h, so Nightvision has to be after Splash etc.

luciensadi (changes requested):

src/newmagic.cpp
```diff
@@ -1708,6 +1708,36 @@ void cast_health_spell(struct char_data *ch, int spell, int sub, int force, char
...
+        act("You successfully sustain that spell on $N.", FALSE, ch, 0, vict, TO_CHAR);
+        create_sustained(ch, vict, spell, force, 0, success, drain);
+      } else
+        send_to_char(FAILED_CAST, ch);
+      break;
+    case SPELL_LEVITATE:
+      WAIT_STATE(ch, (int) (SPELL_WAIT_STATE_TIME));
+      success = success_test(skill, 4 + target_modifiers);
+      if (success > 0 && !AFF_FLAGGED(ch, AFF_LEVITATE)) {
+        send_to_char("Your feet gently lift off from the ground as you levitate.\r\n", vict);
```
luciensadi, line 1735: This one probably needs a room echo.

luciensadi (changes requested):
Looks great! Just one niggling change, and we'll be good to go.

PR #607, opened 2022-02-26
The Flame Wars Begin

TheVilest:
Everything changed when the Fire Nation attacked.

Vile's attempt at the Flame Aura spell, pretty straight forward.

### DIFF
--- a/src/newfight.cpp
+++ b/src/newfight.cpp
@@ -1186,6 +1186,33 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
     }
   }
 
+  // Handle spirits and elementals being little divas with their special combat rules.
+    // Namely: We require that the attack's power is greater than double the spirit's level, otherwise it takes no damage.
+    // If the attack's power is greater, subtract double the level from it.
+    if (IS_SPIRIT(def->ch) || IS_ANY_ELEMENTAL(def->ch)) {
+      if (att->melee->power <= GET_LEVEL(def->ch) * 2) {
+        bool target_died = 0;
+        target_died = damage(att->ch, def->ch, 0, att->melee->dam_type, att->melee->is_physical);
+
+        //Handle suprise attack/alertness here -- spirits melee.
+        if (!target_died) {
+          if (IS_NPC(def->ch)) {
+            if (AFF_FLAGGED(def->ch, AFF_SURPRISE))
+              AFF_FLAGS(def->ch).RemoveBit(AFF_SURPRISE);
+
+            GET_MOBALERT(def->ch) = MALERT_ALARM;
+            GET_MOBALERTTIME(def->ch) = 30;
+          }
+
+          // Process flame aura here since the spirit will otherwise end combat evaluation here.
+          handle_flame_aura(att, def);
+        }
+        return;
+      } else {
+        att->melee->power -= GET_LEVEL(def->ch) * 2;
+      }
+    }
+  
   int damage_total = convert_damage(staged_damage);
 
   {
--- a/src/newfight.cpp
+++ b/src/newfight.cpp
@@ -39,6 +39,8 @@ extern bool damage_without_message(struct char_data *ch, struct char_data *victi
 
 void engage_close_combat_if_appropriate(struct combat_data *att, struct combat_data *def, int net_reach);
 
+bool handle_flame_aura(struct combat_data *att, struct combat_data *def);
+
 bool does_weapon_have_bayonet(struct obj_data *weapon);
 
 #define IS_RANGED(eq)   (GET_OBJ_TYPE(eq) == ITEM_FIREWEAPON || \
@@ -1364,4 +1366,44 @@ void engage_close_combat_if_appropriate(struct combat_data *att, struct combat_d
       }
     }
   }
+}
+
+bool handle_flame_aura(struct combat_data *att, struct combat_data *def) {
+  // Now that we've counterstrike-swapped if applicable, check and try to apply flame aura.
+  if (AFF_FLAGGED(def->ch, AFF_FLAME_AURA) || MOB_FLAGGED(def->ch, MOB_FLAMEAURA)) {
+    int force;
+
+    // First, we require that the attacker really is fighting in melee and unarmed. Flame aura only burns body parts.
+    if (att->ranged_combat_mode || !att->weapon) {
+      // Flameaura-flagged NPCs just use their own level as the force.
+      if (MOB_FLAGGED(def->ch, MOB_FLAMEAURA)) {
+        force = GET_LEVEL(def->ch);
+      }
+      else {
+        // Iterate through their spells, find the flame aura that's applied to them, and extract its force.
+        for (struct sustain_data *sust = GET_SUSTAINED(def->ch); sust; sust = sust->next) {
+          if (!sust->caster && sust->spell == SPELL_FLAMEAURA) {
+            force = sust->force;
+            break;
+          }
+        }
+      }
+
+      if (GET_EQ(att->ch, WEAR_HANDS) && GET_EQ(att->ch, WEAR_FEET)) {
+        // TODO: Apply impact armor.
+      }
+
+      // TODO: Now that we've set the force, deal (Force)M damage to the attacker. Use TYPE_SUFFERING maybe?
+      // Remember that damage() returns true if it's killed them-- if this happens, you'll need to abort handling, because att->ch is already nulled out.
+      if (damage(att->ch, ...)) {
+        // TODO: send a message to the defender and the room. Remember not to use att->ch here since it's been nulled by death.
+
+        // Attacker died, so return TRUE here.
+        return TRUE;
+      }
+    }
+  }
+
+  // We survived! Return FALSE to indicate attacker did not die.
+  return FALSE;
 }
--- a/src/newfight.cpp
+++ b/src/newfight.cpp
@@ -309,7 +309,7 @@ struct melee_combat_data {
       }
       
       // Add +2 power to unarmed/melee, per MitS p147. -Vile
-      if (AFF_FLAGGED(ch, AFF_SPELL_FLAME_AURA)) {
+      if (AFF_FLAGGED(ch, AFF_SPELL_FLAME_AURA) || MOB_FLAGGED(ch, MOB_FLAMEAURA)) {
         power += 2;
       }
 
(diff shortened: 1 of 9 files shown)


src/newmagic.cpp
```diff
@@ -1739,6 +1739,32 @@ void cast_health_spell(struct char_data *ch, int spell, int sub, int force, char
       } else
         send_to_char(FAILED_CAST, ch);
       break;
+    case SPELL_FLAME_AURA:
+      if (!check_spell_victim(ch, vict, spell, arg))
+        return;
+
+      if (would_become_killer(ch, vict)) {
```
luciensadi, line 1746: I think you can chop out the PK parts, unless this spell has a harmful effect?

src/newmagic.cpp
```diff
@@ -1739,6 +1739,32 @@ void cast_health_spell(struct char_data *ch, int spell, int sub, int force, char
...
+      }
+
+      if (!IS_NPC(ch) && !IS_NPC(vict) && PLR_FLAGGED(ch, PLR_KILLER)) {
+        act("You have the KILLER flag, so you can't affect $N with an aggressive spell.", TRUE, ch, 0, vict, TO_CHAR);
+        return;
+      }
+      WAIT_STATE(ch, (int) (SPELL_WAIT_STATE_TIME));
+
+      success = success_test(skill, 4 + target_modifiers);
+      if (success > 0 && !AFF_FLAGGED(ch, AFF_FLAME_AURA)) {
```
luciensadi, line 1758: Checking to see if they already have the aff should come before this, and should give a specific message, like "They're already affected by that spell." See improved invis for an example of how this is done.

luciensadi (changes requested):
It looks like we're missing the part where people who hit a flame-aura person take (force)M damage, feel up to adding that in newfight.cpp's hit()? I think there's also already a MOB_FLAMEAURA flag should be checked for for both adding to combat damage and dealing damage to attackers.

src/newmagic.cpp
TheVilest: Wasn't sure if the intention was to make guard mobs aggro at anyone with a flame aura, in which case, casting it on someone else intentionally to aggro mobs could potentially be abused... But I guess someone could also cast Force 1 Invis on another player and accomplish the same potentially abusive tactic?

src/newmagic.cpp
TheVilest: Gotcha, check/message added.

luciensadi:
Started prototyping this out to see how I'd write it. My first pass ended up like this:

First, add this to the bottom of newfight.cpp, sticking `bool handle_flame_aura(struct combat_data *att, struct combat_data *def);` near the top so it's accessible to the code in the file.
```c
bool handle_flame_aura(struct combat_data *att, struct combat_data *def) {
  // Now that we've counterstrike-swapped if applicable, check and try to apply flame aura.
  if (AFF_FLAGGED(def->ch, AFF_FLAME_AURA) || MOB_FLAGGED(def->ch, MOB_FLAMEAURA)) {
    int force;

    // First, we require that the attacker really is fighting in melee and unarmed. Flame aura only burns body parts.
    if (att->ranged_combat_mode || !att->weapon) {
      // Flameaura-flagged NPCs just use their own level as the force.
      if (MOB_FLAGGED(def->ch, MOB_FLAMEAURA)) {
        force = GET_LEVEL(def->ch);
      }
      else {
        // Iterate through their spells, find the flame aura that's applied to them, and extract its force.
        for (struct sustain_data *sust = GET_SUSTAINED(def->ch); sust; sust = sust->next) {
          if (!sust->caster && sust->spell == SPELL_FLAMEAURA) {
            force = sust->force;
            break;
          }
        }
      }

      if (GET_EQ(att->ch, WEAR_HANDS) && GET_EQ(att->ch, WEAR_FEET)) {
        // TODO: Apply impact armor.
      }

      // TODO: Now that we've set the force, deal (Force)M damage to the attacker. Use TYPE_SUFFERING maybe?
      // Remember that damage() returns true if it's killed them-- if this happens, you'll need to abort handling, because att->ch is already nulled out.
      if (damage(att->ch, ...)) {
        // TODO: send a message to the defender and the room. Remember not to use att->ch here since it's been nulled by death.

        // Attacker died, so return TRUE here.
        return TRUE;
      }
    }
  }

  // We survived! Return FALSE to indicate attacker did not die.
  return FALSE;
}
```

Next, add in calls to handle_flame_aura where damage is dealt. In our case, we care about damage to spirits (approx line 1150-ish):
```c
// Handle spirits and elementals being little divas with their special combat rules.
    // Namely: We require that the attack's power is greater than double the spirit's level, otherwise it takes no damage.
    // If the attack's power is greater, subtract double the level from it.
    if (IS_SPIRIT(def->ch) || IS_ANY_ELEMENTAL(def->ch)) {
      if (att->melee->power <= GET_LEVEL(def->ch) * 2) {
        bool target_died = 0;
        target_died = damage(att->ch, def->ch, 0, att->melee->dam_type, att->melee->is_physical);

        //Handle suprise attack/alertness here -- spirits melee.
        if (!target_died) {
          if (IS_NPC(def->ch)) {
            if (AFF_FLAGGED(def->ch, AFF_SURPRISE))
              AFF_FLAGS(def->ch).RemoveBit(AFF_SURPRISE);

            GET_MOBALERT(def->ch) = MALERT_ALARM;
            GET_MOBALERTTIME(def->ch) = 30;
          }

          // Process flame aura here since the spirit will otherwise end combat evaluation here.
          handle_flame_aura(att, def);
        }
        return;
      } else {
        att->melee->power -= GET_LEVEL(def->ch) * 2;
      }
    }
```

And to characters (approx line 1250-ish):
```c
bool defender_died, defender_was_npc = IS_NPC(def->ch);
  if (att->ranged_combat_mode) {
    combat_message(att->ch, def->ch, att->weapon, MAX(0, damage_total), att->ranged->burst_count);
    defender_died = damage_without_message(att->ch, def->ch, damage_total, att->ranged->dam_type, att->ranged->is_physical);

    if (!defender_died && damage_total > 0)
      perform_knockdown_test(def->ch, (GET_WEAPON_POWER(att->weapon) + att->ranged->burst_count) / (att->ranged->is_gel ? 1 : 2));
  } else {
    // Process flame aura right before damage.
    if (handle_flame_aura(att, def)) {
      // TODO: Print a message to the defender and the room that the attacker burned themselves up, then bail out.
      return;
    }

    defender_died = damage(att->ch, def->ch, damage_total, att->melee->dam_type, att->melee->is_physical);
    ...
```

Once those are in place, I _think_ that'll take care of the damage dealing part. We'll have to test it of course.

src/newfight.cpp
```diff
@@ -305,6 +307,11 @@ struct melee_combat_data {
       if (cyber->cyberarms) {
         power += 2;
       }
+      
+      // Add +2 power to unarmed/melee, per MitS p147. -Vile
+      if (AFF_FLAGGED(ch, AFF_SPELL_FLAME_AURA)) {
```
luciensadi, line 312: Since we're adding support for mob flame aura, let's change this to:
`if (AFF_FLAGGED(ch, AFF_SPELL_FLAME_AURA) || MOB_FLAGGED(ch, MOB_FLAMEAURA))`

src/newfight.cpp
```diff
@@ -1179,6 +1186,33 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
     }
   }
 
+  // Handle spirits and elementals being little divas with their special combat rules.
```
luciensadi, line 1189: This block should have replaced the previous one-- search for 'divas' and plop it in on top of that old block. Down here, it doesn't really affect anything.

src/newfight.cpp
```diff
@@ -1259,6 +1293,23 @@ void hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
...
+  if (att->ranged_combat_mode) {
+    combat_message(att->ch, def->ch, att->weapon, MAX(0, damage_total), att->ranged->burst_count);
+    defender_died = damage_without_message(att->ch, def->ch, damage_total, att->ranged->dam_type, att->ranged->is_physical);
+
+    if (!defender_died && damage_total > 0)
+      perform_knockdown_test(def->ch, (GET_WEAPON_POWER(att->weapon) + att->ranged->burst_count) / (att->ranged->is_gel ? 1 : 2));
+  } else {
+    // Process flame aura right before damage.
+    if (handle_flame_aura(att, def)) {
+      // TODO: Print a message to the defender and the room that the attacker burned themselves up, then bail out.
```
luciensadi, line 1306: Fill this in with something cool 😁

src/newfight.cpp
```diff
@@ -1360,3 +1411,43 @@ void engage_close_combat_if_appropriate(struct combat_data *att, struct combat_d
...
+        // Iterate through their spells, find the flame aura that's applied to them, and extract its force.
+        for (struct sustain_data *sust = GET_SUSTAINED(def->ch); sust; sust = sust->next) {
+          if (!sust->caster && sust->spell == SPELL_FLAMEAURA) {
+            force = sust->force;
+            break;
+          }
+        }
+      }
+
+      if (GET_EQ(att->ch, WEAR_HANDS) && GET_EQ(att->ch, WEAR_FEET)) {
```
luciensadi, line 1436: You'd mentioned wanting to use adept powers / spells / etc instead of checking for hands / feet wearslots, this would be the spot to do it in.

src/newfight.cpp
```diff
@@ -1360,3 +1411,43 @@ void engage_close_combat_if_appropriate(struct combat_data *att, struct combat_d
...
+        // TODO: Apply impact armor.
+      }
+
+      // TODO: Now that we've set the force, deal (Force)M damage to the attacker. Use TYPE_SUFFERING maybe?
+      // Remember that damage() returns true if it's killed them-- if this happens, you'll need to abort handling, because att->ch is already nulled out.
+      if (damage(att->ch, ...)) {
```
luciensadi, line 1442: You'll want to fill this in with the proper variables etc.

src/newfight.cpp
```diff
@@ -1360,3 +1411,43 @@ void engage_close_combat_if_appropriate(struct combat_data *att, struct combat_d
...
+        // TODO: send a message to the defender and the room. Remember not to use att->ch here since it's been nulled by death.
```
luciensadi, line 1443: Cool message here too! Just send it to the defender and to the room instead of directly to the attacker, since the attacker at this point is dead (RIP).

luciensadi (changes requested):
Thanks for the changes! It looks like there are a few TODOs unresolved, let me know if you'd like a hand with those.

src/newfight.cpp
TheVilest: Done

PR #612, opened 2022-03-04
Levitate Fix

TheVilest:
Update to my broken levitate code so it actually works


luciensadi:
Thanks for this! The reason it wasn't working was the spell was filed in Illusion instead of Manipulation, so the cast-manipulation code could not find it. I went ahead and fixed it, and also added drain and other safeguards in the process. Appreciate the PR!

PR #614, opened 2022-03-11
Update axehead's speeches

UndeadNappist:
Updates some of axehead's speeches a little to lead to less item destruction.

### DIFF
--- a/src/spec_procs.cpp
+++ b/src/spec_procs.cpp
@@ -6062,7 +6062,7 @@ SPECIAL(axehead) {
   int message_num;
   const char *axehead_messages[] = {
     "Runners these days don't realize how valuable keeping notes on their pocket secretary is. Like where Johnsons hang out.",
-    "Don't hide your radio or phone in a box, you won't be able to hear them - but putting it right in your coat pocket won't do much but break it if you get shot.",
+    "Best to keep your phone and radio in hand. In a box, you won't hear it, and in your pocket, it'll just get shot.",
     "Seems like every day I hear about another wanna-be runner getting gunned down by the Star for walking around with their gun or cyberdeck in hand.",
     "Back in my day, we didn't have anything like the 8 MHz band available. Being able to talk to runners is a blessing.",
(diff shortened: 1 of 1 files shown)


src/spec_procs.cpp
```diff
@@ -6062,12 +6062,13 @@ SPECIAL(axehead) {
   int message_num;
   const char *axehead_messages[] = {
     "Runners these days don't realize how valuable keeping notes on their pocket secretary is. Like where Johnsons hang out.",
-    "Stick your radio and phone in a pocket. You can still hear 'em, and it keeps your hands free.",
-    "Seems like every day I hear about another wanna-be runner getting gunned down by the Star for walking around with their gun in their hand.",
+    "Don't hide your radio or phone in a box, you won't be able to hear them - but putting it right in your coat pocket won't do much but break it if you get shot.",
```
luciensadi, line 6065: Maybe "Best to keep your phone and radio in hand. In a box, you won't hear it, and in your pocket, it'll just get shot."

src/spec_procs.cpp
```diff
@@ -6062,12 +6062,13 @@ SPECIAL(axehead) {
...
+    "Seems like every day I hear about another wanna-be runner getting gunned down by the Star for walking around with their gun or cyberdeck in hand.",
     "Back in my day, we didn't have anything like the 8 MHz band available. Being able to talk to runners is a blessing.",
     "When in doubt, just take a cab back to somewhere familiar.",
     "If you're on a job and you just can't get it done, call your Johnson and tell them you quit. Easier than hoofing it all the way back.",
-    "It's dangerous to go alone. Make friends."
+    "It's dangerous to go alone. Make friends.",
+    "Keep your eyes and ears open. You never know when you'll make an unfair weather friend."
```
luciensadi, line 6071: Since you've added a message, you also need to increment the NUM_AXEHEAD_MESSAGES below, otherwise this will never be said.

luciensadi:
Looks good, thank you!

PR #615, opened 2022-03-13
Vision overhaul

luciensadi:


lib/text/bugs
```diff
@@ -198,3 +198,4 @@ Aellyn   (Jan  1) [31102] mortally wounded characters can counterstrike and hit
 Shadow   (Jul 10) [ 3474] I remember being able to get a snowflake downloaded from here in JNL to unlock the door to the next room but the files don't seem to be spawning
 Armitage (Sep 23) [ 2961] underflow in points-based creation if you play a race that uses more than what you put into attributes. I made a troll with 6 in attributes and had 65k training points (I deleted and remade with the 11 attribute points needed but most people wouldn't, probably)
 Lucien   (Jul  5) [35501] release all gave no reply when I used it
+Lucien   (Mar 13) [60611] equipment flag from aff_thermo etc stays on you from cyberware/bioware even after it's sold
```
luciensadi, line 201: Need to fix this, then remove this line from the bugs file-- we don't want to edit this file if possible.

src/fight.cpp
```diff
@@ -3921,8 +3921,8 @@ int calculate_vision_penalty(struct char_data *ch, struct char_data *victim) {
   }
 
   // Pre-calculate the things we care about here. First, character vision info.
-  bool ch_has_ultrasound = AFF_FLAGGED(ch, AFF_ULTRASOUND);
-  bool ch_has_thermographic = AFF_FLAGGED(ch, AFF_INFRAVISION) || CURRENT_VISION(ch) == THERMOGRAPHIC;
+  bool ch_has_ultrasound = AFF_FLAGGED(ch, AFF_ULTRASOUND) || has_vision(ch, VISION_ULTRASONIC);
```
luciensadi, line 3924: I want to get rid of AFF_ULTRASOUND here if I can, or at least have AFF_ULTRASOUND consumed behind the scenes to grant VISION_ULTRASONIC. Removing this and the other aff flags from the logic will simplify things.

src/handler.cpp
```diff
@@ -709,24 +712,37 @@ void affect_total(struct char_data * ch)
...
+        break;
+      case CYB_DERMALSHEATHING:
+      case CYB_DERMALPLATING:
+        // Ruthenium sheathing.
+        if (GET_CYBERWARE_TYPE(cyber) == CYB_DERMALSHEATHING && GET_OBJ_VAL(cyber, 3) == 1 && !wearing)
+          AFF_FLAGS(ch).SetBit(AFF_INVISIBLE);
+        // todo
+        break;
+      case CYB_EYES:
+        apply_vision_bits_from_cyberware(ch, cyber);
```
luciensadi, line 736: Do we need to call `remove_racial_vision_due_to_eye_replacement(ch);` here?

src/handler.cpp
```diff
@@ -992,12 +1018,12 @@ void affect_total(struct char_data * ch)
...
-  else if (AFF_FLAGGED(ch, AFF_LOW_LIGHT))
-    CURRENT_VISION(ch) = LOWLIGHT;
-  else
-    CURRENT_VISION(ch) = NATURAL_VISION(ch);
+  if (AFF_FLAGGED(ch, AFF_INFRAVISION)) {
+    set_vision_bit(ch, VISION_THERMOGRAPHIC, VISION_BIT_FROM_EQUIPMENT);
+  }
+  if (AFF_FLAGGED(ch, AFF_LOW_LIGHT)) {
+    set_vision_bit(ch, VISION_LOWLIGHT, VISION_BIT_FROM_EQUIPMENT);
+  }
```
luciensadi, line 1026: Add ultrasonic vision here?

src/utils.cpp
```diff
@@ -240,18 +255,25 @@ int light_level(struct room_data *room)
       break;
   }
 
+  // Indoor rooms have no weather or time-based light influences.
   if (room->sector_type == SPIRIT_HEARTH)
```
luciensadi, line 259: Why are we using SPIRIT_HEARTH to check for indoor rooms instead of `ROOM_FLAGGED(room, ROOM_INDOORS)`?

src/utils.cpp
```diff
@@ -3499,21 +3393,18 @@ bool LIGHT_OK_ROOM_SPECIFIED(struct char_data *sub, struct room_data *provided_r
...
 
-      // TODO: Have the shadow spell on them?
-    }
-  }
+  // At this point, we know that we're using ultrasonic, low-light, or normal vision.
+  // Ultrasonic can see in any light condition. LL can see in minlight and better. Normal is the most restricted.
+  if (room_light_level == LIGHT_FULLDARK)
+    return has_vision(sub, VISION_ULTRASONIC);
+  if (room_light_level == LIGHT_MINLIGHT)
+    return has_vision(sub, VISION_LOWLIGHT);
```
luciensadi, line 3404: or ultrasonic

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+#include "structs.h"
+#include "comm.h"
+#include "constants.h"
+
+bool _vision_prereqs_are_valid(struct char_data *ch, int type, const char *function_name);
+bool _vision_bit_is_valid(int bit, const char *function_name);
+
+// Do we have a given type of vision at all?
+bool has_vision(struct char_data *ch, int type, bool staff_override) {
+  if (!_vision_prereqs_are_valid(ch, type, "has_vision"))
```
luciensadi, line 16: Instead of hardcoding function names, try using `__func__` and seeing if that works.

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+  // All other vision types block spellcasting. This specifically includes:
+  // - Electronic means of vision (SR3 p181)
+  // - Spell vision (same page)
+  // - Adept powers (adepts can't cast)
+  // - Equipment-based vision that you don't already have (LL/IR sensors are still electronic imaging)
+  return FALSE;
+}
+
+void remove_racial_vision_due_to_eye_replacement(struct char_data *ch) {
+  switch (GET_RACE(ch)) {
```
luciensadi, line 62: This needs to handle the Cat's Eye bioware case, which grants natural vision even though it's not technically natural. It would be lost under the current code.

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+  if (!_vision_prereqs_are_valid(ch, type, "set_vision_bit"))
+    return;
+
+  if (!_vision_bit_is_valid(bit, "set_vision_bit"))
+    return;
+
+  // Eye replacements strip racial benefits, so if we have those and are setting the implant flag, strip them.
+  // See M&M p64 (Cat's Eyes) for an example of this rule.
+  if (bit == VISION_BIT_FROM_IMPLANTS) {
+
```
luciensadi, line 99: I see a little silhouetto of a gap-- SCARAMOUCHE, SCARAMOUCHE! would you fill the code in please?

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+    return;
+
+  if (!_vision_bit_is_valid(bit, "remove_vision_bit"))
+    return;
+
+  ch->points.vision[type].RemoveBit(bit);
+}
+
+
+/////////////////// Utilities and helpers for other files //////////////////////
```
luciensadi, line 116: Comment placement confusing, makes it seem like these penalty defines are available elsewhere.

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+      if (tn_with_none > 0) {
+        int new_normal = MAX(0, tn_with_none - temp_room->light[2]);
+        if (new_normal != tn_with_none) {
+          NM_PENALTY("LightSpell[NM]", new_normal - tn_with_none);
+        }
+      }
+    } else {
+      // This happens when you roll no successes at all, right?
+      // light_target /= 2;
+      mudlog("Encountered that weird light case, investigate here!", ch, LOG_SYSLOG, TRUE);
```
luciensadi, line 247: Expand this message as much as we can to include details to help with debugging.

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+  if (temp_room->vision[1] == LIGHT_MIST) {
+    if (!has_natural_lowlight_vision) {
+      LL_PENALTY("Mist[LL-U]", 2);
+    }
+
+    NM_PENALTY("Mist[NM]", 2);
+  }
+
+  if (temp_room->vision[1] == LIGHT_LIGHTSMOKE
+      || (weather_info.sky == SKY_RAINING && temp_room->sector_type != SPIRIT_HEARTH && !ROOM_FLAGGED(temp_room, ROOM_INDOORS)))
```
luciensadi, line 260: SPIRIT_HEARTH again. Are there non-indoor rooms that have hearth and therefore shouldn't receive weather?

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+  else if (has_vision(ch, VISION_LOWLIGHT) && tn_with_ll <= tn_with_thermo && tn_with_ll <= tn_with_none) {
+    penalty_vector = &lowlight_penalties;
+    penalty_chosen = tn_with_ll;
+  }
+  else {
+    penalty_vector = &normal_penalties;
+    penalty_chosen = tn_with_none;
+  }
+
+  for (auto it = penalty_vector->begin(); it != penalty_vector->end(); ++it) {
```
luciensadi, line 316: Don't we have access to range-based for loops here? `for (auto it : penalty_vector)`

src/vision_overhaul.cpp
```diff
@@ -0,0 +1,455 @@
...
+
+  // All other vision types block spellcasting. This specifically includes:
+  // - Electronic means of vision (SR3 p181)
+  // - Spell vision (same page)
+  // - Adept powers (adepts can't cast)
+  // - Equipment-based vision that you don't already have (LL/IR sensors are still electronic imaging)
+  return FALSE;
+}
+
+void remove_racial_vision_due_to_eye_replacement(struct char_data *ch) {
```
luciensadi, line 61: `remove_vision_bit(ch, VISION_NORMAL, VISION_BIT_IS_NATURAL);`

Do we need to stick in a new bit set for it though? I don't think we ever actually check VISION_NORMAL in the code right now, so it's probably a moot point, but in terms of game logic correctness we probably need to add a bit here.

src/handler.cpp
luciensadi: Ans: No, it's done in apply_vision_bits_from_cyberware().

luciensadi:
Remaining work to do:
- Fix the bug that's causing AFF_THERMO etc to stick around even after you sell the cyberware causing it. This is probably a mix of two things-- first, cyberware should lose the AFF_X flags that cause vision, since they now cause it innately; and second, the shop code should be smart enough to strip these flags.
- Go through and rip out as many instances of AFF_INFRA, AFF_LOW, AFF_ULTRA as you can. In theory, they're all consumed in point updates, making the logic that checks for them redundant at best and bug-inducing at worst.

luciensadi:
Remaining work to do:

- Go through and rip out as many instances of AFF_INFRA, AFF_LOW, AFF_ULTRA as you can. In theory, they're all consumed in point updates, making the logic that checks for them redundant at best and bug-inducing at worst.
- Fix the issue that's causing newly-loaded chars to have natural vision until their first affect_total tick.
- Clarify score display of vision.
- Decide on adding flare compensation and magnification levels to vision readout.

luciensadi:
I think I've implemented everything I planned to here! Time for a full review.

PR #616, opened 2022-03-27
Cranial Remote Deck Cyberware (CRD)

TheVilest:
You down with CRD? Yeah you know me~

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -890,7 +890,7 @@ ACMD(do_control)
 
   if (!has_rig)
     for (cyber = ch->cyberware; cyber; cyber = cyber->next_content)
-      if (GET_CYBERWARE_TYPE == CYB_CRD))
+      if (GET_CYBERWARE_TYPE(cyber, 0) == CYB_CRD))
         has_rig = TRUE;
 
   if (!has_rig) {
@@ -976,7 +976,7 @@ ACMD(do_subscribe)
 
   if (!has_deck)
   for (struct obj_data *cyber = ch->cyberware; cyber; cyber = cyber->next_content)
-    if (GET_CYBERWARE_TYPE == CYB_CRD)
+    if (GET_CYBERWARE_TYPE(cyber, 0) == CYB_CRD)
       has_rig = TRUE;
 
   if (!has_deck) {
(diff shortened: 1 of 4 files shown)


src/act.drive.cpp
```diff
@@ -888,6 +888,11 @@ ACMD(do_control)
         if (GET_OBJ_TYPE(GET_EQ(ch, x)) == ITEM_RCDECK)
           has_rig = TRUE;
 
+  if (!has_rig)
+    for (cyber = ch->cyberware; cyber; cyber = cyber->next_content)
+      if (GET_OBJ_VAL(cyber, 0) == CYB_CRD))
```
luciensadi, line 893: `GET_CYBERWARE_TYPE` instead of the raw obj_val call, please!

src/act.drive.cpp
```diff
@@ -955,7 +960,7 @@ ACMD(do_control)
 ACMD(do_subscribe)
 {
   struct veh_data *veh = NULL;
-  struct obj_data *obj;
+  struct obj_data *obj, *cyber;
```
luciensadi, line 963: If you're not referencing this value anywhere other than the for loop, you can just initialize it there.
`for (struct obj_data *cyber = ch->cyberware;...`

src/db.cpp
```diff
@@ -6286,6 +6286,22 @@ void price_cyber(struct obj_data *obj)
           break;
       }
       break;
+    case CYB_CRD:
+      GET_CYBERWARE_ESSENCE_COST(obj) = 30;
+      GET_OBJ_AVAILTN(obj) = 4;
+      GET_OBJ_AVAILDAY(obj) = 3;
+      switch (GET_OBJ_VAL(obj, 1)) {
```
luciensadi, line 6293: `GET_CYBERWARE_RATING` please

luciensadi (changes requested):
Looks pretty tight! Few nits and this should be good to go.

src/act.drive.cpp
TheVilest: That's pretty useful info going forward

src/act.drive.cpp
TheVilest: Is the `(cyber, 0)` value reference still needed to reference type?

src/db.cpp
TheVilest: Gotcha

src/act.drive.cpp
TheVilest: I'm assuming not, since utils.h has a convenience definition

PR #617, opened 2022-03-29
Fixes magic circle cost

UndeadNappist:
As per sr3, page 167. I'll update the helpfile later.

### DIFF
--- a/src/newmagic.cpp
+++ b/src/newmagic.cpp
@@ -2794,7 +2794,7 @@ bool check_spirit_sector(struct room_data *room, int spirit)
 
 void circle_build(struct char_data *ch, char *type, int force)
 {
-  long cost = force * force;
+  long cost = (force * force) * 100;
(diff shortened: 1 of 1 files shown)


luciensadi:
Where did you get the *100 from? What I'm reading on p167 is:
```The cost of creating a circle is negligible—its rating squared in nuyen for the paint or chalk.```

UndeadNappist:
image

luciensadi:
> \> exam circle
> Hermetic circles are the corner stone of a mage's summoning ritual. It is an elaborate design, drawn in chalk that takes up most of the floor space in the room.

Thanks for the PR, but this doesn't seem to be in line with the existing circles. If you want to try your hand at writing permanent circles, we could talk about that!

UndeadNappist:
Full on permanency seems to be more of a factor of PGHQs and not for casual construction.

PR #619, opened 2022-04-07
Replaces catch-all truck skill

UndeadNappist:
Each archetype now has the specific driving skill for their titles.


src/archetypes.cpp
```diff
@@ -151,6 +152,7 @@ struct archetype_data *generate_adept() {
   arch->skills[SKILL_PILOT_BIKE] = 1;
   arch->skills[SKILL_BR_BIKE] = 1;
   arch->skills[SKILL_STREET_ETIQUETTE] = 1;
+  arch->skills[SKILL_PILOT_BIKE] = 1;
```
luciensadi, line 155: This archetype already has the bike skill, increase it to 2?

luciensadi:
Thanks for the PR!

PR #622, opened 2022-04-17
Prep for drone only exit

UndeadNappist:
Make a new room, flag as ROOM_FREEWAY and shove an astral barrier in the way. Make another exit one-way to the neophyte parking lot. You now have a one-way exit for the drone.

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -2230,8 +2230,10 @@ ACMD(do_transfer)
   two_arguments(argument, buf, buf2);
   struct veh_data *veh = NULL;
   struct char_data *targ = NULL;
-
-  if (!*buf || !*buf2)
+  
+  if(PLR_FLAGGED(ch, PLR_NOT_YET_AUTHED) && subcmd != SCMD_NEWBIE)
+    send_to_char("You can't transfer vehicles until authorized.\r\n", ch);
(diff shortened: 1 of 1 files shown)


luciensadi:
Drone-only exits have a lot of prerequisite work that need to be done to make them functional.
- A vehicle claim area in the neophyte guild that only allows claiming drone titles
- A way to ensure that you lose control of vehicles once they leave the guild
- A way to ensure that you can't resume control vehicles once they leave the guild
- A proc to drop the value of all items on the drone to zero once they pass that barrier

The need for this exit stems from weight limits in chargen, so why not just remove weight limits in chargen and give people a warning when they're exceeding them?

UndeadNappist:
> Drone-only exits have a lot of prerequisite work that need to be done to make them functional.
> 
> * A vehicle claim area in the neophyte guild that only allows claiming drone titles
> * A way to ensure that you lose control of vehicles once they leave the guild
> * A way to ensure that you can't resume control vehicles once they leave the guild
> * A proc to drop the value of all items on the drone to zero once they pass that barrier
> 
> The need for this exit stems from weight limits in chargen, so why not just remove weight limits in chargen and give people a warning when they're exceeding them?

Sounds good to me, I'll give that a shot.

PR #623, opened 2022-04-17
Increases weight limit for non-auth'd characters

UndeadNappist:
No idea if this will work, but it's a lot easier then adding an extra if statement to a dozen places.

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -608,8 +608,8 @@ int get_armor_penalty_grade(struct char_data *ch);
 #define GET_MOBALERT(ch)	((ch)->mob_specials.alert)
 #define GET_MOBALERTTIME(ch)	((ch)->mob_specials.alerttime)
 
-#define CAN_CARRY_W(ch)       ((GET_STR(ch) * 10) + 30)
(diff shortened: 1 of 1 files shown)


luciensadi:
This logic won't do what you're expecting it to do, especially the part where you're multiplying the is_authed check by 50. Are you able to test this on your local instance?

UndeadNappist:
I can, but I need to recompile. I was hoping that it would return a value like 1 or the define value itself and then I could multiply that directly; or if they weren't auth'd, it wouldn't return a value at all.

It's either that or have it evaluate as a number directly, with zero if they're authorized. But I'm not quite sure how to go about that.

luciensadi:
Thinking about this more, it would be simple enough to change the definition you're working with to grant extra carrying capacity to newbies, but the system also needs to be clarified to make it obvious that they have extra carrying capacity because they're still in chargen. I'm not sure what this looks like, but it probably includes a note in score and aff, as well as a warning that pops up when they purchase something that puts them over, and also a warning that they get when they exit the store complex.

PR #627, opened 2022-04-26
Bows and crossbows.

luciensadi:
Still need the ability to switch ammo types, and also a pockets-like system for these.


luciensadi:
Still a lot to do here:
- Have a chance for arrows to not break on impact, allowing you to retrieve them
- Check for gunfire noise and make sure it doesn't happen
- Add better ammo handling system
- Add other ammo types for bows/crossbows (they codedly exist but have no effects and no way to select them)
- Make firing the bow a complex action (simple to load, simple to fire)
- Look into crossbows and see if any rules need to change around those

luciensadi:
Closing as stale.

PR #629, opened 2022-05-02
Changes starting rooms

UndeadNappist:
Changes starting rooms to the correct locations. Room 90999 will need to be finished before this is merged.

Shamans and mages are pointed at the same area, so the room description could be tweaked up (or finishing up 90800 zone and then pointing shamans to there, lots of copy-paste).


luciensadi:
Thanks for this one!

PR #630, opened 2022-05-07
Helpfile feedback cleanup.

UndeadNappist:
Changes the helpfile fetch to respond with a list when there are more than 2 helpfiles instead of 5. Some of the helpfiles are quite beefy.


luciensadi:
I suspect this one is going to cause problems from a UI/UX perspective, how did it seem when you tried it out?

UndeadNappist:
Seems fine, fixes the issue of new folks getting spammed by multiple huge (and sometimes unrelated) helpfiles when looking for help emote.

Most helpfiles have been given wide enough name variety being unable to open them isn't an issue.

luciensadi:
Let's give it a shot, thanks!

PR #633, opened 2022-05-09
Renames Improved Attributes

UndeadNappist:
Standardizes the format and should make the pointers for the helpfiles a little easier.

I could also see stripping all the parentheses off for screenreaders and renaming them that way. I don't have any real preference.


luciensadi:
I suspect this one won't be a popular change, because all the people who are used to `acti improved str` now have to type `acti improved attribute (str`. Unless there's a compelling reason to add the extra word, brevity is the way to go here.

UndeadNappist:
The helpfile is named IMPROVED ATTRIBUTES and people repeatedly search for IMPROVED STRENGTH, IMPROVED QUICKNESS, etc. If I renamed it that it'd have a bunch of helpfiles that would pop up all at the same time, plus a huge helpfile name.

I was trying to figure out a decent way for consistency and to make it look cleaner, but Improved A (Strength, etc) might work too?

UndeadNappist:
Made another PR that fixed this issue more cleanly.

PR #635, opened 2022-05-17
Update act.drive.cpp

UndeadNappist:
Allows you to use melee weapons while on a bike.

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -1270,7 +1270,7 @@ ACMD(do_driveby)
   }
  
   if ((wielded = GET_EQ(ch, WEAR_WIELD))) {
-    if ((IS_OBJ_STAT(wielded, ITEM_EXTRA_TWOHANDS) || (!WEAPON_IS_GUN(wielded) && ch->in_veh->type = !VEH_BIKE))  {
+    if ((IS_OBJ_STAT(wielded, ITEM_EXTRA_TWOHANDS) || (!WEAPON_IS_GUN(wielded) && ch->in_veh->type = !VEH_BIKE)))  {
       send_to_char(ch, "You can only do a driveby with a one handed weapon in a vehicle and a melee weapon while on a bike!\r\n");
       return;
     }
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -1271,7 +1271,7 @@ ACMD(do_driveby)
  
   if ((wielded = GET_EQ(ch, WEAR_WIELD))) {
     if ((IS_OBJ_STAT(wielded, ITEM_EXTRA_TWOHANDS) || (!WEAPON_IS_GUN(wielded) && ch->in_veh->type = !VEH_BIKE))  {
-      send_to_char(ch, "You can only do a driveby with a one handed weapon in a vehicle and/or a melee weapon while on a bike!\r\n");
+      send_to_char(ch, "You can only do a driveby with a one handed weapon in a vehicle and a melee weapon while on a bike!\r\n");
       return;
     }
     if (!IS_NPC(vict) && !PRF_FLAGGED(ch, PRF_PKER) && !PRF_FLAGGED(vict, PRF_PKER)) {
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -1270,8 +1270,8 @@ ACMD(do_driveby)
   }
 
   if ((wielded = GET_EQ(ch, WEAR_WIELD))) {
-    if (!(GET_OBJ_VAL(wielded, 4) == SKILL_PISTOLS || GET_OBJ_VAL(wielded, 4) == SKILL_SMG)) {
-      send_to_char(ch, "You can only do a driveby with a pistol or SMG.\r\n");
+    if (!(GET_OBJ_VAL(wielded, 4) == SKILL_PISTOLS || GET_OBJ_VAL(wielded, 4) == SKILL_SMG || ((0 <= GET_OBJ_VAL(wielded, 3) <= 4)  && ch->in_veh->type = VEH_BIKE)))  {
+      send_to_char(ch, "You can only do a driveby with a pistol or SMG, or a melee weapon while on a bike.\r\n");
       return;
     }
     if (!IS_NPC(vict) && !PRF_FLAGGED(ch, PRF_PKER) && !PRF_FLAGGED(vict, PRF_PKER)) {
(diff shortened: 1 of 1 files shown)


src/act.drive.cpp
```diff
@@ -1270,8 +1270,8 @@ ACMD(do_driveby)
   }
 
   if ((wielded = GET_EQ(ch, WEAR_WIELD))) {
-    if (!(GET_OBJ_VAL(wielded, 4) == SKILL_PISTOLS || GET_OBJ_VAL(wielded, 4) == SKILL_SMG)) {
-      send_to_char(ch, "You can only do a driveby with a pistol or SMG.\r\n");
+    if (!(GET_OBJ_VAL(wielded, 4) == SKILL_PISTOLS || GET_OBJ_VAL(wielded, 4) == SKILL_SMG || ((0 <= GET_OBJ_VAL(wielded, 3) <= 4)  && ch->in_veh->type = VEH_BIKE)))  {
```
luciensadi, line 1273: This logic hasn't been touched in a dog's age. It probably needs to include IS_GUN for the weapon, and should have the magic number values removed (see utils.hpp for a list of GET_WEAPON_X calls to replace the GET_OBJ_VAL calls with).

Also, this should only allow doing drivebys with one-handed weapons.

UndeadNappist: I want to be able to lance people or cut them in half with a ridiculously oversized sword. Also, this leaves open shooting people with one-handed shotguns.

Bikes are objectively worse than cars in pretty much all circumstances.

UndeadNappist:
> There's messaging in the driveby function that talks about using firearms for it, that will need to be fixed. This also will need to be tested for balance before we actually put it into the game!

Default driveby is insanely powerful disregarding the changes I've made. A nerf would require an all-encompassing one, but most people don't do drivebys due to it being slow.

UndeadNappist:
Still working on the other firearms things later in the logic chain, PR is incomplete.

UndeadNappist:
Undid using two-handed melee weapons on a bike, as there are one-handed monoswords and one-handed lances, so you can still fulfill that dream.

src/act.drive.cpp
```diff
@@ -1268,19 +1268,23 @@ ACMD(do_driveby)
...
+ 
   if ((wielded = GET_EQ(ch, WEAR_WIELD))) {
-    if (!(GET_OBJ_VAL(wielded, 4) == SKILL_PISTOLS || GET_OBJ_VAL(wielded, 4) == SKILL_SMG)) {
-      send_to_char(ch, "You can only do a driveby with a pistol or SMG.\r\n");
+    if (GET_OBJ_TYPE(wielded) != ITEM_WEAPON) {
+      send_to_char(ch, "You have to be wielding a weapon!"
+      return;
+    }
+    if ((IS_OBJ_STAT(wielded, ITEM_EXTRA_TWOHANDS) || (!WEAPON_IS_GUN(wielded) && ch->in_veh->type = !VEH_BIKE)))  {
+      send_to_char(ch, "You can only do a driveby with a one handed weapon in a vehicle and a melee weapon while on a bike!\r\n");
```
luciensadi, line 1278: This message is confusing, try splitting it with an if-check and giving them the proper message for the vehicle they're in.

src/act.drive.cpp
```diff
@@ -1268,19 +1268,23 @@ ACMD(do_driveby)
...
       return;
     }
     if (!IS_NPC(vict) && !PRF_FLAGGED(ch, PRF_PKER) && !PRF_FLAGGED(vict, PRF_PKER)) {
       send_to_char(ch, "You have to be flagged PK to attack another player.\r\n");
       return;
     }
 
-    send_to_char(ch, "You point %s towards %s and open fire!\r\n", GET_OBJ_NAME(wielded), GET_NAME(vict));
-    snprintf(buf, sizeof(buf), "%s aims %s towards %s and opens fire!\r\n", GET_NAME(ch), GET_OBJ_NAME(wielded), GET_NAME(vict));
+    send_to_char(ch, "You swing %s towards %s and attack!\r\n", GET_OBJ_NAME(wielded), GET_NAME(vict));
```
luciensadi, line 1286: Instead of rewriting all driveby attack messages, print a different attack message depending on the weapon type used.

src/act.drive.cpp
```diff
@@ -1292,9 +1296,13 @@ ACMD(do_driveby)
   }
 
   for (pass = ch->in_veh->people; pass; pass = pass->next_in_veh) {
-    if (pass != ch) {
-      if (PLR_FLAGGED(pass, PLR_DRIVEBY) && AWAKE(pass) && GET_EQ(pass, WEAR_WIELD) &&
-          GET_OBJ_VAL(GET_EQ(pass, WEAR_WIELD),4) >= SKILL_FIREARMS) {
+    if (pass != ch && (wielded = GET_EQ(pass, WEAR_WIELD))) {
+      if (GET_OBJ_TYPE(wielded) != ITEM_WEAPON) {
```
luciensadi, line 1300: You did a wielded check earlier in the code, are you expecting that they could un-wield between then and now?

src/act.drive.cpp
```diff
@@ -1317,9 +1325,9 @@ ACMD(do_driveby)
   order_list(list);
   roll_individual_initiative(vict);
   if (GET_INIT_ROLL(vict) < 12)
-    send_to_char(vict, "You suddenly feel a cap busted in your ass as %s zooms past!\r\n", GET_VEH_NAME(ch->in_veh));
+    send_to_char(vict, "You suddenly feel force slamming into your hoop as %s zooms past!\r\n", GET_VEH_NAME(ch->in_veh));
```
luciensadi, line 1328: Split between weapon types.

src/act.drive.cpp
```diff
@@ -1317,9 +1325,9 @@ ACMD(do_driveby)
...
   else
-    send_to_char(vict, "A hail of bullets flies from %s as it zooms past!\r\n", GET_VEH_NAME(ch->in_veh));
+    send_to_char(vict, "A flurry of attacks flies from %s as it zooms past!\r\n", GET_VEH_NAME(ch->in_veh));
```
luciensadi, line 1330: This doesn't read well even if it's purely melee attacks. Something needs to change, probably along the lines of being more specific about what attack is being done and with what melee weapon.

luciensadi (changes requested):
I've filed some initial thoughts on this, but I'm not sure that adding melee driveby attacks is even something that we can support without considerable additional work. Are there book rules we can follow here? This seems like the kind of thing that should induce crash tests and possibly risk damage for the swinger (for more details, hit a mailbox with a baseball bat at 40mph)

PR #636, opened 2022-05-21
Stop anti-sniper hate

TheVilest:
*In the arms of an angel fades into the foreground*

My name is Vile, and i'm here today to stand up for all of the neglected sniper rifles in the Awakened Worlds. For too long have these innocent firearms been subjected to unfair treatment at the hands of their oppressors. For one hundred nuyen a day you can support a sniper that only wants to put a single semi-automatic round through their opponents head at a distance. This outlandish prejudice that supports fully automatic 18-20D assault rifles and machine guns with vastly less dodge'able attacks with APDS (compared to, at most, 14S/14D at max non-dodge modifying attacks) must be stopped in its tracks if we're to move forward into a future where sniper rifles can join hand in hand with all of the other firearms of the world in harmonious tranquility of murder. Gone are the days are improperly calculated vision modifiers of weather and light conditions, and now are the days to embrace our single round "holy fragging hell these are way easier to dodge, lol" breathren in a united front against all of the NPC's we can throw our hard earned APDS rounds at.

Yours truly,
Vile

P.S. - Fight me.

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -1062,9 +1062,7 @@ void affect_total(struct char_data * ch)
     }
     // Ranged weapons grant 0 reach if pistol, 2 reach if anything else with a bayonet, 1 reach for all other categories.
     else {
-      if (GET_WEAPON_SKILL(weapon) != SKILL_PISTOLS)
-      if (GET_WEAPON_SKILL(weapon) != SKILL_TASERS) // Anything smaller than rifle-sized (shotguns/ARs/rifles) has 0 reach. - Vile
-      if (GET_WEAPON_SKILL(weapon) != SKILL_SMG) // That goes for you too, SMG's. - Vile
+      if (GET_WEAPON_SKILL(weapon) != SKILL_PISTOLS && GET_WEAPON_SKILL(weapon) != SKILL_TASERS && GET_WEAPON_SKILL(weapon) != SKILL_SMG) // Cannon Companion page 33 (Or errata if misprint). Page 11 for Improvised Melee table, RIFLE-BUTT has 1 reach. - Vile
         GET_REACH(ch)++;
       struct obj_data *attach = get_obj_proto_for_vnum(GET_WEAPON_ATTACH_UNDER_VNUM(weapon));
       if (attach && GET_ACCESSORY_TYPE(attach) == ACCESS_BAYONET)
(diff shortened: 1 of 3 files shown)


src/handler.cpp
```diff
@@ -1063,6 +1063,8 @@ void affect_total(struct char_data * ch)
     // Ranged weapons grant 0 reach if pistol, 2 reach if anything else with a bayonet, 1 reach for all other categories.
     else {
       if (GET_WEAPON_SKILL(weapon) != SKILL_PISTOLS)
```
luciensadi, line 1065: Chaining the check like this technically works, but it's a lot cleaner to write `if (GWS != pistols && GWS != tasers &&...)`. More readable too.

src/newfight.hpp
```diff
@@ -197,9 +197,11 @@ struct melee_combat_data {
     if (weapon) {
       if (ranged_combat_mode) {
         if (does_weapon_have_bayonet(weapon)) {
+          power += 2;
```
luciensadi, line 200: Why are we adding 2 power? Can you toss in a comment with the source reference?

luciensadi (changes requested):

src/newfight.hpp
TheVilest: Done. Rifle-butts do +2, pistol whipping is only +1, no differentiation between the two in code atm.

src/newfight.hpp
```diff
@@ -197,9 +197,11 @@ struct melee_combat_data {
     if (weapon) {
       if (ranged_combat_mode) {
         if (does_weapon_have_bayonet(weapon)) {
+          power += 2; // Cannon Companion page 33 table, bayonet is listed as +2 Power (or errata if misprinted) -Vile
           dam_type = TYPE_PIERCE;
           skill = SKILL_POLE_ARMS;
         } else {
+          power += 2; // Cannon Companion page 11 table, rifle-butt is listed as +2 Power. -Vile
```
luciensadi, line 204: What about pistol butt etc? This applies to all ranged weapons if it's done here. If they're all supposed to have +2, that's fine.

TheVilest: Pistol is supposed to only be +1 in comparison

luciensadi:
Thanks for this!

PR #637, opened 2022-05-27
Followup to sniper rifles change

UndeadNappist:
Just some probe information/comments cleanup as a followup to another PR.


luciensadi:
Thanks!

PR #638, opened 2022-05-28
Fomori Shopkeeper Buff

UndeadNappist:
You're paying extra build points to be worse than a bog-standard troll, especially considering the Charisma attribute has a smaller number of uses and is much more niche than the Body, Strength, and impact armor you give up.

Fomori are explicitly noted as 'being more attractive' and 'magically active', so now you won't have to feel baited for picking them to negotiate with merchants.

I decided that lowering the cost to be magically active is already covered by charisma and would be way too strong.

### DIFF
--- a/src/newshop.cpp
+++ b/src/newshop.cpp
@@ -1022,6 +1022,7 @@ void shop_buy(char *arg, size_t arg_len, struct char_data *ch, struct char_data 
         case RACE_ORK:
         case RACE_TROLL:
         case RACE_DWARF:
+        case RACE_FOMORI:
           break;
         default:
           target += 4;


luciensadi:
Fomori are still great conjurers, and still retain the troll reach bonus etc, so a melee conjurer build is done best with these. I'm thinking they're not all that disadvantaged.

UndeadNappist:
> Fomori are still great conjurers, and still retain the troll reach bonus etc, so a melee conjurer build is done best with these. I'm thinking they're not all that disadvantaged.

That's a pretty fair point, I was just thinking of opening up the space for people to play non-mages, particularly as (sadly) shaman summons have limited uses in MUD. You could make the same argument for Oni (+1 willpower), but willpower has broader implications and impacts combat pool.

luciensadi:
I think there's plenty of space for people to play non-mages, they just have to deal with the prejudice inherent in the system for being rare metavariants. I think I'll leave this one as canon for the time being. Thanks for the PR!

PR #640, opened 2022-06-02
Lightens restriction on shop carrying

UndeadNappist:
Strength will still have an impact as it flat checks your carry amount, but shopkeepers will now ignore anything else you're carrying.

This means you do still need a certain threshold of strength to purchase (and carry!) certain heavier items, enables newbies to buy up all the mounts/goods that they can want to carry, and makes shopping less annoying in general.

### DIFF
--- a/src/newshop.cpp
+++ b/src/newshop.cpp
@@ -655,7 +655,7 @@ bool shop_receive(struct char_data *ch, struct char_data *keeper, char *arg, int
       send_to_char("You can't carry any more items.\r\n", ch);
       return FALSE;
     }
-    if (IS_CARRYING_W(ch) + GET_OBJ_WEIGHT(obj) > CAN_CARRY_W(ch)) {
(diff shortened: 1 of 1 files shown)


luciensadi:
Have you tested this? I remember the overburdened state being a pain, stopping you from doing things like getting things from containers you're carrying etc, but I don't recall if this was fixed or not.

UndeadNappist:
> Have you tested this? I remember the overburdened state being a pain, stopping you from doing things like getting things from containers you're carrying etc, but I don't recall if this was fixed or not.

It still does all that, and you also won't be able to do things like pull ammo boxes out of your bullet pants, and I also left in the check for seeing if you can actually carry that number of items. Should I change them?

UndeadNappist:
Turns out I was wrong about both; even when overburdened you can still shuffle your inventory around just fine.

luciensadi:
Let's give it a shot then

PR #641, opened 2022-06-03
Enables flare compensation power

UndeadNappist:
Flare compensation power is no longer NERP. Comes with bonus bugfix for adept vision magnification.


luciensadi:
Thanks for this!

PR #642, opened 2022-06-11
Update spec_assign.cpp

UndeadNappist:
Siggy and Toh now have flare comp adept power and it's pulled out of the NERPcropolis.

Refshuffled the metamagics to more appropriate trainers that fit their themes and their rooms' themes.


UndeadNappist:
And an extra fix for anchoring, the NERP metamagic.

src/awake.hpp
```diff
@@ -739,7 +738,8 @@ enum {
 #define META_QUICKENING    6
 #define META_REFLECTING    7
 #define META_SHIELDING    8
-#define META_MAX    9
+#define META_ANCHORING    9
+#define META_MAX    10
```
luciensadi, line 742: Whenever you change something in this file, you also usually need to change constants.cpp. To do that, search that for 'anchoring' and update the array accordingly, leaving some placeholder string in the current spot where anchoring is so the ordering doesn't get messed up.

src/spec_assign.cpp
```diff
@@ -36,12 +36,12 @@ struct train_data trainers[] = {
 /* if they teach < 5 skills, put a 0 in the remaining skill slots    */
 /* available levels are NEWBIE (4), AMATEUR (6/7), and ADVANCED (10/11)   */
 struct teach_data metamagict[] = {
-                       { 24806, { META_REFLECTING, 0, 0, 0, 0, 0, 0, 0 }, "", 0 },
-                         { 4250, { META_CLEANSING, 0, 0, 0, 0, 0, 0, 0 }, "", 0 },
-                         { 4251, { META_MASKING, 0, 0, 0, 0, 0, 0, 0 }, "", 0 },
+                       { 24806, { META_MASKING, 0, 0, 0, 0, 0, 0, 0 }, "", 0 },
```
luciensadi, line 39: What's the rationale behind these changes?

luciensadi (changes requested):

src/spec_assign.cpp
UndeadNappist: Masking is the most common and assigned as a generalist to the more obviously positioned trainers, the two reflecting trainers are identical mobs, cleansing was given to a mob who occupies one of the solitary clean rooms in the middle of grotesque squalor.

Essentially, flavor.

src/awake.hpp
UndeadNappist: Got it! Thanks for the help.

luciensadi (approved):

PR #643, opened 2022-06-26
Treat buff

UndeadNappist:
You can now treat yourself and other people even if they aren't prone. Limitations on time and whether they're in combat still applies.

### DIFF
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -2091,14 +2091,8 @@ ACMD(do_treat)
     return;
   }
 
-  if (vict == ch) {
-    send_to_char("You can't treat yourself!\r\n", ch);
-    return;
-  } else if (CH_IN_COMBAT(vict)) {
+  if (CH_IN_COMBAT(vict)) {
     act("Not while $E's fighting!", FALSE, ch, 0, vict, TO_CHAR);
-    return;
-  } else if (GET_POS(vict) > POS_LYING && !subcmd) {
-    act("$N must at least be lying down to receive first aid.", FALSE, ch, 0, vict, TO_CHAR);
     return;
   } else if (LAST_HEAL(vict) != 0
              || (!IS_NPC(vict)


luciensadi:
I'm not sure about this one, treating someone from 1/10 health involves some pretty serious work that the average person can't do on themselves, especially while standing. We already have accelerated healing rates due to not having hunger etc, are players complaining about it being too slow?

UndeadNappist:
> I'm not sure about this one, treating someone from 1/10 health involves some pretty serious work that the average person can't do on themselves, especially while standing. We already have accelerated healing rates due to not having hunger etc, are players complaining about it being too slow?

Had complaints about it being silly you couldn't treat yourself, especially given how it's objectively worse than magical healing and other effects, which I couldn't disagree with. I could put a health threshold on it. I also have some more medical ideas kicking around, this was just the most simple to address.
Personally, I don't think it'll make much of a difference either way. Healing is pretty fast as you said.

jdevnull:
I think it's less a question of self-treating from 1/10, and more a question of how smaller injuries affect game play. If a character takes light or medium wounds early in one of the more combat extensive autoruns, then a spellcaster patches that up with a heal spell, or an adept makes it disappear in a minute or two with rapid healing. A mundane, however, has the choice of pushing on with the associated penalties, or risking respawns by waiting for that to heal naturally.

In this sort of situation, healing from 1/10 is not exactly meaningful, as the current mud mechanics means that such a character would not be able to heal more than a fixed 3 boxes on success (additional successes have no effect... I also feel that boxes healed should scale with successes, but that's a different topic), so they're still at the seriously wounded stage for the purposes of penalties and pushing on would still be suicidal.

Arguing from a different angle, SR3 does not have a rule preventing self-treating, just an understanding that it is more difficult to self-treat because of wound penalties. Yes, healing mechanics in-mud are intentionally quite different from tabletop, but the additional restriction against self-treating feels arbitrary.

luciensadi:
Well-reasoned, I'll add this in.

PR #644, opened 2022-06-29
First pass on PP menu update

UndeadNappist:
An attempt to clarify some of the feedback, and rework the current ability display.

Current:
```
You know the following abilities:
Astral Perception    (active)
Combat Sense         +3 (3)
etc
```
Proposed:
```
You know the following abilities:
PP     Ability              Level
2.00   Astral Perception    (active)
1.00   Combat Sense         +3 (3.00)
etc
```

### DIFF
--- a/src/act.obj.cpp
+++ b/src/act.obj.cpp
@@ -3511,7 +3511,7 @@ ACMD(do_activate)
 
       int delta = ((int)(GET_REAL_MAG(ch) / 100) * 100) - GET_POWER_POINTS(ch);
       if (total > delta)
-        send_to_char(ch, "That costs %d points to activate, but you only have %d free.\r\n", (total / 100), (delta / 100));
+        send_to_char(ch, "That costs %.2d points to activate, but you only have %.2d free.\r\n", (total / 100), (delta / 100));
       else if (GET_POWER_ACT(ch, i) == x) {
         send_to_char(ch, "%s is already active at rank %d.\r\n", CAP(adept_powers[i]), x);
         return;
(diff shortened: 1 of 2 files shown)


src/act.other.cpp
```diff
@@ -1509,7 +1509,8 @@ ACMD(do_skills)
         continue;
 
       if (GET_POWER_TOTAL(ch, i) > 0) {
-        snprintf(buf2, sizeof(buf2), "%-20s", adept_powers[i]);
+        snprintf(ENDOF(buf), sizeof(buf), "\r\nPP      Ability              Level\r\n")
```
luciensadi, line 1512: Did you test this? Using `ENDOF(buf)` here without initializing the buf means you're gonna have random junk at the beginning of your abilities command. You'll want to use standard `buf` instead.

luciensadi: Also, you put this inside the for-loop, it's gonna stick itself in buf for every power you have active.

luciensadi (changes requested):

src/act.obj.cpp
```diff
@@ -3511,7 +3511,7 @@ ACMD(do_activate)
 
       int delta = ((int)(GET_REAL_MAG(ch) / 100) * 100) - GET_POWER_POINTS(ch);
       if (total > delta)
-        send_to_char(ch, "That costs %d points to activate, but you only have %d free.\r\n", (total / 100), (delta / 100));
```
luciensadi, line 3514: These should be cast as (float) as well, with %.2f instead of %.2d (otherwise it'll perpetually display things like 2.00 for 2.50, etc)

luciensadi (approved):

luciensadi:
Thanks!

PR #646, opened 2022-07-04
First pass on vehicle repo

UndeadNappist:
First pass on vehicle repo. 

Needs: Whatever the arbitrary nuyen value should be set at, locations on where to stick the defines for sanity's sake, adding flags.

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -1209,6 +1209,8 @@ ACMD(do_repair)
   if (veh->damage <= 0) {
     send_to_char(ch, "You go to work on the vehicle and it looks as good as new.\r\n");
     veh->damage = 0;
+    if (IS_NPC(veh->owner) && !(veh->cost) = ambnormal_nuyen_value) { // #define ambnormal_nuyen_value			Some arbitrary number || <= 0 - figure out somewhere to put this. A warning check on loot having too high a sellval.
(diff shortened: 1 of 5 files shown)


UndeadNappist:
Clarity on how the system is outlined:
1. NPC and unowned vehicles are flagged when destroyed and given a timer.
2. This timer only ticks down when not in another vehicle or garage.
3. When the timer ticks to zero the flag is removed regardless of how things are handled.
4. The vehicle is checked. If it's at zero value or less, or above a certain threshold in value, the vehicle is deleted. The upper threshold is going to be low, 5,000 nuyen (or less), so you can only get 500 nuyen worth vehicles here.
5. If it's between those thresholds it's teleported to the junkyard instead.
This is to force players to get the hyper-valuable vehicles immediately, while also having a repo/cleanup mechanic for them.

6. Unowned/NPC flagged vehicles if above 0 value and below some arbitrary value (100k nuyen, or something), when fully repaired, swap their ownership to the repair person. This is to prevent people from getting super vehicles or other problematic things (although this will require building effort to support as it checks value).

Potentially adding a check to the junkyard workshop itself so it may only be used on vehicles below a certain price. This keeps it firmly in the hands of newbies, while also putting a barrier between players who can afford their own private workshops and such.

Anyway, this requires building work, discussion, and staff input, so for now, this is sitting with some framework and in an open draft.

PR #647, opened 2022-07-04
Adds garage door openers

UndeadNappist:
Moves has_key from do_move to utils.cpp and utils.hpp
Vehicles now straight-up ignore doors if there's a valid apartment they are a guest or own and also the driver has a key in their pocket.
Drones with MOD_POKEYSTICK and people on bikes can now push elevator buttons (there are no pokeystick mods yet).

~~None of this is properly tested, but it compiles without breaking. The logic for ignoring doors is also ugly as sin, but it'll probably work.~~

### DIFF
--- a/src/act.movement.cpp
+++ b/src/act.movement.cpp
@@ -679,7 +679,8 @@ void move_vehicle(struct char_data *ch, int dir)
     send_to_char("You aren't the Kool-Aid Man, so you decide against ramming your way out of here.\r\n", ch);
     return;
   }
-  if (!EXIT(veh, dir) || !EXIT(veh, dir)->to_room || EXIT(veh, dir)->to_room == &world[0] || IS_SET(EXIT(veh, dir)->exit_info, EX_CLOSED)) {
+  if (!EXIT(veh, dir) || !EXIT(veh, dir)->to_room || EXIT(veh, dir)->to_room == &world[0] || (IS_SET(EXIT(veh, dir)->exit_info, EX_CLOSED) &&
+  !(ROOM_FLAGGED(EXIT(veh, dir)->to_room, ROOM_HOUSE) && House_can_enter(ch, EXIT(veh, dir)->to_room->number) && has_key(ch, (EXIT(veh, dir)->key))))) {
     send_to_char(CANNOT_GO_THAT_WAY, ch);
     return;
   }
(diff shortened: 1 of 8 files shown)


UndeadNappist:
It works, but it basically silently phases you through the door without a message. I could also switch to having it be a door opener.

src/act.movement.cpp
```diff
@@ -679,7 +679,8 @@ void move_vehicle(struct char_data *ch, int dir)
     send_to_char("You aren't the Kool-Aid Man, so you decide against ramming your way out of here.\r\n", ch);
     return;
   }
-  if (!EXIT(veh, dir) || !EXIT(veh, dir)->to_room || EXIT(veh, dir)->to_room == &world[0] || IS_SET(EXIT(veh, dir)->exit_info, EX_CLOSED)) {
+  if (!EXIT(veh, dir) || !EXIT(veh, dir)->to_room || EXIT(veh, dir)->to_room == &world[0] || (IS_SET(EXIT(veh, dir)->exit_info, EX_CLOSED) &&
+  !(ROOM_FLAGGED(EXIT(veh, dir)->to_room, ROOM_HOUSE) && House_can_enter(ch, EXIT(veh, dir)->to_room->number) && has_key(ch, (EXIT(veh, dir)->key))))) {
```
luciensadi, line 683: You should probably move this added logic to a boolean variable and use that variable here: `bool can_enter_closed_apartment = (your logic)`. Then you can also use that same variable in an if-check to print a message to both the vehicle and the room about opening / closing the door as it passes through.

luciensadi (changes requested):

src/act.movement.cpp
```diff
@@ -679,10 +680,13 @@ void move_vehicle(struct char_data *ch, int dir)
     send_to_char("You aren't the Kool-Aid Man, so you decide against ramming your way out of here.\r\n", ch);
     return;
   }
-  if (!EXIT(veh, dir) || !EXIT(veh, dir)->to_room || EXIT(veh, dir)->to_room == &world[0] || IS_SET(EXIT(veh, dir)->exit_info, EX_CLOSED)) {
+  if (!EXIT(veh, dir) || !EXIT(veh, dir)->to_room || EXIT(veh, dir)->to_room == &world[0] || (IS_SET(EXIT(veh, dir)->exit_info, EX_CLOSED) && (can_enter_closed_apartment == FALSE)) ) {
```
luciensadi, line 683: I think I led you slightly wrong on this. The code here is checking for crash conditions (the exit must exist, and it must point to a room, and that room must not be limbo), which must be checked before your can_enter_closed_apartment bool is populated. There's also not a need to reuse the check, so we don't actually need the bool (we'd only defined it to make the check more readable and portable, but splitting it up like below makes it readable enough).
```c++
if (!EXIT(veh, dir) 
    || !EXIT(veh, dir)->to_room 
    || EXIT(veh, dir)->to_room == &world[0])
{
    send_to_char(CANNOT_GO_THAT_WAY, ch);
    return;
}

if (IS_SET(EXIT(veh, dir)->exit_info, EX_CLOSED)) {
    if (ROOM_FLAGGED(EXIT(veh, dir)->to_room, ROOM_HOUSE)
        && House_can_enter(ch, EXIT(veh, dir)->to_room->number)
        && has_key(ch, (EXIT(veh, dir)->key)))
    {
        send_to_char("The garage door opener on your key beeps, allowing the door to swing open briefly enough to slide through.\r\n", ch);
        // TODO: print a message to the room about the door auto-opening
    } else {
        send_to_char(CANNOT_GO_THAT_WAY, ch);
        return;
    }
}
```

src/transport.cpp
```diff
@@ -1482,7 +1482,7 @@ SPECIAL(call_elevator)
 {
   int i = 0, j, index = -1;
   long rnum;
-  if (!cmd || !ch || !ch->in_room)
+  if (!cmd || !ch || (!ch->in_room && !ch->in_veh->in_room))
```
luciensadi, line 1485: There's no guarantee that the character is in a vehicle here, and even if this worked you'd still be making it so riggers can't use this remotely. Revert this line, then add a RIG_VEH after it to populate a `struct veh_data *veh` variable (search code for examples). Then you can do vehicle-specific checks with a flow structure like:
```c++
if (veh) {
    if (veh->in_room) {
        do something
    } else {
        send_to_veh some error message
```

luciensadi: Actually, thinking about this further, I'm not sure if this roomspec would even trigger at all for someone in a vehicle in the room. This one will be tricky to get to work (maybe trigger the spec function from inside do_push if the only argument is a number or G/B1/B2/etc?)

luciensadi (changes requested):

src/transport.cpp
UndeadNappist: After banging my head on this for several days, I'll toss in the towel. call_elevator() does not work at all for vehicles. I'll leave in the pokestick flag if someone wants to work on this, but I suspect using the mobact logic for elevators is the way to go.

UndeadNappist:
image

luciensadi:
Thanks!

PR #648, opened 2022-07-05
Adds close combat to status

UndeadNappist:
What it says on the tin. Also cuts out reach if close combat is toggled on.


luciensadi:
Looks good, thanks!

PR #649, opened 2022-07-05
Adds zoo cages and windows

UndeadNappist:
Meant for the Seattle zoo.

Should probably be a switch.


luciensadi:
This one should probably be condensed down to a function that's called from each if-block to save on code duplication.

luciensadi:
Thanks!

PR #650, opened 2022-07-08
Vehicle decay

UndeadNappist:
Adds WIP vehicle decay, including underlying logic for certain safety checks.

In dblist.cpp, the logic for decaying vehicle needs to be moved to where the handler is for timers on characters (wherever that is).

VEH_FLAGS logic is still a bit of a mystery to me, will do later.


UndeadNappist:
Reached the upper limits of what I know what I'm doing. I know I have to insert
```
    // Decay smashed NPC/unowned vehicles.
    for (!((veh->in_veh) || (veh->in_room && ROOM_FLAGGED(veh->in_room, ROOM_GARAGE))) && (VEH_FLAGGED(veh, VFLAG_LOOTWRECK)) ) {
       if(GET_VEH_TIMER(veh) < 6) {
         GET_VEH_TIMER(veh)++;
       } else {
          veh->flags.RemoveBit(VFLAG_LOOTWRECK);
          extract_veh(veh);
           }
         }
   ```
   somewhere into a loop for vehicles, but I have no idea where that goes.

Doesn't want to read the new var out of veh_data either.

File: src/constants.cpp [hunk 1]
```diff
@@ -1779,6 +1779,7 @@ const char *veh_flag[] =
     "Workshop",
     "Newbie",
     "Ultrasound"
```
luciensadi, line 1781: You need a comma here

src/fight.cpp
```diff
@@ -5551,10 +5551,16 @@ void chkdmg(struct veh_data * veh)
...
+    extern int max_npc_vehicle_lootwreck_time;
+      
     // Remove any vehicle brains, we don't want them thrown into the street.
     remove_vehicle_brain(veh);
-
+    
+    if (veh->owner <= 0) {
+     veh->flags.SetBit(VFLAG_LOOTWRECK);
+     GET_VEH_DESTRUCTION_TIMER(veh) = max_npc_vehicle_lootwreck_time;
+     }
```
luciensadi, line 5562: nit: indentation

src/memory.cpp
```diff
@@ -47,6 +47,7 @@ struct obj_data *memoryClass::GetObject()
 struct veh_data *memoryClass::GetVehicle()
 {
   struct veh_data *temp = new veh_data;
+  int veh_destruction_timer
```
luciensadi, line 50: missing semicolon, this doesn't compile does it? Also, this isn't the right place to define new data for the vehicle structure, you want structs.hpp

luciensadi (changes requested):

src/memory.cpp
UndeadNappist: Does not compile, no. Thank you for the help.

UndeadNappist:
I have no idea where or how to call the process_vehicle_decay() without crashing everything. Logic is based off how SPECIAL(paint) command works.

src/act.drive.cpp
```diff
@@ -2351,3 +2352,17 @@ int calculate_vehicle_entry_load(struct veh_data *veh) {
 ACMD(stop_rigging_first) {
   send_to_char(ch, "You'll need to stop rigging by using the %s command before you can do that.\r\n", AFF_FLAGGED(ch, AFF_RIG) ? "^WRIG^n" : "^WRETURN^n");
 }
+
+process_vehicle_decay() {
+  struct room_data *room;
+  struct veh_data *veh = NULL;
```
luciensadi, line 2358: This is why it crashes, your veh is null and you're not populating it before you try to check its flags. You need to iterate over the global vehicle list and test the flags for each one.

src/act.drive.cpp
```diff
@@ -2351,3 +2352,17 @@ int calculate_vehicle_entry_load(struct veh_data *veh) {
...
+    if (VEH_FLAGGED(*veh, VFLAG_LOOTWRECK) && !((veh->in_veh) || ROOM_FLAGGED(veh->in_room, ROOM_GARAGE)) ) {
+      if (veh->spare <= time(0)) {
+        snprintf(buf, sizeof(buf), "A hulking utility forklift drives up, lifts the remains of %s up, and drives off to the junkyard.\r\n", veh);
```
luciensadi, line 2361: This makes it sound like the vehicle is retrievable if they go to the junkyard, but it's extracted instead. "Drives off into the sunset with it" maybe.

src/fight.cpp
```diff
@@ -5551,10 +5551,17 @@ void chkdmg(struct veh_data * veh)
...
     int damage_rating, damage_tn;
-
+    extern int max_npc_vehicle_lootwreck_time;
+      
     // Remove any vehicle brains, we don't want them thrown into the street.
     remove_vehicle_brain(veh);
-
+    
+      if (veh->owner <= 0) {
+        veh->locked = TRUE; // Keeps players from messing around with it.
```
luciensadi, line 5560: This means they can't tow it, is that the intent?

luciensadi (changes requested):

src/fight.cpp
UndeadNappist: Can't tow, can't pick it up, and can't repair it. I don't want people messing around with something that will despawn.

luciensadi: This means that all NPC vehicles, once wrecked, can't be towed, repaired, joyridden, etc. Any idea if people are currently doing that stuff?

UndeadNappist: People are driving forklifts and such, and might feel obligated to repair them, but with the decay mechanic they should just respawn, shouldn't they?

People do occasionally do that with the lone star vehicles and a few other things, but the novelty wears off quickly. I'd be more inclined to make a more complicated system (based on what I've written in the issue tracker on the github),  but that requires way more input then I currently have.

luciensadi (approved):
Thanks for this!

PR #652, opened 2022-07-10
Cleaner graffiti messaging

UndeadNappist:
Removes quotation marks and 'is sprayed here' and replaces it with a single line at the top of the graffiti, supposedly.

### DIFF
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -265,8 +265,9 @@ void show_obj_to_char(struct obj_data * object, struct char_data * ch, int mode)
   *buf = '\0';
   if ((mode == SHOW_MODE_ON_GROUND) && object->text.room_desc) {
     strlcpy(buf, CCHAR ? CCHAR : "", sizeof(buf));
-    if (object->graffiti)
+    if (object->graffiti) {
       strlcat(buf, object->graffiti, sizeof(buf));
+    }
     else {
       // Gun magazines get special consideration.
       if (GET_OBJ_TYPE(object) == ITEM_GUN_MAGAZINE && GET_MAGAZINE_BONDED_MAXAMMO(object)) {
@@ -536,9 +537,11 @@ list_obj_to_char(struct obj_data * list, struct char_data * ch, int mode,
   struct obj_data *i;
   int num = 1;
   bool found;
+  bool found_graffiti;
 
   found = FALSE;
-
+  found_graffiti = FALSE;
+  
   for (i = list; i; i = i->next_content)
   {
     if ((i->in_veh && ch->in_veh) && i->vfront != ch->vfront)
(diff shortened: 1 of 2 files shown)


src/act.other.cpp
```diff
@@ -4456,9 +4455,11 @@ ACMD(do_spray)
         return;
       }
       struct obj_data *paint = read_object(OBJ_GRAFFITI, VIRTUAL);
-      snprintf(buf, sizeof(buf), "a piece of graffiti that says \"%s^n\"", argument);
+      if (existing_graffiti_count => 1) {
+        snprintf(buf, sizeof(buf), "Some people have sprayed here:", argument);
```
GarbageJet, line 4459: "Some people have sprayed here:" is weird. How about "Sprayed here," ?

luciensadi:
Every call to snprintf overwrites the buffer. You might be looking for `snprintf(ENDOF(buf), sizeof(buf) - strlen(buf),`, but even that wouldn't work quite right for this case, because if I'm reading this right, the next time someone sprays you'll have:
```
Sprayed here,  %s^n
Sprayed here,  %s^n
  %s^n```
What is your desired output? Can you test and see if you get that?

**Comment by UndeadNappist** [comment]
Desired output is
```
Sprayed here,
   %s^n
   %s^n
```
Currently just prints
```
   %s^n
   %s^n
```

**Comment by UndeadNappist** [comment]
image

Works correctly now.

File: src/act.other.cpp [hunk 2]
```diff
@@ -4456,12 +4456,13 @@ ACMD(do_spray)
         return;
       }
       struct obj_data *paint = read_object(OBJ_GRAFFITI, VIRTUAL);
-      snprintf(buf, sizeof(buf), "a piece of graffiti that says \"%s^n\"", argument);
+      snprintf(buf, sizeof(buf), "^g   %s^n", argument);
```
luciensadi, line 4459: This is the item name itself, it should be left as the piece of graffiti string since it doesn't show up in the room.

src/act.other.cpp
```diff
@@ -4456,12 +4456,13 @@ ACMD(do_spray)
...
       paint->restring = str_dup(buf);
-      snprintf(buf, sizeof(buf), "\"%s^g\" is sprayed here.", argument);
+      snprintf(buf, sizeof(buf), "^g   %s^n", argument);
       paint->graffiti = str_dup(buf);
       obj_to_room(paint, ch->in_room);
 
+      send_to_char("You spray graffiti into the room.\r\n", ch);
```
luciensadi, line 4465: A 'room' is an artificial concept that doesn't apply in a lot of situations, a more generic message like "You tag the area" might fit better here.

luciensadi (changes requested):
Looks a lot better, just a few nits.

src/act.informative.cpp
```diff
@@ -598,6 +601,10 @@ list_obj_to_char(struct obj_data * list, struct char_data * ch, int mode,
         show_obj_to_char(i, ch, mode);
       } else
         if (!corpse && !mode && !IS_OBJ_STAT(i, ITEM_EXTRA_CORPSE)) {
+          if ((GET_OBJ_VNUM(i) == OBJ_GRAFFITI) && (!found_graffiti) ) {
+            found_graffiti = TRUE;
+            send_to_char(ch, "^gSprayed here,^n\r\n");
```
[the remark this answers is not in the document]
UndeadNappist: Nothing. Item goes on top.
image

luciensadi:
Looks good, thanks!

PR #660, opened 2022-07-30
Caribbean League Transport Update

TheVilest:
Halved the times involved with the visa required flight, added a second public-use plane to transport both passengers/vehicles that utilizes the old transports wait times. 

Timing cases might need some tweaking.

Also...

Astral Projection attributes now calculate from modified attributes instead of GET_REAL references, as per canon noted below.

SR3 Core, page 173: "Modifications to your Mental Attributes from spells, cyberware and other sources do affect the abilities of your astral form."

Also...

Lowered the extreme DAMOBJ type multipliers vs object materials, capped them at a 2x multiplier, electronics now have a fighting chance against lightning bolts instead of being vaporized by every spark.

### DIFF
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -2271,14 +2271,14 @@ ACMD(do_astral)
   GET_PHYSICAL(astral) = GET_PHYSICAL(ch);
   GET_MENTAL(astral) = GET_MENTAL(ch);
 
-  GET_REAL_STR(astral) = GET_REAL_CHA(ch);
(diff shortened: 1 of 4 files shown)


TheVilest:
Astral Projection attributes now calculate from modified attributes instead of GET_REAL references, as per canon noted below.

SR3 Core, page 173: "Modifications to your Mental Attributes from spells, cyberware and other sources do affect the abilities of your astral form."

luciensadi:
Thanks for this!

PR #661, opened 2022-08-01
creates a visibility flag system for players

UndeadNappist:
Creates the ability for players to simply flag themselves as visible or non-visible on where, regardless of where they are.

If you aren't flagged visible you don't get socialization bonus.

This is separate from flagging yourself anonymous on where.

### DIFF
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -5457,7 +5457,7 @@ void perform_mortal_where(struct char_data * ch, char *arg)
       struct char_data *i = (d->original ? d->original : d->character);
 
       // Skip them if they aren't in a social-bonus room.
-      if (!i || !i->in_room || !ROOM_FLAGGED(i->in_room, ROOM_ENCOURAGE_CONGREGATION))
+      if (!i || !i->in_room || !AFF_FLAGGED(ch, PRF_VISIBLE))
         continue;
 
       // Skip them if you can't see them for various reasons.
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -5457,7 +5457,7 @@ void perform_mortal_where(struct char_data * ch, char *arg)
       struct char_data *i = (d->original ? d->original : d->character);
 
       // Skip them if they aren't in a social-bonus room.
-      if (!i || !i->in_room || !AFF_FLAGGED(ch, PRF_VISIBLE))
+      if (!i || !i->in_room || !AFF_FLAGGED(ch, PRF_HIDDEN_FROM_WHERE))
         continue;
 
       // Skip them if you can't see them for various reasons.
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -1408,7 +1408,10 @@ ACMD(do_toggle)
         if (ch->desc->pProtocol)
           ch->desc->pProtocol->do_coerce_ansi_capable_colors_to_ansi = FALSE;
       }
-    } else {
+    } else if (is_abbrev(argument, "RP") || is_abbrev(argument, "RPING") || is_abbrev(argument, "roleplay") || is_abbrev(argument, "visible")) {
+      result = PRF_TOG_CHK(ch, PRF_VISIBLE);
+      mode = 48;
+      } else {
       send_to_char("That is not a valid toggle option.\r\n", ch);
       return;
     }
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -1408,8 +1408,8 @@ ACMD(do_toggle)
         if (ch->desc->pProtocol)
           ch->desc->pProtocol->do_coerce_ansi_capable_colors_to_ansi = FALSE;
       }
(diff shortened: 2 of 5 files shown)


stacey-p:
omg i love this so much

src/act.other.cpp
```diff
@@ -1408,7 +1408,10 @@ ACMD(do_toggle)
         if (ch->desc->pProtocol)
           ch->desc->pProtocol->do_coerce_ansi_capable_colors_to_ansi = FALSE;
       }
-    } else {
+    } else if (is_abbrev(argument, "RP") || is_abbrev(argument, "RPING") || is_abbrev(argument, "roleplay") || is_abbrev(argument, "visible")) {
```
luciensadi, line 1411: Using 'rp' to toggle this won't actually work (it's shadowed by the RPE flag). Let's drop most of these and call it 'nosocialization' so it's clearer what its function is. ("nosocialization", "socialization")

You also need to add messages to the end of the big block above, as they're what's printed when this command is run. The game will have undefined behavior (will likely crash) if run in its current state.

src/awake.hpp
```diff
@@ -439,7 +439,8 @@ enum {
 #define PRF_NO_WEATHER                   64
 #define PRF_DISABLE_XTERM                65
 #define PRF_COERCE_ANSI                  66
-#define PRF_MAX                          67
+#define PRF_VISIBLE                      67
```
luciensadi, line 442: **Please test your code before you submit a pull request.** Whenever you add a preference flag, you need to add it to the constants block for preference flags, and I've configured the game to not even boot up before this step is complete. You'll get the following error message if you launch the game in its submitted state:
```
Aug  1 07:03:14 :: Error: You added a prf flag but didn't add it to preference_bits in constants.cpp (or forgot a comma)!
Process 27388 exited with status = 54 (0x00000036)
```

Also, PRF_VISIBLE isn't easy to decipher what it is when reading the code-- it sounds like it's related to staff visibility. Let's rename the definition to PRF_HIDDEN_FROM_WHERE or something to make it self-documenting.

luciensadi (changes requested):

luciensadi:
I should also add that I'm on the fence about this feature in the first place. What's the driver for people wanting this? What experiences are they trying to avoid that aren't solved by the existing block feature?

UndeadNappist:
I have had several people comment about a docwagon service for players to locate them. This would enable their exact location to be visible - regardless of whether they are in a social room or not - as part of an opt in process. You wouldn't get social bonus except in a socialization room but it would be a definite "I am here and available to play."

Similarly, if you don't want to be bothered but happen to spend .01 seconds in a where room and someone catches you grinding, this would ensure you never pop up on the where list to begin with, and be an outright statement of "I do not wish to interact for the moment."

I could definitely see some QoL room here where it automatically de-flags you have spending some time in a social room, or you can choose to have you WHERE visibility be manual or.. logic like that. Something to make lives more convenient instead of simply switching to full manual controls.

src/act.informative.cpp
```diff
@@ -5457,7 +5457,7 @@ void perform_mortal_where(struct char_data * ch, char *arg)
       struct char_data *i = (d->original ? d->original : d->character);
 
       // Skip them if they aren't in a social-bonus room.
-      if (!i || !i->in_room || !ROOM_FLAGGED(i->in_room, ROOM_ENCOURAGE_CONGREGATION))
+      if (!i || !i->in_room || !AFF_FLAGGED(ch, PRF_VISIBLE))
```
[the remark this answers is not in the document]
UndeadNappist: This is intentional.

src/awake.hpp
UndeadNappist: Good to know. At the moment, I wanted to submit a draft, but didn't know if the code was even going to get merged. Just wanted it out of my head.

I'll make those changes anyway, good practice.

luciensadi:
Docwagon for players is better served by hooking into the docwagon alert code and pinging online players who have opted in when someone is downed. It'll require an opt-in toggle for the downed person, and probably some equipment (a docwagon transceiver?) for the people who want to be notified about it, and it'll need to integrate with the block/ignore code to not alert over a blocked person being downed.

Someone being spotted while traveling through a where room and causing someone else to go there is inconvenient, but I don't think it happens all that often. I'd bet the person who traveled there to find them would hang out there a bit anyways, raising the chance for someone else to join them for a scene.

Adding more QoL around this doesn't change the baseline issue-- why are we letting people opt out of the socialization system? What's the driver behind this request? What's making people not want to be visible, and how can we address that root cause?

UndeadNappist:
That's all pretty sensible, things I hadn't considered.

The intention is not opting out of where and more opening the door for opting _in_ for having the Where command useable on you regardless of where you are in an indicator that you are perfectly willing to have other people show up, interact, and/or roleplay with them. Along with having the ability to turn it off if you wish to have a private scene (or just grind and not talk to people).

This was the just the simplest (albeit probably not best) way I could think of implementing it.

Along with leaving the option to disappear off where if you wanted entirely if you just weren't in the mood.

PR #663, opened 2022-08-01
Odds and Ends... Again

TheVilest:
- Added spec assignment for Portland DocWagon ATM.
- Updated Archetype Decker firearm to more suitable loadout.

### DIFF
--- a/src/archetypes.cpp
+++ b/src/archetypes.cpp
@@ -396,7 +396,6 @@ struct archetype_data *generate_decker() {
 
   // Inventory.
   arch->weapon = OBJ_SCK_MODEL_100;
-  arch->weapon_barrel = OBJ_GAS_VENT_IV;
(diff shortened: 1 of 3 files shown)


src/archetypes.cpp
```diff
@@ -396,7 +396,6 @@ struct archetype_data *generate_decker() {
 
   // Inventory.
   arch->weapon = OBJ_SCK_MODEL_100;
-  arch->weapon_barrel = OBJ_GAS_VENT_IV;
```
luciensadi, line 399: Why are we removing the vent IV?

TheVilest: Added the gas vent IV to the new firearm itself (obj vnum 91000), along with a smartlink-2 to match the decker archetype cyberware selection (it was previously a standard smartlink accessory).

PR #664, opened 2022-08-02
Adds color-code based flagging to ##

UndeadNappist:
When you use toggle color codes in helpfiles, resends any instance of # as ^^#, which then displays as ^#.

This is to ensure builders can realize ## is within a block of text that cause sound triggers on screenreaders and not simply have to be aware of them by looking directly into game files, as they are normally invisible.

image


luciensadi:
Useful, thanks!

PR #665, opened 2022-08-08
Rip and Tear

TheVilest:
Added two new cyber weapons, with retractable options:
- Cyber Fangs
- Horn Implants (AKA, Cyber Horns)

Recycled THRASH into GORE for horn attacks, added new messaging to support it.

Changed Handrazors attack messaging from STAB to CLAW.


luciensadi:
Nice, I didn't have a single change to suggest. You're getting good at this!

PR #669, opened 2022-08-20
Quick fixes

jdevnull:
A few quick fixes:
- Previously, do_matrix_score was changed to round-up the calculation of a decker's detection factor to match the core rules, but this wasn't propagated to the actual tests. This adds the proper rounding to the system_test function.

- Fixed damage compensators conditional.

- Increase Reaction spell should have the same bioware compatibilities as the Increase Reflexes +X spells.

Khai/Azazel


jdevnull:
Increase Reaction fix requires a bit more work; reverted for now.

jdevnull:
- Rolls feedback for ranged attacks has a separate entry for combat pool (bonus_if_not_too_tall), so it should be displayed before the dice are added together.

luciensadi:
Looks good, thanks!

PR #670, opened 2022-09-02
Autodismissal of spirits at sunset

jdevnull:
Previously, autodismissal of spirits at sunrise was changed to decrement a counter, with autodismissal when that counter runs out. But sunset still had the code to immediately autodismiss, without regard to the counter. This fixes sunset to do the same thing as sunrise.

-Khai


luciensadi:
Looks good, thanks!

PR #671, opened 2022-09-04
Fix to foot anchor vs knockdown

jdevnull:
In db.cpp line 6460, cyb_footanchor is set to 0 rating, so the existing code ends up having no effect on knockdown tests. This fixes foot anchors to match MM pg 29: when anchored, each foot anchor reduces the TN to resist knockdown by 1.

- Khai


luciensadi:
Looks good, thanks!

PR #672, opened 2022-09-05
Pass the Dutchie 'pon the left hand side (Wagwan?)

TheVilest:
Added a new drug: Deepweed
- Forces Awakened to astrally perceive, even adepts without the Astral Perception power.
- Very addictive magical compound. 7p addiction code.

Added in two pieces of Genetech (SOTA 63, page 25-28), essentially functions as cultured bioware by canon.
- Calcitonin Treatment, +1 Bod, incompatible with Muscle Replacement, Cyberlimbs per canon.
- Erythropoitin Treatment, +2 Athletics dice, incompatible with Muscle Replacements per canon.

Reworked Reflex Recorder bioware to give them applicable functionality
- Gives +1 die to Physical Tests, each reflex recorder has its own individual linked skill .

### DIFF
--- a/src/deck_build.cpp
+++ b/src/deck_build.cpp
@@ -1010,9 +1010,10 @@ ACMD(do_build) {
       success = secondary_successes;
   }
 
+  GET_PART_BUILD_SUCCESSES_ROLLED(obj) = success;
+
   // If we failed, shorten the build time.
   if (success < 1) {
-      GET_PART_BUILD_SUCCESSES_ROLLED(obj) = success;
       success = srdice() + srdice();
   }
 
@@ -1201,12 +1202,15 @@ void clear_all_cyberdeck_part_pointers_pointing_to_deck(struct obj_data *deck) {
   log_vfprintf("DEBUG_DECKPART: Clearing deck part pointers to %s.", GET_OBJ_NAME(deck));
 
   // Iterate over the in-progress deck parts.
-  for (std::vector<struct obj_data *>::iterator it = global_in_progress_deck_parts.begin(); it != global_in_progress_deck_parts.end(); it++) {
+  for (std::vector<struct obj_data *>::iterator it = global_in_progress_deck_parts.begin(); it != global_in_progress_deck_parts.end(); ) {
     // If it's pointing to our deck, clear its pointer and drop it from the list.
-    if ((*it)->cyberdeck_part_pointer == deck) {
+    if ((*it) && (*it)->cyberdeck_part_pointer == deck) {
       log_vfprintf("DEBUG_DECKPART: Clearing %s's deck pointer to %s.", GET_OBJ_NAME(*it), GET_OBJ_NAME(deck));
       (*it)->cyberdeck_part_pointer = NULL;
+      (*it) = NULL;
       it = global_in_progress_deck_parts.erase(it);
+    } else {
+      ++it;
     }
   }
 }


src/drugs.cpp
```diff
@@ -412,6 +416,13 @@ void apply_drug_modifiers_to_ch(struct char_data *ch) {
           GET_TARGET_MOD(ch)++;
           GET_WIL(ch)++;
           break;
+        case DRUG_DEEPWEED:
+          GET_WIL(ch)++;
+          if (GET_TRADITION(ch) == TRAD_HERMETIC || GET_TRADITION(ch) == TRAD_SHAMANIC || GET_TRADITION(ch) == TRAD_ADEPT)
+          PLR_FLAGS(ch).SetBit(PLR_PERCEIVE);
+          else if (GET_TRADITION(ch) != TRAD_MUNDANE)
+          snprintf(buf, sizeof(buf), "Your vision blurs but it fails to pierce the astral veil.\r\n");
```
luciensadi, line 424: Have you tested this? I think it's gonna spam mundies since this function is called relatively often.

src/drugs.cpp
```diff
@@ -427,6 +438,10 @@ void apply_drug_modifiers_to_ch(struct char_data *ch) {
           GET_CHA(ch) = 1;
           GET_WIL(ch) /= 2;
           break;
+        case DRUG_DEEPWEED:
+          GET_TARGET_MOD(ch)++;
+          PLR_FLAGS(ch).RemoveBit(PLR_PERCEIVE);
```
luciensadi, line 443: This will pretty regularly strip perception. I think this function is called once every minute or so, maybe even faster? Worth testing.

src/iedit.cpp
```diff
@@ -277,6 +277,24 @@ void iedit_disp_mod_vehicle_type_menu(struct descriptor_data *d)
...
+                " 7) Pistols\r\n"           " 8) Rifles\r\n"            " 9) Shotguns\r\n"
+                "10) Assault Rifles\r\n"    "11) Submachine Guns\r\n"   "12) Mounted Gunnery\r\n"
+                "13) Tasers\r\n"            "14) Machine Guns\r\n"      "15) Assault Cannons\r\n"
+                "16) Projectiles\r\n"       "17) Throwing Weapons\r\n"  "18) Athletics\r\n"
+                "19) Stealth\r\n"           "20) Electronics\r\n"       "21) Biotech\r\n"
+                "22) Bike B/R\r\n"          "23) Car B/R\r\n"           "24) Drone B/R\r\n"
+                "25) Truck B/R\r\n"         "26) Pistol B/R\r\n"        "27) Shotgun B/R\r\n"
+                "28) Rifle B/R\r\n"         "29) Heavy Weapons B/R\r\n" "30) SMG B/R\r\n"
+                "31) Piloting Bike\r\n"     "32) Piloting Car\r\n"      "33) Piloting Truck\r\n"
+               "Enter skill for reflex recorder:\r\n", CH);
```
luciensadi, line 295: I didn't see any bounds checking for this added in, what happens if someone enters -1 or 50? 

Also, it'd probably be better to have this draw directly from the skills constants-- use a for-loop to go over the skills from 1 to X, printing the number and name for each. Check out medit's skill assignment menu for an example of that.

src/iedit.cpp
```diff
@@ -806,6 +824,10 @@ void iedit_disp_val4_menu(struct descriptor_data * d)
           return;
       }
       break;
+    case ITEM_BIOWARE:
+      if (GET_OBJ_VAL(OBJ, 0) == BIO_REFLEXRECORDER)
+      iedit_disp_record_menu(d);
```
luciensadi, line 829: nit: tab, but it's even better if it has brackets:
```c++
if (x) {
  // thing to do
}
```

src/utils.cpp
```diff
@@ -1516,13 +1516,14 @@ int get_skill(struct char_data *ch, int skill, int &target)
   // Iterate through their bioware, looking for anything important.
   if (should_gain_physical_boosts && ch->bioware) {
     for (struct obj_data *bio = ch->bioware; bio; bio = bio->next_content) {
-      if (GET_BIOWARE_TYPE(bio) == BIO_REFLEXRECORDER && GET_OBJ_VAL(bio, 3) == skill) {
-        strlcat(gskbuf, "Reflex Recorder skill increase: 1. ", sizeof(gskbuf));
-        skill_dice++;
-      } else if (GET_BIOWARE_TYPE(bio) == BIO_ENHANCEDARTIC)
+      if (GET_OBJ_VAL(bio, 0) == BIO_REFLEXRECORDER) {
+        recorder = GET_OBJ_VAL(bio, 5);
```
luciensadi, line 1520: Instead of using a raw value call like this, please create a new reflex recorder block in utils.hpp that has a descriptive name of what this value is. Example from another piece of bioware:

```c++
// Platelet factory extra data.
#define GET_BIOWARE_PLATELETFACTORY_DATA(bioware)       (GET_OBJ_VAL((bioware), 5))
#define GET_BIOWARE_PLATELETFACTORY_DIFFICULTY(bioware) (GET_OBJ_VAL((bioware), 6))
```

src/utils.cpp
```diff
@@ -1516,13 +1516,14 @@ int get_skill(struct char_data *ch, int skill, int &target)
...
+      } else if (GET_OBJ_VAL(bio, 0) == BIO_ENHANCEDARTIC)
         enhan = TRUE;
-      else if (GET_BIOWARE_TYPE(bio) == BIO_SYNTHACARDIUM)
-        synth = GET_BIOWARE_RATING(bio);
+      else if (GET_OBJ_VAL(bio, 0) == BIO_SYNTHACARDIUM)
+        synth = GET_OBJ_VAL(bio, 1);
```
luciensadi, line 1524: `synth = GET_BIOWARE_RATING(bio);`

src/utils.cpp
```diff
@@ -1628,6 +1629,179 @@ int get_skill(struct char_data *ch, int skill, int &target)
     }
   }
 
+  // Reflex Recorder, adds +1 die to Physical Skill tests for its associated skill. -Vile
+  // Utilizes Value 6 to reference its associated skill.
+  if (recorder)
```
luciensadi, line 1634: Not having brackets to mark your code is going to cause a lot of hidden bugs. You should always put {} around your statements to avoid logical flow issues that are hard to diagnose.

src/utils.cpp
```diff
@@ -1628,6 +1629,179 @@ int get_skill(struct char_data *ch, int skill, int &target)
...
+  for (struct obj_data *bio = ch->bioware; bio; bio = bio->next_content)
+  if (skill == SKILL_EDGED_WEAPONS && GET_OBJ_VAL(bio, 5) == 1) {
+    snprintf(ENDOF(gskbuf), sizeof(gskbuf) - strlen(gskbuf), "Reflex Recorder Skill Increase: +1. ");
+    skill_dice += 1;
+  }
```
luciensadi, line 1639: I think something went wrong with your copy-paste in this section. The effect of this combined for/if is: "For every piece of bioware you have, add 1 skill if the unchanging skill value is edged weapons." Thus, if someone has 10 pieces of ware and is using the edged weapons skill, they get +10 to that skill.

src/utils.cpp
```diff
@@ -1628,6 +1629,179 @@ int get_skill(struct char_data *ch, int skill, int &target)
...
+  for (struct obj_data *bio = ch->bioware; bio; bio = bio->next_content)
+  if (skill == SKILL_WHIPS_FLAILS && GET_OBJ_VAL(bio, 5) == 3) {
+    snprintf(ENDOF(gskbuf), sizeof(gskbuf) - strlen(gskbuf), "Reflex Recorder Skill Increase: +1. ");
+    skill_dice += 1;
+  }
+  for (struct obj_data *bio = ch->bioware; bio; bio = bio->next_content)
+  if (skill == SKILL_CLUBS && GET_OBJ_VAL(bio, 5) == 4) {
```
luciensadi, line 1651: Wait, did you define a custom mapping for these skills? SKILL_CLUBS is 6, but you have 4 here. If you did, rip it out and start again, because this is going to be a nightmare to maintain (imagine glancing at this a year later and trying to intuitively understand what's going on).

Happy to help you figure out the logic for that in iedit.

src/utils.cpp
```diff
@@ -1628,6 +1629,179 @@ int get_skill(struct char_data *ch, int skill, int &target)
...
+    skill_dice += 1;
+  }
+  for (struct obj_data *bio = ch->bioware; bio; bio = bio->next_content)
+  if (skill == SKILL_UNARMED_COMBAT && GET_OBJ_VAL(bio, 5) == 5) {
+    snprintf(ENDOF(gskbuf), sizeof(gskbuf) - strlen(gskbuf), "Reflex Recorder Skill Increase: +1. ");
+    skill_dice += 1;
+  }
+  for (struct obj_data *bio = ch->bioware; bio; bio = bio->next_content)
+  if (skill == SKILL_CYBER_IMPLANTS && GET_OBJ_VAL(bio, 5) == 6) {
+    snprintf(ENDOF(gskbuf), sizeof(gskbuf) - strlen(gskbuf), "Reflex Recorder Skill Increase: +1. ");
```
luciensadi, line 1662: Since you're not changing any values here and are just appending text, it's better to use strlcat:
```c++
strlcat(gskbuf, "message", sizeof(gskbuf));
```
This avoids unexpected behavior from percent signs (if any are ever added), and also cuts down processing time by a truly miniscule amount. It's also a tiny bit more readable that way.

src/drugs.cpp
```diff
@@ -412,6 +416,13 @@ void apply_drug_modifiers_to_ch(struct char_data *ch) {
           GET_TARGET_MOD(ch)++;
           GET_WIL(ch)++;
           break;
+        case DRUG_DEEPWEED:
+          GET_WIL(ch)++;
+          if (GET_TRADITION(ch) == TRAD_HERMETIC || GET_TRADITION(ch) == TRAD_SHAMANIC || GET_TRADITION(ch) == TRAD_ADEPT)
```
luciensadi, line 421: Brackets {} please

luciensadi (changes requested):

src/utils.cpp
luciensadi: Actually, you should just restore the code that's being deleted here, you want the bioware type and bioware rating macros intact.

src/utils.cpp
luciensadi: To clarify, the ideal state here is that value 5 would contain the skill that we're intending to have the reflex recorder apply to.

PR #673, opened 2022-09-10
Show room on gridguide arrival

jdevnull:
Quality of life changes:
- When gridguide arrives at its destination, automatically show the room to vehicle occupants.
- When the character switches between the front and back of a vehicle, automatically show the room to the character.

-Khai

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -1996,6 +1996,10 @@ void process_autonav(void)
         return;
       }
       if (veh->in_room == veh->dest) {
+        // QoL - show destination room to vehicle occupants on arrival
+        for (struct char_data *ch = veh->people; ch; ch = ch->next_in_veh)
+          if (!IS_NPC(npc))
+            look_at_room(ch, 0, 0)
         send_to_veh("Having reached its destination, the autonav shuts off.\r\n", veh, 0, TRUE);
         veh->cspeed = SPEED_OFF;
         veh->dest = NULL;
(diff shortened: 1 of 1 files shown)


src/act.drive.cpp
```diff
@@ -1996,6 +1996,12 @@ void process_autonav(void)
         return;
       }
       if (veh->in_room == veh->dest) {
+        // QoL - show destination room to vehicle occupants on arrival
+        for (struct char_data *ch = veh->people; ch; ch = ch->next_in_veh) {
+          if (!IS_NPC(ch)) {
```
luciensadi, line 2001: A check here for being flagged screenreader would be good, reader users have definitely asked for the ability to stifle automatic room description delivery.

src/act.drive.cpp
```diff
@@ -2031,6 +2037,7 @@ ACMD(do_switch)
   ch->vfront = !ch->vfront;
   snprintf(buf, sizeof(buf), "$n climbs into the %s.", ch->vfront ? "front" : "back");
   act(buf, TRUE, ch, 0, 0, TO_ROOM);
+  look_at_room(ch, 0, 0)
```
luciensadi, line 2040: A check here would be good too! IIRC it's something along the lines of `if (!PRF_FLAGGED(ch, PRF_SCREENREADER))`

luciensadi (changes requested):

jdevnull:
Done!

luciensadi:
Perfect, thanks!

PR #674, opened 2022-09-22
NPC mage buffs balance pass

jdevnull:
With the imp invis vs ultrasound fix, NPC spellcasters can now be both incredibly hard to hit in addition to already being incredibly hard to damage. This makes them very difficult to handle by player characters that are restricted to physical weapons.

In RPGs, it's not uncommon for balance to be pursued by making tanky and dodgy builds mutually exclusive. Following that idea, let NPC mages commit to a tanky or dodgy strategy, i.e. use an armor or invis spell, but not both. It would not be unreasonable for a thinking opponent to focus on one or the other. This also increases opponent variety and potentially exploration of different strategies (i.e., tanky opponents are better handled with high power weaponry and/or APDS, while dodgy opponents are better handled with full auto and/or surprise).

-Khai

### DIFF
--- a/src/mobact.cpp
+++ b/src/mobact.cpp
@@ -1214,14 +1214,18 @@ bool mobact_process_self_buff(struct char_data *ch) {
     int max_force = GET_MAG(ch) / 100;
     int min_force = MIN(4, max_force);
 
-    // If not invisible already, apply an invisibility spell based on my magic rating and sorcery skill.
-    if (!imp_invis && !std_invis && !IS_ASTRAL(ch)) {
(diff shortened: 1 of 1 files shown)


luciensadi:
I've removed the invis penalties on mundanes for now, which should help with the physical-weapon restriction in the near term. I'll consider this change, but I'm concerned as to how big of an impact it would make on balancing, so will need to think it over. Thanks for submitting the PR for it though!

luciensadi:
I'm in the process of adding flashbangs, which allow players to strip spells from mages for a cost, so this will hopefully solve the remaining components of this issue. I'm going to close this for now, but will keep it in my back pocket in case the current balance changes aren't enough. Thanks again for the submission!

PR #676, opened 2022-11-12
Allow login->delete of hardcore characters

jdevnull:
There already exists a check->disconnect for dead hardcore characters when selecting the "Enter the game" option from the login menu. On the death of a hardcore character, the player does reach this menu and so has the option to delete the character immediately after death.

Removing the additional check at the login prompt allows the player to decide to delete the character at a later time instead of only being able to do so immediately after death.

Khai


luciensadi:
Looks good, thanks!

PR #677, opened 2022-11-14
has_kit refactoring and related code

jdevnull:
Modified has_kit to search all worn slots instead of just hold. When updating related code, also found and fixed two bugs:

- In do_treat, the variable _i_ was first set as biotech skill, and then used as a counter to search equipment for a kit, before finally being used for the success test.
- In do_repair, using a vehicle workshop that is not a facility would result in a +4 TN penalty (as if the ch had no tools at all).

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -1151,7 +1151,8 @@ ACMD(do_repair)
 {
   struct obj_data *obj, *shop = NULL;
   struct veh_data *veh;
-  int target = 4, skill = 0, success, mod = 0;
+  int target = 4, skill = 0, success = 0;
+  bool kit = FALSE;
 
   skip_spaces(&argument);
   if (IS_ASTRAL(ch)) {
@@ -1182,12 +1183,7 @@ ACMD(do_repair)
   target += modify_target(ch);
 
   if (!access_level(ch, LVL_ADMIN)) {
-    for (obj = ch->carrying; obj && !mod; obj = obj->next_content)
-      if (GET_OBJ_TYPE(obj) == ITEM_WORKSHOP && GET_WORKSHOP_TYPE(obj) == TYPE_VEHICLE && GET_WORKSHOP_GRADE(obj) == TYPE_KIT)
-        mod = TYPE_KIT;
-    for (int i = 0; i < NUM_WEARS && !mod; i++)
-      if ((obj = GET_EQ(ch, i)) && GET_OBJ_TYPE(obj) == ITEM_WORKSHOP && GET_WORKSHOP_TYPE(obj) == TYPE_VEHICLE && GET_WORKSHOP_GRADE(obj) == TYPE_KIT)
-        mod = TYPE_KIT;
+    kit = has_kit(ch, TYPE_VEHICLE);
     shop = find_workshop(ch, TYPE_VEHICLE);
     if (!shop) {
       if (veh->damage >= VEH_DAMAGE_NEEDS_WORKSHOP) {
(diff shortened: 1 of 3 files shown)


src/utils.cpp
```diff
@@ -1947,17 +1947,26 @@ void magic_loss(struct char_data *ch, int magic, bool msg)
...
+#define IS_KIT(obj, type) ( GET_OBJ_TYPE((obj)) == ITEM_WORKSHOP && GET_WORKSHOP_TYPE((obj)) == type && (GET_WORKSHOP_GRADE((obj)) == TYPE_KIT || GET_WORKSHOP_GRADE((obj)) == 0) )
 bool has_kit(struct char_data * ch, int type)
 {
-  for (struct obj_data *o = ch->carrying; o; o = o->next_content)
-    if (IS_KIT(o, type))
-      return TRUE;
+  struct obj_data *obj;
+  bool found = FALSE;
+  
+  for (obj = ch->carrying; obj && !found; obj = obj->next_content) {
```
luciensadi, line 1957: Since `obj` isn't being used outside of this scope, let's go ahead and keep the `struct obj_data *obj` declaration line in here as well. Declaring a variable outside of the scope it's used leads to bugs like the one you saw with `int i` above.

src/utils.cpp
```diff
@@ -1947,17 +1947,26 @@ void magic_loss(struct char_data *ch, int magic, bool msg)
...
+    if (IS_KIT(obj, type)) {
+      found = TRUE;
```
luciensadi, line 1959: Is there a reason behind the change from `return TRUE` to `found = TRUE`? Short-circuiting execution with a return the moment you've established its value means the server spends less time looping over things unnecessarily.

src/act.other.cpp
```diff
@@ -2135,17 +2136,14 @@ ACMD(do_treat)
   }
   if (vict->in_room && ROOM_FLAGGED(vict->in_room, ROOM_STERILE))
     target -= 2;
-  i = get_skill(ch, SKILL_BIOTECH, target);
+  skill = get_skill(ch, SKILL_BIOTECH, target);
```
luciensadi, line 2139: Good find on this one!

src/act.drive.cpp
```diff
@@ -1197,9 +1193,11 @@ ACMD(do_repair)
       target += 2;
     }
 
-    if (shop && GET_WORKSHOP_GRADE(shop) == TYPE_FACILITY) {
-      target -= 2;
-    } else if (mod == TYPE_KIT) {
+    if (shop) {
+      if (GET_WORKSHOP_GRADE(shop) == TYPE_FACILITY) {
+        target -= 2;
+      } // if TYPE_WORKSHOP, no target modifier
```
luciensadi, line 1199: This should make vehicle upgrades easier.

luciensadi (commented):
Overall looks good, nice find on those bugs! Some minor nits in the changes to utils.cpp.

src/utils.cpp
jdevnull: Can't test it right now, but a question: when simultaneously declaring and initializing obj in the for loop, should it be:
> struct obj_data *obj = ch->carrying
or
> struct obj_data *obj = &(ch->carrying)

jdevnull: I can never seem to get pointers and references right >.<

Going with:
> struct obj_data *obj = ch->carrying

Since that's how it's done everywhere else in the code.

luciensadi:
Looks good, thanks!

PR #678, opened 2022-11-16
Fixes for adept flare comp & killing hands

jdevnull:
### Flare Comp
set_vision_bit modifies points->vision, which is declared as Bitfield vision[NUM_VISION_TYPES].

EYE_FLARECOMP is defined as (1 << 3), which has a value greater than NUM_VISION_TYPES, so set_vision_bit(ch, EYE_FLARECOMP, VISION_BIT_IS_ADEPT_POWER) might be doing something unexpected.

### Killing Hands
SR3 pg 170, killing hands ignore immunity to normal weapons.

Khai

### DIFF
--- a/src/newfight.cpp
+++ b/src/newfight.cpp
@@ -891,7 +891,7 @@ bool hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
     // This also requires that the attacker is not using killing hands while unarmed, and is not using a weapon focus that they are bonded to.
     if ((IS_SPIRIT(def->ch) || IS_ANY_ELEMENTAL(def->ch))
         && !(att->weapon && GET_POWER(att->ch, ADEPT_KILLING_HANDS))
-        && !(GET_WEAPON_FOCUS_RATING(att->weapon) >= 0 && WEAPON_FOCUS_USABLE_BY(att->weapon, att->ch)))
+        && !(GET_WEAPON_FOCUS_RATING(att->weapon) > 0 && WEAPON_FOCUS_USABLE_BY(att->weapon, att->ch)))
     {
       // We require that the attack's power is greater than double the spirit's force, otherwise it takes no damage.
       // If the attack's power is greater, subtract double the level from it.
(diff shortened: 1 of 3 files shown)


src/newfight.cpp
```diff
@@ -891,7 +891,7 @@ bool hit_with_multiweapon_toggle(struct char_data *attacker, struct char_data *v
     // Namely: We require that the attack's power is greater than double the spirit's force, otherwise it takes no damage.
     // If the attack's power is greater, subtract double the level from it.
     if ((IS_SPIRIT(def->ch) || IS_ANY_ELEMENTAL(def->ch))
-        && (!att->weapon || GET_WEAPON_FOCUS_RATING(att->weapon) == 0 || !WEAPON_FOCUS_USABLE_BY(att->weapon, att->ch)))
+        && ((!att->weapon && !GET_POWER(att->ch, ADEPT_KILLING_HANDS)) || GET_WEAPON_FOCUS_RATING(att->weapon) == 0 || !WEAPON_FOCUS_USABLE_BY(att->weapon, att->ch)))
```
luciensadi, line 894: What is the intent of this change? I'm not quite understanding it.

jdevnull: Generally, if a character is attempting to use unarmed or cyberweapon combat against spirits, immunity to normal weapons apply. But that's not the case if the character is using the killing hands power. 

Thus, if unarmed/cyberweapons and no killing hands, then proceed with immunity to normal weapons. Same applies if killing hands is active while using a weapon, since killing hands doesn't transfer to weapons.

luciensadi: Got it, in that case the logic tree is wrong here. In the case of someone who has no weapon and has turned on killing hands, it'll then query the focus rating of their non-existent weapon and crash. Might need an inline conditional here, like:

```c++
if ((IS_SPIRIT(def->ch) || IS_ANY_ELEMENTAL(def->ch))
     && (att->weapon ? (GET_WEAPON_FOCUS_RATING(att->weapon) == 0 || !WEAPON_FOCUS_USABLE_BY(att->weapon, att->ch)) : !GET_POWER(att->ch, ADEPT_KILLING_HANDS))
```

jdevnull: Ok, think I got it now. If the target is an elemental or spirit, and the attacker is not using killing hands while unarmed, and the attacker is not using a weapon focus that they are bonded to, then immunity to normal weapons applies.

jdevnull: Yeah, I should read your comment all the way through >.>

luciensadi (approved):
Thanks for these!

PR #679, opened 2022-11-22
Autonav wording in veh probe

jdevnull:
A rating N autonav allows up to Nx10 gridguide entries. "Slots" is a closer synonym to "entries" than to "rating", so "It has N slots in its autonav" causes confusion.


luciensadi (approved):

luciensadi:
Thanks for this!

PR #680, opened 2022-11-27
Rework of housing system

luciensadi:
Rips out old housing logic and replaces it with a flexible format.
- Files are stored as human-readable JSON in a structured tree, enabling editing even while game is running
- Apartments can be any number of rooms in any shape
- Staff can restore apartment storage contents without needing to copyover and risk crashes
- OLC for builder-driven creation of apartments
- No more struggling with random multiplications of cost due to lifestyle settings
- Less overall processing (no need to look up apartment info by scanning landlords every time you need it)
- Decorations are stored separately from the world, meaning they're not lost when something changes on the buildport


luciensadi:
Reviewing this is going to be a nightmare.

jdevnull:
Maybe there's a way to incentivize players to join the build server and do testing?

jdevnull (commented):
Lines 2354-2363 still transfers the item from ch to vict before doing quest checks, so the item still disappears from ch's inventory. I'm thinking that part should probably be limited to !IS_NPC(vict).

Edit: Sorry, this was meant to be for commit [980c457](https://github.com/luciensadi/AwakeMUD/pull/680/commits/980c457815daa03e6d5b538a574eeda59ecef213)

PR #681, opened 2022-11-30
Prevent reaching 0 essence

jdevnull:
Using < here instead of <= means you're not prevented from reaching 0 essence on a cyberware install. SR3 pg 56, an essence of 0 means you're dead.

Khai


luciensadi:
It's a step in the right direction, thanks!

PR #682, opened 2022-12-02
Changes to initiative handling

jdevnull:
1. Prevent actions for zero/negative init even when not flagged PRF_SEE_TIPS.
    - NPCs were attacking with negative initiative because they're not flagged PRF_SEE_TIPS

2. Don't let new combatants replace the head of the combat list.
    - Because of the global combat list, and because the global initiative re-roll happens when the head of the list reaches 0 init, re-ordering the list for new combatants can cause already-engaged characters to experience arbitrary length out-of-initiative states throughout the duration of their fight. This prevents that by keeping the original head of the combat list while re-ordering the rest. The tradeoff is that new combatants may reach the global init re-roll while still having some remaining init.

Khai

### DIFF
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -463,10 +463,22 @@ void set_fighting(struct char_data * ch, struct veh_data * vict)
   if (CH_IN_COMBAT(ch))
     return;
 
-  ch->next_fighting = combat_list;
-  combat_list = ch;
-  roll_individual_initiative(ch);
-  order_list(TRUE);
+  // Check to see if they're already in the combat list.
+  bool already_there = FALSE;
+  for (struct char_data *tmp = combat_list; tmp; tmp = tmp->next_fighting) {
+    if (tmp == ch) {
+      mudlog("SYSERR: Attempted to re-add character to combat list!", ch, LOG_SYSLOG, TRUE);
+      already_there = TRUE;
+    }
+  }
+  if (!already_there) {
+    // Setting init to 0 here means new combat actually starts with the next global re-roll.
+    // This prevents arbitray length out-of-initiative states as new combats are initiated.
+    // Also need to avoid being first in the list to avoid setting off an early global re-roll.
+    GET_INIT_ROLL(ch) = 0;
+    ch->next_fighting = combat_list->next_fighting;
+    combat_list->next_fighting = ch;
+  }
 
   FIGHTING_VEH(ch) = vict;
   GET_POS(ch) = POS_FIGHTING;
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -359,7 +359,7 @@ void check_killer(struct char_data * ch, struct char_data * vict)
 void set_fighting(struct char_data * ch, struct char_data * vict, ...)
 {
   struct follow_type *k;
-  if (ch == vict)
+  if (ch == vict || !vict)
     return;
 
   if (IS_NPC(ch)) {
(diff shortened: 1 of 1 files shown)


jdevnull:
Possible bug - can vehicles be surprised?

src/fight.cpp
```diff
@@ -395,17 +395,30 @@ void set_fighting(struct char_data * ch, struct char_data * vict, ...)
     }
   }
   if (!already_there) {
-    ch->next_fighting = combat_list;
-    combat_list = ch;
+    // Setting init to 0 here means new combat actually starts with the next global re-roll.
+    // This prevents arbitray length out-of-initiative states as new combats are initiated.
+    // Also need to avoid being first in the list to avoid setting off an early global re-roll.
+    GET_INIT_ROLL(ch) = 0;
+    ch->next_fighting = combat_list->next_fighting;
```
luciensadi, line 402: This assumes that the combat list is populated, and will crash if nobody is in combat at the time it runs (`combat_list` = `NULL`).

src/fight.cpp
```diff
@@ -5522,21 +5547,9 @@ void roll_individual_initiative(struct char_data *ch)
...
-    // init state and either the flag was getting stripped or the mob became alert before we ever tried to hit it.
-    // Also the no defense condition happens in the hit_with_multi_weapon_toggle() and we can't remove surprise
-    // flag until we process all that.
-    if (FIGHTING(ch)
-        && IS_NPC(FIGHTING(ch))
-        && !MOB_FLAGGED(FIGHTING(ch), MOB_INANIMATE)
-        && GET_MOBALERT(FIGHTING(ch)) == MALERT_CALM
-        && success_test(GET_REA(ch), 4) > success_test(GET_REA(FIGHTING(ch)), 4)) {
-      GET_INIT_ROLL(FIGHTING(ch)) = 0;
-      act("You surprise $n!", TRUE, FIGHTING(ch), 0, ch, TO_VICT);
```
luciensadi, line 5537: What was the decision process behind moving this code from here to the previous function?

src/fight.cpp
```diff
@@ -395,17 +395,30 @@ void set_fighting(struct char_data * ch, struct char_data * vict, ...)
     }
   }
   if (!already_there) {
-    ch->next_fighting = combat_list;
-    combat_list = ch;
+    // Setting init to 0 here means new combat actually starts with the next global re-roll.
+    // This prevents arbitray length out-of-initiative states as new combats are initiated.
+    // Also need to avoid being first in the list to avoid setting off an early global re-roll.
+    GET_INIT_ROLL(ch) = 0;
```
luciensadi, line 401: The implication of this is that if combat is already going on and you do `kill guard`, you'll sit there getting out-of-initiative warnings until the next combat reset, which is not a good experience. I see you've added a message about it below, but I'm not convinced that this is the best way to do it. Maybe roll their init as normal, then cap it to the highest init already in the combat list?

src/fight.cpp
```diff
@@ -5698,8 +5711,10 @@ void perform_violence(void)
     }
 
     // You get no action if you're out of init.
-    if (GET_INIT_ROLL(ch) <= 0 && !IS_JACKED_IN(ch) && PRF_FLAGGED(ch, PRF_SEE_TIPS)) {
```
luciensadi, line 5701: Hah, this didn't just affect mobs, it was an exploit where someone could turn off tips and get unlimited initiative. Nice find.

luciensadi (changes requested):

src/fight.cpp
jdevnull: Oh, right.

src/fight.cpp
jdevnull: Setting init to 0 instead of calling roll_individual_initiative meant that it was possible for the opponent to change mobalert state while waiting for the fight to start, thus blocking surprise.

src/fight.cpp
jdevnull: Ah, that does sound like the better solution, I'll go ahead and do that (which also means the above re-location of the surprise flagging is no longer needed).

One thing I need to think about is the order of actions if both the player and their target roll higher initiative than the current cap. If I understand right, the attacker will be appended to the front of the list first, and then the defender, and order_list will retain that order. So if both exceed the cap, the defender will always act before the attacker, regardless of their actual rolls.

jdevnull:
Ok, the idea now is that, because the global re-roll happens when the head of the list reaches 0 initiative, we just keep the original head while re-ordering the rest of the list whenever a new combatant is added. This should prevent arbitrarily long strings of out-of-initiative messages, with the trade-off that new combatants may encounter the global re-roll early (i.e., while still having some init remaining).

luciensadi:
I'll test it out on my end-- thanks for submitting!

PR #683, opened 2022-12-07
Fix biotech treat math for non-morted chars

jdevnull:
Biotech treat is supposed to recover some number of boxes (see SR3 pg 129). One box has integer value 100, so values added to GET_PHYSICAL should be multiples of 100.


luciensadi:
Thanks for the patch! However, the max physical values are already x100 (see handler.cpp lines 813-814), so this change actually causes them to immediately heal to full from any health.

jdevnull:
Ah, I see the new house rule gets rid of this part of the mechanic entirely. Closing!

PR #684, opened 2022-12-09
Fix detox spell drug detection

jdevnull:
Previously, detox identified the presence of drugs through the drugs' power attribute. Using a bool instead lets detox apply to drugs with 0 power.

### DIFF
--- a/src/act.offensive.cpp
+++ b/src/act.offensive.cpp
@@ -709,8 +709,8 @@ bool _sort_pairs_by_weight(std::pair<int, int> a, std::pair<int, int> b) {
 
 ACMD(do_flee)
 {
-  if (AFF_FLAGGED(ch, AFF_PRONE)) {
-    send_to_char("It's a struggle to flee while prone!\r\n", ch);
+  if (AFF_FLAGGED(ch, AFF_PRONE) || GET_POS(ch) <= POS_SITTING) {
+    send_to_char("It's a struggle to flee while not on your feet!\r\n", ch);
     return;
   }
 
--- a/src/limits.cpp
+++ b/src/limits.cpp
@@ -104,7 +104,7 @@ void mental_gain(struct char_data * ch)
     gain *= 1.5;
 
   if (GET_TRADITION(ch) == TRAD_ADEPT && GET_POWER(ch, ADEPT_HEALING) > 0)
-    gain *= ((float) GET_POWER(ch, ADEPT_HEALING) / 2 + 1);
+    gain *= (((float) GET_POWER(ch, ADEPT_HEALING) / 2) + 1);
   if (GET_BIOOVER(ch) > 0)
     gain /= GET_BIOOVER(ch);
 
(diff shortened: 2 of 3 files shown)


jdevnull:
Two more small fixes:
1. Adept rapid healing was truncating for odd ratings /2. Fixed by changing order of operations.

2. There was a report of a morted mob being able to flee. Fixed by having do_flee also check entity's position. In the event that it ever happens, this fix will also prevent sitting, resting, lying, sleeping, stunned, or dead entities from fleeing.

src/act.offensive.cpp
```diff
@@ -709,8 +709,8 @@ bool _sort_pairs_by_weight(std::pair<int, int> a, std::pair<int, int> b) {
 
 ACMD(do_flee)
 {
-  if (AFF_FLAGGED(ch, AFF_PRONE)) {
-    send_to_char("It's a struggle to flee while prone!\r\n", ch);
+  if (AFF_FLAGGED(ch, AFF_PRONE) || GET_POS(ch) <= POS_SITTING) {
```
luciensadi, line 712: This is already solved for in interpreter.cpp-- the FLEE command has a minimum position of POS_FIGHTING, so if you're not standing and/or fighting, you'll get "Maybe you should get on your feet first?"

src/limits.cpp
```diff
@@ -104,7 +104,7 @@ void mental_gain(struct char_data * ch)
     gain *= 1.5;
 
   if (GET_TRADITION(ch) == TRAD_ADEPT && GET_POWER(ch, ADEPT_HEALING) > 0)
-    gain *= ((float) 1 + GET_POWER(ch, ADEPT_HEALING) / 2);
+    gain *= ((float) GET_POWER(ch, ADEPT_HEALING) / 2 + 1);
```
luciensadi, line 107: I see why you moved the +1 to the end of the expression, but could you wrap the get power division portion in parens so it's clear to the reader that we really do intend to do (x/2) + 1 instead of x/(2+1)? I know that it works correctly already due to PEMDAS, this is just a readability thing.

src/newmagic.cpp
```diff
@@ -1663,13 +1663,15 @@ void cast_health_spell(struct char_data *ch, int spell, int sub, int force, char
       }
 
       base_target = 0;
+      bool drugged = FALSE;
       for (int i = MIN_DRUG; i < NUM_DRUGS; i++) {
         if (GET_DRUG_STAGE(vict, i) != DRUG_STAGE_UNAFFECTED) {
           base_target = MAX(base_target, drug_types[i].power);
```
luciensadi, line 1669: Since we have some drugs with power 0, we probably need to rethink this portion of it, otherwise the spell is working against no TN. Does the book say anything about how this should be handled?

luciensadi (commented):
Good catch, thanks!

src/act.offensive.cpp
jdevnull: For mobs flagged MOB_WIMPY, where do_flee is called automatically at less than half health, does the interpreter apply?

luciensadi: Ah, if that's what you're fixing, then this works.

jdevnull: Oh, related to this, I know you and Vile use checkmarks to mark off the bug reports that are complete. Should I checkmark the ones where I send in a fix, or should I leave that to you?

luciensadi: Leave it to me please, worst-case I end up forgetting it was fixed and fail to reproduce the issue. I'm looking to avoid situations where someone submits a patch and checks the issue off, but the patch doesn't make it into the codebase for whatever reason.

jdevnull: Yeah, makes sense.

src/newmagic.cpp
jdevnull: I haven't been able to find any mention of a power rating for non-damaging drugs, though I only know of rules involving drugs in SR3 and M&M. I'm thinking this is a question for Vile.

luciensadi:
Looks good, thanks!

PR #686, opened 2022-12-14
Calc Highest Bal/Imp separately

Lanphrost:
Currently a 4/2 and a 2/4 will come to a different total of armor (either 5/4 or 4/5)  depending on how they're being layered (either at the time of equipping, or based on the order in armor slots). This change should 'normalize' the formula, giving consistent results (5/5). If the same formula is used elsewhere and needs changed, I couldn't find it.

Thanks for submitting code! It's highly recommended that you discuss changes / new features with Lucien before merging, so if you haven't done that already, please reach out to him in Discord.

Please include in this PR a description of what your code changes (e.g. "Colors all magazines bright purple for easier locating in inventory lists") or what it's intended to do. The more descriptive you can be, the easier it will be for us to review your code!

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -671,8 +671,10 @@ void affect_total(struct char_data * ch)
             }
 
             // Keep track of which worn item provides the highest total value.
-            if (bal + imp > highestbal + highestimp) {
+            if (bal > highestbal) {
               highestbal = bal;
+            }
+            if (imp > highestimp) {
               highestimp = imp;
             }
 


luciensadi:
Thanks for the submission! I'm not sure this is in line with canon layering though. Take the case of having two pieces of armor (10/2, 0/8): Under canon layering, you'd have `10/2 + 0/4 = 11/6` armor (leaving the first piece untouched and halving both of the armor values on the second piece), but under this patch you would have `10/1 + 0/8 = 11/9` armor (halving the second value of the first piece and the first value of the second).

luciensadi:
Appreciate the submission, but this one would unbalance the game! There are a few not-commonly-used pieces in the game that have a high ballistic and low impact or v/v, and if this change went in, the meta would shift so that those items became best in slot for maximum ballistic / impact coverage. At the maximum values, average player armor levels would increase by a good few points in both metrics, which would require rebalancing mobs etc to match the new power level.

Lanphrost:
Understandable! Thanks for taking a look <3

PR #688, opened 2022-12-23
Use non-exploding dice for drug durations

jdevnull:
Rule of 1 and rule of 6 only apply to success tests and open tests (SR3 pg 38-39).


luciensadi:
Good catch, thanks!

PR #690, opened 2022-12-26
Fix get_drug_heal_multiplier truncation

jdevnull:
This solves the problem of near 0 recovery when only affected by a single drug (stage comedown+).


luciensadi:
Good spot, thanks!

PR #691, opened 2023-01-01
Improve DIE command messaging

jdevnull:
Questions on these come up often enough that I think clarification would be useful.

1. When DocWagon pickup is ready, let's make that clearer in OOC., Also, use an alias COMEGETME instead of DIE so as to reduce the confusion on whether or not typing the command is going to kill them.
2. When a newbie is mortally wounded or stunned, let them know that they don't need to be too concerned about dying.

### DIFF
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -1058,6 +1058,17 @@ void die(struct char_data * ch)
   raw_kill(ch);
 }
 
+ACMD(do_dw_retrieve)
+{
+  // If they're ready to be docwagon'd out, save them.
+  if (PLR_FLAGGED(ch, PLR_DOCWAGON_READY)) {
+    docwagon_retrieve(ch);
+  } else {
+    send_to_char(ch, "You have not received a DocWagon trauma team confirmation! You can either wait for help, or give up by typing ^WDIE^n.\r\n");
+  }
+  return;
+}
+  
 /*
  * Lets the player give up and die if they're at 0 or less
  * physical points.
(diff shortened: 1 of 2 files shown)


jdevnull:
Let COMEGETME call do_dw_retrieve. This answers https://discord.com/channels/564618629467996170/863816502465003530/1059237027620589618

src/fight.cpp
```diff
@@ -1058,6 +1058,17 @@ void die(struct char_data * ch)
   raw_kill(ch);
 }
 
+ACMD(do_dw_retrieve)
+{
+  // If they're ready to be docwagon'd out, save them.
+  if (PLR_FLAGGED(ch, PLR_DOCWAGON_READY)) {
```
luciensadi, line 1064: There should probably be a check in here to see if they're wearing a valid modulator at all. Help won't arrive if they're not wearing one.

src/interpreter.cpp
```diff
@@ -1686,9 +1688,15 @@ void command_interpreter(struct char_data * ch, char *argument, const char *tcna
         break;
       case POS_MORTALLYW:
         send_to_char("You are in a pretty bad shape! You can either wait for help, or give up by typing ^WDIE^n.\r\n", ch);
+        if (PLR_FLAGGED(ch, PLR_NEWBIE)) {
+          send_to_char(ch, "^L[OOC: While your TKE is less than 50, your only penalty for dying is losing your current job. See ^wHELP NEWBIE^L for more details.]\r\n");
```
luciensadi, line 1692: For messages where there are no format specifiers in the output, please use the form of `send_to_char("message", ch);` as this avoids putting the string through printf entirely. This is good practice to be in so we don't inadvertently introduce memory leaks / crash conditions if we write messages that have unguarded percent signs in them.

src/fight.cpp
luciensadi: Also, maybe give them an error if they're not mortally wounded? `FAILURE_CASE(GET_POS(ch) != POS_MORTALLYW, "message");` sort of thing

jdevnull: Ah, right.

src/interpreter.cpp
jdevnull: Got it.

luciensadi (approved):
Looks good, thanks!

PR #692, opened 2023-01-13
Ritualcast updates

jdevnull:
1) Enable decrease attribute and decrease cybered attribute for ritualcast, since they can be used as buffs in order to improve the TN for subsequent increase attribute spells. They are still limited by the decrease + increase stacking penalty house rule so abuse shouldn't be a problem.
This answers: https://discord.com/channels/564618629467996170/788953927269613608/1060260644227665930

2) Let ritualcast spells be cast at specified force instead of being nailed to max force.
This answers: https://discord.com/channels/564618629467996170/788953927269613608/1060266306902962358

### DIFF
--- a/src/newmagic.cpp
+++ b/src/newmagic.cpp
@@ -3746,6 +3746,8 @@ ACMD(do_release)
 bool spell_is_valid_ritual_spell(int spell) {
   switch (spell) {
     case SPELL_COMBATSENSE:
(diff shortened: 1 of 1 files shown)


luciensadi:
Looks good, thanks! The decrease spells being enabled will be experimental for a bit to see how it plays out balance-wise, but I'm fine with putting it in for now.

PR #693, opened 2023-02-17
Skullduggery - When My Chopper Goes

TheVilest:
3rd Ward Stepper

- Adds Tactical Computer mods to cyberskulls at reduced essence cost via cyberlimb ECU rules (Man & Machine page 36)

### DIFF
--- a/src/invis_resistance_tests.cpp
+++ b/src/invis_resistance_tests.cpp
@@ -182,6 +182,7 @@ bool can_see_through_invis(struct char_data *ch, struct char_data *vict) {
       if (GET_CYBERWARE_TYPE(cyber) == CYB_TACTICALCOMPUTER) {
         dice += GET_CYBERWARE_RATING(cyber);
         buf_mod(resistance_test_rbuf, sizeof(resistance_test_rbuf), "TacComp", GET_CYBERWARE_RATING(cyber));
+        break;
       }
       else if (GET_CYBERWARE_TYPE(cyber) == CYB_SKULL) {
         if (IS_SET(GET_CYBERWARE_FLAGS(cyber), SKULL_MOD_TAC_COMP)) {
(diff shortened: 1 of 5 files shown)


src/invis_resistance_tests.cpp
```diff
@@ -182,7 +182,13 @@ bool can_see_through_invis(struct char_data *ch, struct char_data *vict) {
       if (GET_CYBERWARE_TYPE(cyber) == CYB_TACTICALCOMPUTER) {
         dice += GET_CYBERWARE_RATING(cyber);
         buf_mod(resistance_test_rbuf, sizeof(resistance_test_rbuf), "TacComp", GET_CYBERWARE_RATING(cyber));
+      }
+      else if (GET_CYBERWARE_TYPE(cyber) == CYB_SKULL) {
+        if (IS_SET(GET_CYBERWARE_FLAGS(cyber), SKULL_MOD_TAC_COMP)) {
+        dice += GET_CYBERWARE_RATING(cyber);
+        buf_mod(resistance_test_rbuf, sizeof(resistance_test_rbuf), "TacComp", GET_CYBERWARE_RATING(cyber));
         break;
```
luciensadi, line 190: The break statement is meant to stop the game from searching the rest of their cyberware when it finds the skull comp, so it should be at the end of both the `if` and `else if` statements here. The as-written behavior will stop it from iterating when it finds a tac-comp skull but will keep going after finding a non-skull comp.

luciensadi (changes requested):
One flow fix, and also a thought: It's probably worth adding logic to the no-overlapping-cyberware function that triggers at shopkeepers to prevent the installation of a tac-skull alongside a standalone tac-comp.

src/invis_resistance_tests.cpp
TheVilest: Good catch

luciensadi (approved):

PR #695, opened 2023-02-19
Fix banishing magic loss and actions

jdevnull:
(1) Temp magic loss is recovered 1 per mud hour.
(2) Cap temp magic loss.
(3) Spirits don't otherwise act when in banishment clashes.

Discussed at https://discord.com/channels/564618629467996170/1076958471288787105

- Khai

### DIFF
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -5966,7 +5966,7 @@ void perform_violence(void)
         stop_fighting(mage);
         end_spirit_existance(spirit, TRUE);
         AFF_FLAGS(mage).RemoveBit(AFF_BANISH);
-      } else if (GET_REAL_MAG(mage) - GET_TEMP_MAGIC_LOSS(mage) < 1) {
+      } else if ((GET_REAL_MAG(mage) / 100) - GET_TEMP_MAGIC_LOSS(mage) < 1) {
         send_to_char(mage, "Your magic spent from your battle with %s, you fall unconscious.\r\n", GET_NAME(spirit));
         act("$n falls unconscious, drained by magical combat.", FALSE, mage, 0, 0, TO_ROOM);
         if (damage(spirit, mage, TYPE_DRAIN, DEADLY, MENTAL)) {
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -5946,7 +5946,7 @@ void perform_violence(void)
       if (success < 0) {
         GET_EXTRA(mage) = 0;
         GET_EXTRA(spirit) = 1;
-        GET_TEMP_MAGIC_LOSS(mage) += MAX(-success, GET_MAG(mage));
+        GET_TEMP_MAGIC_LOSS(mage) += MIN(-success, GET_MAG(mage) / 100);
         send_to_char(mage, "You lock minds with %s, but are beaten back by its force!\r\n", GET_NAME(spirit));
       } else if (success == 0) {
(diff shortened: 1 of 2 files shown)


src/fight.cpp
```diff
@@ -5946,7 +5946,7 @@ void perform_violence(void)
       if (success < 0) {
         GET_EXTRA(mage) = 0;
         GET_EXTRA(spirit) = 1;
-        GET_TEMP_MAGIC_LOSS(mage) -= success;
+        GET_TEMP_MAGIC_LOSS(mage) += MAX(-success, GET_MAG(mage));
```
luciensadi, line 5949: Important to note here is that GET_MAG is 100x the value you'd expect it to be since it's an integer that masquerades as a floating-point number. GET_MAG 600 is equivalent to magic 6.00.

I think the logic here isn't quite right regardless though-- in the case of `success = -1, GET_MAG(mage)/100 = 6`, this would add 6 to their temp magic loss var every time without actually capping the result.

Maybe this: `GET_TEMP_MAGIC_LOSS(mage) = MAX(GET_TEMP_MAGIC_LOSS(mage) - success, GET_MAG(mage) / 100);`

luciensadi (changes requested):

src/fight.cpp
jdevnull: Oops, when I switched the signs, I forgot to switch MAX to MIN, which I think is what we want.

Rubber ducking - if successes is negative, then magic loss should increase by a positive number, where that positive number should not exceed that character's current magic rating.

luciensadi: Note that you're doing a `+=` on the result of your max/min, so whatever value you choose will be added to the current temp loss instead of overwriting it. That's why the version I suggested is a flat assignment `=`.

jdevnull: Yeah, the problem is affect_total does:
`  GET_MAG(ch) -= MIN(GET_MAG(ch), GET_TEMP_MAGIC_LOSS(ch) * 100);
`
So doing a flat assignment comparing to GET_MAG won't work. Comparing against GET_REAL_MAG will work, though I think that's functionally identical to the current solution when taking affect_total into account.

jdevnull: Oh wait, that line in affect_total also is wrong.

jdevnull: Nope, nevermind, missed the part where affect_totals set `ch->aff_abils = ch->real_abils;`

jdevnull: My current `GET_TEMP_MAGIC_LOSS(mage) += MIN(-success, GET_MAG(mage) / 100);` will let GET_MAG hit zero, but when a power focus is in play, that means GET_TEMP_MAGIC_LOSS can become greater than real magic.

Alternate solution `GET_TEMP_MAGIC_LOSS(mage) = MIN(GET_TEMP_MAGIC_LOSS(mage) - success, GET_REAL_MAG(mage) / 100);` will more correctly cap temp magic loss to real magic, but that would also mean using a power focus prevents a character from hitting zero current mag against a tougher spirit and so prevents the spirit from knocking out the mage and ending the banishing attempt.

jdevnull: Alright, new solution follows https://discord.com/channels/564618629467996170/564621949842751508/1077145056357666826, capping temp magic loss = real magic and knocking out the character at that point, regardless of power focus.

luciensadi (approved):
Looks good, thanks!

PR #698, opened 2023-04-08
Fix get_bioover math, rebalance bioover recovery penalty

jdevnull:
Per M&M pg 78, all excessive bioware penalties are based on "points (or part thereof) their Bio Index exceeds their Essence Index". So we want to round up, whereas integer division truncates towards zero. So this fixes all the get_bioover related math.

Related to this, the current recovery penalty for excessive bioware doesn't kick in until get_bioover = 2, at which point it causes a massive drop, cutting recovery in half. So with the above fix, this would cause the recovery penalty to be cut in half when at 1.01 over instead of at 2 over. This request changes the penalty calculation to be -10% per point of get_bioover, which makes for a smoother drop and more closely resembles the rules written in M&M pg 78.

### DIFF
--- a/src/utils.hpp
+++ b/src/utils.hpp
@@ -460,7 +460,7 @@ extern bool PLR_TOG_CHK(char_data *ch, dword offset);
 #define GET_ESSHOLE(ch)       ((ch)->real_abils.esshole)
 #define GET_INDEX(ch)         ((ch)->real_abils.bod_index)
 #define GET_HIGHEST_INDEX(ch)      ((ch)->real_abils.highestindex)
-#define GET_BIOOVER(ch)       (-((int)((GET_ESS((ch)) + 300) - GET_INDEX((ch))) / 100))
+#define GET_BIOOVER(ch)       ((int)(GET_INDEX(ch) - (GET_ESS(ch) + 300) + 99) / 100)
 #define GET_TEMP_MAGIC_LOSS(ch)	((ch)->points.magic_loss)
 #define GET_TEMP_ESSLOSS(ch)	((ch)->points.ess_loss)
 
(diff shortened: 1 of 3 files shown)


src/utils.hpp
```diff
@@ -460,7 +460,7 @@ extern bool PLR_TOG_CHK(char_data *ch, dword offset);
 #define GET_ESSHOLE(ch)       ((ch)->real_abils.esshole)
 #define GET_INDEX(ch)         ((ch)->real_abils.bod_index)
 #define GET_HIGHEST_INDEX(ch)      ((ch)->real_abils.highestindex)
-#define GET_BIOOVER(ch)       (-((int)((GET_ESS((ch)) + 300) - GET_INDEX((ch))) / 100))
```
luciensadi, line 463: What's the goal of adding 99 here? Ensuring that we round up instead of down? If so, does this not compound with the other changes you made that also added +1 to the bio-over value?

jdevnull: Right, the +99 here is to round up. The combination with other places is needed to match the rules as written.

Drug(toxin) resistance rule: "lose 1 die on Resistance Tests for every 2 points (or part thereof) their Bio Index exceeds their Essence Index" (M&M pg78). Checking the rounding limits:

1. Exceeding by 0.01 makes this part (1+99)/100 = 1, then the drug resist part is (1+1)/2 = 1.
2. Exceeding by 2.00 makes this part (200+99)/100 = 2, then the drug resist part is (2+1)/2 = 1.

Because of integer truncation, both parts need to be rounded up to make the 0.01 case cause a penalty, and rounding up both parts won't cause the 2.00 case to give too high of a penalty.

luciensadi:
Merged, thank you!

PR #699, opened 2023-04-09
Make drug guided withdrawal TN easier instead of harder

jdevnull:
Given the treatment cost and increased time to withdrawal, I believe the TN(s) should be easier instead of harder than a forced withdrawal.


luciensadi:
I see the bug here, but I'm not sure if it's supposed to be that you take an additional penalty _without_ the chems, or get a penalty deduction with them-- if drug TN is 5 and you have chems, is it supposed to be 4, or is it supposed to be 6 without chems? Will have to check the book and understand the original intent here before we commit the change.

luciensadi:
Modified behavior to reflect above!

PR #701, opened 2023-06-04
Handle persona programs in store-bought decks like components

jdevnull:
In store-bought decks, persona programs were being displayed like regular programs. This has led to confusion, with several newbie deckers believing that persona programs aren't being loaded because they can't be set to default, but can be manually loaded once connected to the matrix. Persona programs in store-bought decks are actually used to set matrix attributes at the time of connection to the matrix, and so manually loading persona programs do nothing but use up active memory space.

This pull request makes the following changes:
1. persona programs in store-bought decks do not take up memory space
2. do not permit persona programs to be manually loaded when in the matrix
3. do not display persona programs with other software (instead, list them separately like components)

-Khai

### DIFF
--- a/src/act.obj.cpp
+++ b/src/act.obj.cpp
@@ -981,9 +981,11 @@ void perform_get_from_container(struct char_data * ch, struct obj_data * obj,
           return;
         }
 
-        if (GET_OBJ_TYPE(obj) == ITEM_PROGRAM ||
-            (GET_OBJ_TYPE(obj) == ITEM_DECK_ACCESSORY && GET_DECK_ACCESSORY_TYPE(obj) == TYPE_FILE))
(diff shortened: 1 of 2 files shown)


src/act.obj.cpp
```diff
@@ -268,7 +268,12 @@ void perform_put_cyberdeck(struct char_data * ch, struct obj_data * obj,
 
     int space_required = 0;
     if (GET_OBJ_TYPE(obj) == ITEM_PROGRAM) {
-      space_required = GET_PROGRAM_SIZE(obj);
+      // Persona programs don't take up storage memory in store-bought decks
```
luciensadi, line 271: Just to confirm, does this code change cover storage used when uninstalling as well?

jdevnull: Oops, no it doesn't. And fixed below.

luciensadi:
Looks good to me, thank!

PR #703, opened 2023-06-25
Fix store-bought persona programs being put into the wrong output buffer

jdevnull:
Persona programs were being added to buf (with normal software) instead of buf2 (components).


luciensadi:
Thanks!

PR #706, opened 2023-08-08
Fix geas handling

jdevnull:
Bugfix: Using GET_FOCUS_BOND_TIME_REMAINING to track geas could cause the geas to be lost if the character accidentally targets that focus with a do_bond action. This fix moves geas tracking to a different field accessed via GET_FOCUS_GEAS.

Bugfix: In addition, geas couldn't be attached to weapon foci. Because all uses of GET_WEAPON_POSSIBLE_FIREMODES are prefaced with IS_GUN checks, we're able to reuse that field for weapon foci geas. Conveniently, we're able to use the same entry position for both worn foci and weapon foci geas (i.e., GET_FOCUS_GEAS can be used in place of GET_WEAPON_FOCUS_GEAS).

While we're at it, we should show that wielded weapon foci that are WEAPON_FOCUS_USABLE_BY the wielder are in fact activated.

- Khai

### DIFF
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -313,7 +313,9 @@ void show_obj_to_char(struct obj_data * object, struct char_data * ch, int mode)
       strlcat(buf, " (Plan)", sizeof(buf));
     if (GET_OBJ_VNUM(object) == 108 && !GET_OBJ_TIMER(object))
       strlcat(buf, " (Uncooked)", sizeof(buf));
-    if (GET_OBJ_TYPE(object) == ITEM_FOCUS && GET_FOCUS_BOND_TIME_REMAINING(object) == GET_IDNUM(ch))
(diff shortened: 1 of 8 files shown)


luciensadi:
Good spot! This will need a transition snippet added to newdb.cpp to migrate existing geasa from the old slot to the new slot. Probably something in the automatic repair section where if the time remaining matches the idnum, transfer it over?

jdevnull:
So, since I couldn't figure out a clean way to go from the object to the carrying character, I've modified `auto_repair_obj` to also take an `idnum_t owner` parameter.

There's a potential edge case where, if a character is in the process of bonding a foci is saved when the GET_FOCUS_BOND_TIME_REMAINING randomly ends up equal to their character ID, then we might accidentally turn it into a geas.

luciensadi (approved):

luciensadi:
Looks good, thanks!

PR #707, opened 2023-08-13
Make handling of perma/temp attribute modifiers consistent, also fix troll dermal bonus

jdevnull:
Previously, some permanent bonuses (i.e., those that are untimed) were not subject to caps, and some temporary bonuses (i.e., those with time limits and typically have some type of crash effect when the boost wears off) were subject to caps. This makes it so that all permanent modifiers are calculated before caps and all temporary modifiers are calculated after caps.

This fixes nerve strike so that it can now disable somebody that has adept improved quickness.

This does negatively impact troll/giant/cyclops adepts that could previously reach 21/22/23 permanent strength (adept improved strength 6, muscle augmentation 4, and a suprathyroid gland), but would now be capped to 20. Question: do we want to increase the cap to support these character types?

While we're at it, this also fixes the troll/minotaur dermal skin bonus, which should provide +1 body, not impact (SR3, pg 56).

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -766,23 +766,23 @@ void affect_total(struct char_data * ch)
     GET_REA(ch) += has_wired * 2;
   }
   /* effects of bioware */
+  struct obj_data *adrenal = NULL;
   for (cyber = ch->bioware; cyber; cyber = cyber->next_content)
   {
-    if (GET_OBJ_VAL(cyber, 0) != BIO_ADRENALPUMP || (GET_OBJ_VAL(cyber, 0) == BIO_ADRENALPUMP &&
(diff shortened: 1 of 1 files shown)


jdevnull:
Oops, can't assign null to a pointer to a struct. What's the best way to fix this for the adrenal pump?

jdevnull:
Oh, wait, that does work. Nevermind...

luciensadi:
Thanks! This one will take a long time to review just because of how important-yet-spaghettified that particular section of the code is and how difficult it is to test / validate properly.

jdevnull:
I'm now thinking that soft-capping total attributes at the end of everything is better than having to differentiate between "permanent" and "temporary" bonuses. Closing this.

PR #708, opened 2023-08-26
Implement attribute soft capping + fix troll/mino dermal skin

jdevnull:
So, instead of capping attributes with some bonuses but not others, which would require making judgement calls on what should or should not be capped, let's use a soft cap of the totals, so that any points over the cap have half effectiveness. This allows characters to continue to get benefits if they wish to make the effort but will discourage excessive attribute buffing.

Additional notes:
Halfed points are rounded up so that a completely maxed out mundane human/elf still has a chance to deal damage to Oeez.
Nerve strike penalty occurs last, so that modifiers can't make a char immune to nerve strike disabling.
Reaction is derived after int/qui caps and nerve strike.
Grouped skill_dice calculation with where that value is actually used.

And while we're at it:
Troll/mino dermal skin should not increase impact rating, rather, it should provide a bonus die to body (SR3 pg 56).


jdevnull:
Related question: is there a reason for attribute minimums to be capped at 1 instead of 0?

Per decrease (cybered) attribute spell description (SR3 pg 193): "If a Physical Attribute is reduced to 0, the victim is paralyzed. If a Mental Attribute is reduced to 0, the victim stands about mindlessly." Allowing attributes to hit 0 will re-enable an interesting Health category alternative for disabling targets for spellcasting characters and mobs. Balance shouldn't be too difficult since it takes 2 successes per attribute point, it targets TN = 10-Essence, and is resisted by the attribute to be affected.

luciensadi:
> Related question: is there a reason for attribute minimums to be capped at 1 instead of 0?

I'm pretty sure it's because the code has no real way to model a stat at 0. It only really matters for qui, the rest of it is just ignored as far as I'm aware, and qui being zero is not well supported since it's historically only happened with nerve strikes. We'd need a lot of additional supporting code to allow it to go to 0, and even then the effects for players would absolutely suck (imagine an NPC mage dropping a reduce-to-0 on you and wandering off while you're permanently stunlocked until someone kills it)

src/handler.cpp
```diff
@@ -852,8 +825,33 @@ void affect_total(struct char_data * ch)
...
+  GET_MAX_MENTAL(ch) -= GET_MENTAL_LOSS(ch) * 100;
+  GET_MAX_PHYSICAL(ch) -= GET_PHYSICAL_LOSS(ch) * 100;
+
+  // Here so that other modifiers can't make the char immune to disabling via nerve strike
+  if (GET_TEMP_QUI_LOSS(ch))
+    GET_QUI(ch) = MAX(0, GET_QUI(ch) - (GET_TEMP_QUI_LOSS(ch) / TEMP_QUI_LOSS_DIVISOR));
 
+  // Reaction is derived from current atts, so we calculate it after all att modifiers
+  // We don't cap reaction because qui and int are already capped
+  GET_REA(ch) += (GET_INT(ch) + GET_QUI(ch)) >> 1;
```
luciensadi, line 852: Did GET_REA() get zeroed before this? If not, this will cause it to explode.

luciensadi (commented):
The code change looks good from a logical standpoint, but I'm still very much on the fence about the balance aspect here. Do we have an idea of how the average endgame build will change with this?

src/handler.cpp
jdevnull: Yeah, that happens on line 581, which surprisingly is still in that same function despite being 300 lines earlier >.<

jdevnull:
This proposed change is really because I don't like the arbitrariness of the current capping code. Spells are affected but not adept powers. Adrenal pumps are affected but drugs are not. Beyond that, things I'm happy to change are number to use as the cap, whether it should be a hard or soft cap, or how the diminishing returns should scale.

I don't think the numbers I've proposed have any really significant effect on endgame builds, since its typically not worth the effort (though it'd be good to check with Jank and Alak, since there very well might be extreme characters that might be particularly affected). As proposed, effective points won't be lost until a character reaches 22 in an attribute.

The highest attribute that *actually* matters in autorun content is enough strength for a mundane to have a chance to damage Oeez, which would require between 21 and 25, depending on choice of weapon (strength higher than that doesn't matter unless you can get to 35-ish power because Oeez is also a high force spellcaster with an armor spell, so they'll be rolling soak against TN 2 until then). This doesn't matter for adepts and spellcasters because spells, killing hands, and weapon foci ignore immunity to normal weapons. Technically, there's also cybered implant weapons that would let a mundane get to 35 power, but I'd argue that a reach weapon is the better choice, in order to offset the TN penalty from improved invis.

Being able to reach a strength of 30-ish *might* matter for characters that intend to go up against dragons in melee, but that's incredibly risky (as it should be, since it's a DRAGON). This involves rolling against at least TN 6 (because dragons have both reach and improved invis), and any time the dragon wins a clash, their claws will tear right through whatever impact armor rating you've managed to reach. It is much less suicidal to just use a full auto firearm with APDS. Admittedly, adept distance strike does present an exception here,

For context, the maximum possible strength currently is reached by: cyclops 12 + muscle aug 4 + suprathyroid 1 + adept improved str 6 + adept boosted str 6 + aderenal pump 4 + kamikaze 2 + nitro 2 + shaman increase cybered attribute spell 12 = 49. The force 12 increase cybered attribute spell is only really possible right now because of ritualcast, so if we drop that, we're looking at 37 on an adept. On a mundane, we're looking at 27.

For quickness, the most important threshold is getting to 11-12 for the most efficient armor setups, up to 16 for the maximum possible ballistic rating. Past that point, you only get +1 combat pool die per +3 qui.

For body, I'd say it's generally a good idea to stick with permanent sources because taking additional damage from drain/crashes is counterproductive to the reason characters want more body for. So, troll 11 + dermal sheathing III 4 + titanium bone lacing 2 + adept improved bod 6 = 23 (or 35 if we include a force 12 shaman increase cybered attribute spell).

It's not currently possible to get any mental attributes past 21 (gnome willpower 8 + 1 pain editor + 12 shaman increase cybered attribute).

jdevnull:
Correction on maximum possible strength currently on live (assuming I understand the code right): cyclops 12 + muscle aug 4 + suprathyroid 1 + adrenal pump 4 + shaman increase cybered attribute spell 12 = 33, but this is then capped to 20.

So, the above 20 + adept improved str 6 + adept boosted str 6 + kamikaze 2 + nitro 2 = 36.

With the proposed soft cap, the 49 turns into 20 + (49-20)/2 (round up) = 35.

src/handler.cpp
jdevnull: It looks like where GET_REA is zeroes is also where all the attributes are reset to their real values and the pools are emptied. After all that, we start adding everything back in, including bonuses to reaction directly from things like wired reflexes and spells. Unless we want to tackle breaking this function down into more digestible subfunctions, where GET_REA is currently zeroed does make sense.

luciensadi:
Merging in for balance testing, thanks!

PR #709, opened 2023-09-09
Various fixes

jdevnull:
The save_char function calls unequip_char, obj_from_cyberware, obj_from_bioware, obj_to_cyberware, obj_to_bioware, and equip_char for every item worn or installed into the character. Every one of these function calls result in an affect_total call. In addition to the performance hit for the many redundant calls, this causes hacking pool and spell pool allocations to be reduced/reset every time save_char is called (e.g., through point_update every 2 minutes).

This solution gives unequip_char, obj_from_cyberware, obj_from_bioware, obj_to_cyberware, obj_to_bioware, and equip_char a new recalc parameter (which defaults to TRUE) that controls whether affect_total is called.

Unless there are other extra calls to affect_total that I missed, this solves the long-standing hacking pool resets as well as the more recent spell pool resets.


src/handler.cpp
```diff
@@ -1667,12 +1676,14 @@ bool equip_char(struct char_data * ch, struct obj_data * obj, int pos)
                   obj->affected[j].modifier,
                   obj->obj_flags.bitvector, TRUE);
 
-  affect_total(ch);
-  calc_weight(ch);
+  if (recalc) {
+    affect_total(ch);
+    calc_weight(ch);
```
luciensadi, line 1681: `calc_weight()` should always be called, it's a safeguard function against the many (many) issues we've had around weight issues in the past.

src/handler.cpp
```diff
@@ -1723,8 +1734,10 @@ struct obj_data *unequip_char(struct char_data * ch, int pos, bool focus)
     GET_FOCI(ch)--;
     GET_OBJ_VAL(obj, 4) = 0;
   }
-  affect_total(ch);
-  calc_weight(ch);
+  if (recalc) {
+    affect_total(ch);
+    calc_weight(ch);
```
luciensadi, line 1739: Tagging second `calc_weight()` as well

src/limits.cpp
```diff
@@ -450,7 +450,7 @@ void remove_patch(struct char_data *ch)
         act("You resist the feeble effects of $p.", FALSE, ch, patch, 0, TO_CHAR);
       break;
     case PATCH_TRAUMA:
-      if (success_test(GET_REAL_BOD(ch) - (GET_BIOOVER(ch) > 0 ? (GET_BIOOVER(ch) + 1) / 2 : 0), GET_OBJ_VAL(patch, 1)) > 0)
+      if (success_test(GET_REAL_BOD(ch) - (GET_BIOOVER(ch) > 0 ? (GET_BIOOVER(ch) + 1) / 2 : 0), GET_PATCH_RATING(patch)) > 0)
```
luciensadi, line 453: These are my favorite types of changes, thanks for helping make the code more readable!

src/newdb.cpp
```diff
@@ -1105,22 +1105,22 @@ static bool save_char(char_data *player, DBIndex::vnum_t loadroom, bool fromCopy
...
-    obj_from_cyberware(temp);
+    obj_from_cyberware(temp, FALSE);
     obj_to_char(temp, player);
   }
 
   /* Remove their bioware to inventory. */
   for (temp = player->bioware; temp; temp = next_obj) {
     next_obj = temp->next_content;
-    obj_from_bioware(temp);
+    obj_from_bioware(temp, FALSE);
```
luciensadi, line 1123: If there isn't another guaranteed `affect_total()` call after this, you must put one in right before the SQL string is composed (~1170), otherwise character stats will balloon every time they log out and back in.

jdevnull:
Added a fix to stim patch behavior.

Previously, we stored the mental value before a stim patch activates, and then if the patch was removed early, we returned to the stored mental value. This could be exploited to negate predictable sources of stun: e.g., you have an upcoming nitro crash, you use a stim patch to save your current stun track, you then take the crash, and then remove the stim patch to immediately heal back to where you were.

src/newdb.cpp
```diff
@@ -1218,7 +1218,7 @@ static bool save_char(char_data *player, DBIndex::vnum_t loadroom, bool fromCopy
   /* Re-equip equipment. */
   for (i = 0; i < NUM_WEARS; i++) {
     if (char_eq[i])
-      equip_char(player, char_eq[i], i);
+      equip_char(player, char_eq[i], i, FALSE);
```
luciensadi, line 1221: Same deal here, you must have a guaranteed `affect_total()` run after this to ensure the character is set back up properly.

src/act.obj.cpp
```diff
@@ -3674,11 +3674,8 @@ void perform_remove(struct char_data * ch, int pos)
   if (GET_OBJ_TYPE(obj) == ITEM_HOLSTER)
     GET_HOLSTER_READY_STATUS(obj) = 0;
   /* add damage back from stim patches */
-  /* it doesn't do anything to keep track, */
-  /* so I'm just makeing it a mod mental damage to it */
   else if ( GET_OBJ_TYPE(obj) == ITEM_PATCH && GET_PATCH_TYPE(obj) == PATCH_STIM ) {
-    GET_MENTAL(ch) = GET_PATCH_STIMPATCH_ORIGINAL_MENTAL(obj);
-    GET_PATCH_STIMPATCH_ORIGINAL_MENTAL(obj) = 0;
+    GET_MENTAL(ch) = MAX(0, GET_MENTAL(ch) - (GET_PATCH_RATING(patch) * 100));
```
luciensadi, line 3678: What was the issue with the old code? I'm not sure why it was written the way it was, but I expect it was something to do with preventing underflow issues (which this change does as well)

jdevnull: Ah, you looked at it in the minute after I committed and before I could comment about it. The problem was that it's exploitable with predictable sources of stun. If you know stun damage is coming, you patch right before it happens and then remove the patch right after it happened.

src/newdb.cpp
jdevnull: Doesn't the affect_total in load_char prevent stats from ballooning?

luciensadi: No, because you're saving it with increased stats. Take the example of muscle aug 4 on a strength 1 character:

1. at function start, str = 5
2. after removal, str = 5 (previously, str = 1)
3. at SQL write, str = 5 (previously, str = 1)
4. at re-apply, str = 5 (previously, str = 1)
5. at function exit, str = 5

Then load that char from DB:

1. at function start, str = 5
2. at bioware apply, str = 9
3. at function exit, str = 9

That's my current understanding, although admittedly I haven't read the code in that area in a while. If I've gotten it wrong, let me know!

src/act.obj.cpp
luciensadi: Makes sense, appreciate the patch then.

src/newdb.cpp
jdevnull: It looks to me like we're saving real attributes, which we shouldn't be modifying anywhere. Still, this should be tested to confirm (and regrettably, I don't have a testing environment set up).

For some reason, it looks like we do save hacking and combat pool dice to the database, so those will be out of sync with their saved attributes, but affect_modify calculates those pool limits from scratch anyway.

src/newdb.cpp
jdevnull: There's an affect_total a few lines further down.

luciensadi: Ah, just below the code fold on GitHub. Thanks!

jdevnull:
Fixed dragonslayer conjuring bonus and added missing firebringer conjuring bonus.

jdevnull:
When somebody who actually is mortally wounded is dragged into a recovery room, and then they attempt to DIE, they end up having a positive physical track but still mortally wounded. This fixes that by also setting their position (depending on mental track).

jdevnull:
Character health boxes are tracked to 2 decimal places. The heal spell was capping successes to boxes, round down, thus preventing the last box of health from being healed. This fix changes this line to round up so that the last box of health can be healed. Note that this won't allow the heal spell to exceed max since spell_modify caps to GET_MAX_PHYSICAL.

This answers https://discord.com/channels/564618629467996170/788953927269613608/1153333686629052496

luciensadi:
Can you pull the master branch on both PRs please? That'll allow GH Actions to compile-test the code.

jdevnull:
Per commit e397678, cyberskulls are compatible with skin mods (dermal plating, dermal sheathing) and muscle replacements are compatible with cyberskulls and cybertorsos. This fix makes it so that it works regardless of which augment is applied first. This fix also counts orthoskin as a skin mod.

jdevnull:
Previously, there were inconsistencies in 'ware compatibilities, where some 'ware combinations were permitted when installed in one order but forbidden if the same 'ware were installed in a different order. This fix makes things consistent.

Need confirmation that the following is what is intended:
- muscle 'ware (replacements, toners, augs) are incompatible with cyberarms and cyberlegs
- bone lacings are incompatible with cyberarms, cyberlegs, and cybertorsos
- skin mods (dermal plating, sheathing, and orthoskin) are incompatible with cyberarms, cyberlegs, cybertorsos, and cyberskulls

jdevnull:
Setting net successes to zero does not actually prevent a counterstrike against an adept using distance strike. The fix here is to just return false instead.

jdevnull:
When the attacker is prone, the melee penalty should be applied to the attacker, not the defender.

jdevnull:
It's possible to be fighting while prone, so this changes the feedback message when POS_FIGHTING but not AFF_PRONE to something more reasonable.

luciensadi:
Merging for testing on buildport, thanks!

PR #711, opened 2023-09-25
Drug system fixes

jdevnull:
To prevent drug tolerance and edge values from spiraling out of control, this PR blocks edge/tolerance increases from involuntary doses, and caps tolerance and edge to MAX_DRUG_TOLERANCE and MAX_DRUG_EDGE (both currently set to 10).

Involuntary dosing can only occur if addicted, so this creates what is essentially an addiction+ state, IS_TAKING_INVOLUNTARY, that can be returned by GET_DRUG_ADDICT(ch, drug_id). When we're in this state, then the code that increases tolerance/edge is skipped. It can take multiple doses to activate drug effects. IS_TAKING_INVOLUNTARY is only set/kept if no other doses for the current activation are deliberately taken.

### DIFF
--- a/src/config.hpp
+++ b/src/config.hpp
@@ -277,6 +277,8 @@ extern const char *CHARACTER_DELETED_NAME_FOR_SQL;
 #define INVOLUNTARY_DRUG_PURCHASE_COST_MULTIPLIER              3
 #define AVG_HOURS_PER_TOLERANCE_TICKDOWN                       (24 * 7)
 #define MAX_ADDICTION_TEST_DIFFICULTY                          12
+#define MAX_DRUG_TOLERANCE                                     10
+#define MAX_DRUG_EDGE                                          10
 
 /////////////// OLC permissions configuration /////////////////////////
 #define LVL_FOR_SETTING_ZONE_EDITOR_ID_NUMBERS                 LVL_VICEPRES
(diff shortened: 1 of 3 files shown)


jdevnull:
The edge_delta isn't quite right. Do we also want involuntary doses to not count towards GET_DRUG_LIFETIME_DOSES?

jdevnull:
We don't want to count involuntary doses in lifetime doses, otherwise that screws with the calculation of edge_delta (which is what is added to edge/tolerance) the next time deliberate doses are taken.

jdevnull:
For now, put the edge reduction and testing / chem usage on the same timing. The condition that I commented out tries to use fix_factor but is broken, since that value stays constant. So once days_since_last_fix exceeded the the fix_factor, the testing / chem usage would happen every 2 minutes. We would be testing 24 times (if they managed to succeed every time) or use chems 48 times for every point of edge.

jdevnull:
Using `fix_factor `for the initial `GET_DRUG_LAST_WITHDRAWAL_TICK` for guided withdrawal lets the `days_since_last_fix >= GET_DRUG_LAST_WITHDRAWAL_TICK` condition match what I believe was intended for the testing / chem usage timing.

Setting `GET_DRUG_ADDICTION_EDGE = 2` when transitioning from "NOT_ADDICTED" to "IS_ADDICTED" ensures that forced withdrawal will always face at least one test.

High penalties when undergoing withdrawal makes a strong case for just letting such characters idle until withdrawal is complete. Reducing the penalties for undergoing withdrawal lets those characters run easier content and strengthens the argument to pay for guided withdrawal,

luciensadi (approved):
Looks good to me, I'll bring it in so we can test it.

PR #715, opened 2023-10-23
Fix adept/spell reaction/initiative compatibility

jdevnull:
Missed a rule in the previous PR: the adept improved reflexes power doesn't stack with other magical increases to reaction/initiative. This PR interprets that as letting the character have whichever bonus is higher for reaction and initiative independently, e.g. if an adept has the increase reflexes +3 spell cast on them (+3d6 init) and then activates adept improved reflexes 2 (+4 rea +2d6 init), then their total bonuses are +4 rea +3d6 init.

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -806,8 +806,10 @@ void affect_total(struct char_data * ch)
 
   if (GET_TRADITION(ch) == TRAD_ADEPT)
   {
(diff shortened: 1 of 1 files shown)


luciensadi:
Just realized I forgot to comment-- thanks for the fix!

PR #716, opened 2023-10-28
More drug fixes

jdevnull:
The previous version set edge = 2 when transitioning into addiction in order to guarantee that forced withdrawal would face at least one withdrawal test instead of automatically succeeding from edge 1. But that had the unintended consequence of also increasing the overall time to complete withdrawal. Also, because edge was always reduced in withdrawal before the withdrawal test, the removal of edge increases from forced drug use meant that a character undergoing forced withdrawal could fail every test and still successfully complete withdrawal. 

This PR fixes the above, returning to the previous edge = 1 when transitioning into addiction, and instead changes the order of operations for process_withdrawal; the test now happens first (or automatic success if in guided withdrawal), and edge is reduced only if successful.

Second, this PR now updates GET_DRUG_LAST_WITHDRAWAL_TICK alongside GET_DRUG_LAST_FIX on drug onset instead of when entering withdrawal, to more accurately reflect how Fix Factor is described in M&M pg 109.

This PR also replaces the nonfunctional time until withdrawal ends with the time until the next withdrawal test in status information.

### DIFF
--- a/src/act.informative.cpp
+++ b/src/act.informative.cpp
@@ -7116,15 +7116,15 @@ ACMD(do_status)
     }
     else if (GET_DRUG_ADDICT(targ, i) > 0) {
       if (GET_DRUG_STAGE(targ, i) == DRUG_STAGE_GUIDED_WITHDRAWAL) {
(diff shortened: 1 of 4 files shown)


luciensadi:
Looks good, thanks!

PR #717, opened 2023-11-01
More drug fixes

jdevnull:
1. MM causes confusion by using the same terms for both Tolerance Rating (target number for tolerance tests) and tolerance level (fail the tolerance test, now you need more doses to get high). The increase in tolerance for edge_delta > 0 is supposed to be an increase in Tolerance Rating (which is already implemented by adding GET_DRUG_ADDICTION_EDGE to tests), not an increase in tolerance level. (MM pg 108 vs pg 109).

2. The tolerance test is a body test only (MM pg 109).

3. After successfully reaching the DRUG_STAGE_GUIDED_WITHDRAWAL state, the character no longer needs to pass tests (except in the unlikely event where they run out of anticraving chems), so the house rule giving a 1 TN bonus doesn't seem significant.

4. Withdrawal tests only occur on hour ticks, so  get_time_until_withdrawal_test now just reports to within 2 irl minutes. This also solves the bug where if a withdrawing character is offline, they could end up seeing a large negative number when they log in.

### DIFF
--- a/src/drugs.cpp
+++ b/src/drugs.cpp
@@ -729,11 +729,8 @@ void attempt_safe_withdrawal(struct char_data *ch, const char *target_arg) {
     lose_nuyen(ch, treatment_cost, NUYEN_OUTFLOW_DRUGS);
   }
 
-  // Test for success. If they fail, they immediately seek out drugs.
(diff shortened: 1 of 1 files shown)


luciensadi:
LGTM, thanks!

PR #718, opened 2023-11-08
Add prestige races

TheVilest:
Adds Drakes, which are their base race in metahuman form. Shapeshifting is not implemented (NERP). Drakes:
- Suffer double essence loss from cyberware
- Suffer double bioindex loss from bioware
- Are only dual-natured in their drake form (shapeshifting to this form is not implemented)

Adds Ghouls, which template from base races. Ghouls:
- Are always dual-natured and perceiving, taking no TN from doing so
- Suffer double essence loss from cyberware
- Mild Allergy to Sunlight (+1 TN, same conditions as Night-Ones)

Adds Dragons in both Western, Eastern and Feathered Serpent varieties. Dragons:
- Cannot use cyber/bio at all

Adds Dryads, which are based on elves. Dryads:
- Have the highest base charisma
- Have cool NERP powers
- Suffer a +1 TN penalty to everything while in cities


src/awake.hpp
```diff
@@ -209,14 +209,26 @@
...
+#define RACE_DRAKE_ORK             27
+#define RACE_DRAKE_TROLL           28
+#define RACE_GHOUL_HUMAN           29
+#define RACE_GHOUL_DWARF           30
+#define RACE_GHOUL_ELF             31
+#define RACE_GHOUL_ORK             32
+#define RACE_GHOUL_TROLL           33
+#define RACE_ELEMENTAL             34
+#define RACE_SPIRIT                35
+#define RACE_PC_CONJURED_ELEMENTAL 36
```
luciensadi, line 227: The `RACE_ELEMENTAL`, `RACE_SPIRIT`, and `RACE_PC_CONJURED_ELEMENTAL` definitions shouldn't be changed, because doing so will change the races of all existing elementals / spirits / etc (they would become feathered serpents, human drakes, and dwarven drakes respectively).

Note also that all current `RACE_DRAGON` mobs (if there are any) would become Western dragons with this change, but I think that part is acceptable?

luciensadi: I see why you moved them though, since the "maximum valid player race" define kicks in there. The easier solution is probably to just remove the `X_VALID_PLAYER_RACE` definitions and convert the code that uses them from if-checks / for-loops to switch statements.

src/constants.cpp
```diff
@@ -3002,9 +3048,21 @@ const char *pc_race_types_decap[] =
     "ogre",
     "minotaur",
     "satyr",
-    "night-One",
+    "night-one",
```
luciensadi, line 3051: Nice find 😁

src/utils.cpp
```diff
@@ -477,7 +477,7 @@ int modify_target_rbuf_raw(struct char_data *ch, char *rbuf, size_t rbuf_len, in
 
   // If you're astrally perceiving, you don't take additional vision penalties, and shouldn't have any coming in here.
   if (SEES_ASTRAL(ch)) {
-    if (!skill_is_magic && IS_PERCEIVING(ch)) {
+    if (!skill_is_magic && IS_PERCEIVING(ch) && !IS_GHOUL(ch)) {
```
luciensadi, line 480: The effect of this is that ghouls can astrally perceive (bypassing vision penalties) without the +2 TN penalty, is that intended? That will make them very powerful as awakened since they can leave perceive on 24/7.

src/utils.cpp
```diff
@@ -514,6 +514,18 @@ int modify_target_rbuf_raw(struct char_data *ch, char *rbuf, size_t rbuf_len, in
     buf_mod(rbuf, rbuf_len, "Sunlight", 1);
     WRITEOUT_MSG("Sunlight Allergy", 1);
   }
+  if (IS_GHOUL(ch) && ((time_info.hours > 6) && (time_info.hours < 19)) && OUTSIDE(ch) && weather_info.sky < SKY_RAINING)
```
luciensadi, line 517: Is there a difference between this check and the RACE_NIGHTONE above? If not, recommend removing the duplicated code and changing the check for the previous one instead.

`if ((GET_RACE(ch) == RACE_NIGHTONE || IS_GHOUL(ch)) && ((time_info.hours...`

src/utils.cpp
```diff
@@ -514,6 +514,18 @@ int modify_target_rbuf_raw(struct char_data *ch, char *rbuf, size_t rbuf_len, in
...
+  {
+    base_target += 1;
+    buf_mod(rbuf, rbuf_len, "Sunlight", 1);
+    WRITEOUT_MSG("Sunlight Allergy", 1);
+  }
+  if (GET_RACE(ch) == RACE_DRYAD)
+    if (ROOM_FLAGGED(get_ch_in_room(ch), ROOM_INDOORS && ROOM_ROAD) || SECT(get_ch_in_room(ch)) == SPIRIT_HEARTH && SPIRIT_CITY)
```
luciensadi, line 524: You'll want a separate ROOM_FLAGGED check for each condition here. Same with SPIRIT_HEARTH vs SPIRIT_CITY.

`if (ROOM_FLAGGED(get_ch_in_room(ch), ROOM_INDOORS) || ROOM_FLAGGED(get_ch_in_room(ch), ROOM_ROAD) || SECT(get_ch_in_room(ch)) == SPIRIT_HEARTH || SECT(get_ch_in_room(ch)) == SPIRIT_CITY)`

You can think of `&&` as meaning "if both the condition on the right and the condition on the left are truthy". When you do `SECT(get_ch_in_room(ch)) == SPIRIT_HEARTH && SPIRIT_CITY`, the condition you're writing is true if "the comparison of sector and spirit hearth is true, AND if spirit_city is truthy". Truthiness in C is defined as not being equal to 0, which is always true since SPIRIT_CITY is really just a human-friendly way of writing `1`.

luciensadi (commented):
Thanks for the PR! Had a few thoughts, but I like where this is going.

src/awake.hpp
TheVilest: I actually tested this since I thought the same thing when writing it, but all pre-existing race type settings for mobs remained intact. Spirits are still Spirits, Elementals are still Elementals, PC Conjured Elementals are still PC Conjured Elementals. Just ran the code on my branch again to double check: Still using their old race type setting across the board

src/utils.cpp
TheVilest: Intended, actually. Ghouls are Dual Natured but receive no penalties associated with astral perception when doing physical tasks. It's one of the biggest boons they have (but also makes them susceptible to astral entities 24/7 since they can't turn it off). I was actually hoping to get your help on this since technically Mundane Ghouls should also have this, as well as IS_DRAGON types, so it needs some tweaking

src/utils.cpp
TheVilest: Nope, I just separated em it while testing, merged

src/utils.cpp
TheVilest: Yeh, my brain struggles with this one for some reason, hopefully this resolves it

src/awake.hpp
TheVilest: Poked at it some more, but I'm pretty sure mobs save race definition type instead of race definition value

luciensadi: Oh nice, I didn't realize that! Sounds good then.

src/newshop.cpp
```diff
@@ -113,16 +113,16 @@ bool is_ok_char(struct char_data * keeper, struct char_data * ch, vnum_t shop_nr
   if ((shop_table[shop_nr].races.IsSet(RACE_HUMAN) && GET_RACE(ch) == RACE_HUMAN) ||
```
luciensadi, line 113: Do we need DRAKE_HUMAN and GHOUL_HUMAN in here?

src/constants.cpp
```diff
@@ -2975,6 +3009,18 @@ const char *pc_race_types[] =
     "Night-One",
     "Dryad",
     "Dragon",
+    "Dragon",
+    "Serpent",
+    "Drake",
```
luciensadi, line 3014: Do these need to be identified as 'Drake (Human)' or 'Humanoid Drake' or something? We won't be able to differentiate them as staff if they're all the same string.

luciensadi:
Overall looks pretty solid. If I'm reading this right, ghouls and drakes:
- Are extensions to other races (templates layered on top?)
- Are dual (always perceive, take no penalties from doing so)
- Pay double essence for 'ware (both cyber and bio, or just cyber?)
Are there any other penalties for ghouls / drakes?

Dryads:
- Are basically elves
- Get a +1 TN penalty in urban settings
Are there any bonuses for dryads to offset this?

src/constants.cpp
TheVilest: Bastardized em so they fit

TheVilest:
> Overall looks pretty solid. If I'm reading this right, ghouls and drakes:
> 
> * Are extensions to other races (templates layered on top?)
> * Are dual (always perceive, take no penalties from doing so)
> * Pay double essence for 'ware (both cyber and bio, or just cyber?)
>   Are there any other penalties for ghouls / drakes?
> 
> Dryads:
> 
> * Are basically elves
> * Get a +1 TN penalty in urban settings
>   Are there any bonuses for dryads to offset this?

- Drakes are just their base race in their metahuman form, they can shapeshift into their drake form (which has its own rules and modifiers on top of that. But, that's whole other can of worms, that's intended to be NERP stuff).
- Ghouls and Dragons are always dual natured and perceiving (but take no TN from being so), Drakes use normal rules in metahuman form, but are dual natured only while in their drake form.
- Ghouls and Drakes suffer double Essence Loss from Cyberware. Drakes also suffer double Bioindex Loss (missing code for this atm). Dragons cannot use Cyber/Bioware period. (Missing code for this).

Dryads:
- Yep.
- Yep.
- Highest base Charisma, NERP powers make them Disney Princesses, rarity flex factor

src/newshop.cpp
TheVilest: > if ((shop_table[shop_nr].races.IsSet(RACE_HUMAN) && GET_RACE(ch) == RACE_HUMAN) || GET_RACE(ch) == RACE_DRAKE_HUMAN) ||

Little sketchy here, since it'll spit out any error with this additional syntax

luciensadi: The logic here is a mess, I need to rewrite this whole check some day. In the meantime, you're getting an error because your conditions are grouped wrong. Try:
```c++
if ((shop_table[shop_nr].races.IsSet(RACE_HUMAN) && (GET_RACE(ch) == RACE_HUMAN || GET_RACE(ch) == RACE_DRAKE_HUMAN || GET_RACE(ch) == RACE_GHOUL_HUMAN)) ||
```

TheVilest: Ah, I see. Should be fixed, I also dropped ghouls from the table in favour of the YES_GHOUL shop_flag

src/newshop.cpp
```diff
@@ -1212,6 +1225,11 @@ void shop_buy(char *arg, size_t arg_len, struct char_data *ch, struct char_data
         case RACE_ORK:
         case RACE_TROLL:
         case RACE_DWARF:
+        case RACE_DRAKE_HUMAN:
+        case RACE_DRAKE_ELF:
+        case RACE_DRAKE_ORK:
+        case RACE_DRAKE_TROLL:
+        case RACE_DRAKE_DWARF:
```
luciensadi, line 1232: Should ghouls avoid metavariant penalties too?

src/newshop.cpp
```diff
@@ -110,19 +112,19 @@ bool is_ok_char(struct char_data * keeper, struct char_data * ch, vnum_t shop_nr
     return TRUE;
   }
 
-  if ((shop_table[shop_nr].races.IsSet(RACE_HUMAN) && GET_RACE(ch) == RACE_HUMAN) ||
+  if ((shop_table[shop_nr].races.IsSet(RACE_HUMAN) && (GET_RACE(ch) == RACE_HUMAN || GET_RACE(ch) == RACE_DRAKE_HUMAN)) ||
```
luciensadi, line 115: Should ghouls be in here too? Is it possible to have a yes-ghoul-but-no-ghoul-trolls vendor?

src/newshop.cpp
TheVilest: Hmm, suppose not, since we opted for the YES_GHOUL function

src/newshop.cpp
TheVilest: Good call, re-added ghouls and tested yes-ghoul-but-no-ghoul-troll mix-and-match, should be working fine. Ghouls are now properly discriminated against by both their condition and their metatype, just in case they forgot that no one loves them.

src/newshop.cpp
```diff
@@ -279,6 +281,10 @@ bool uninstall_ware_from_target_character(struct obj_data *obj, struct char_data
   } else {
     obj_from_cyberware(obj);
     GET_ESSHOLE(victim) += GET_CYBERWARE_ESSENCE_COST(obj);
+    if (GET_RACE(victim) >= RACE_GHOUL_HUMAN && GET_RACE(victim) <= RACE_GHOUL_TROLL)
+    GET_ESSHOLE(victim) += GET_CYBERWARE_ESSENCE_COST(obj) *= 1;
```
luciensadi, line 285: `GET_CYBERWARE_ESSENCE_COST(obj) *= 1` means "the essence cost after multiplying by 1", so we don't need the `*= 1` component here. Same for below.

src/newshop.cpp
```diff
@@ -374,11 +391,22 @@ bool install_ware_in_target_character(struct obj_data *ware, struct char_data *i
     return FALSE;
   }
 
+  // Double Bioindex Loss for Drakes.
+  if (GET_OBJ_TYPE(ware) == ITEM_BIOWARE) {
+    int biocost = GET_BIOWARE_ESSENCE_COST(ware);
+    if (GET_RACE(recipient) >= RACE_DRAKE_HUMAN && GET_RACE(recipient) <= RACE_DRAKE_TROLL)
+      biocost = GET_BIOWARE_ESSENCE_COST(ware) *= 2;
```
luciensadi, line 398: I think the use of `*= 2` here is going to cause you issues down the line. What this code does is it doubles _and saves_ the essence cost of the 'ware, which means the object itself now reflects a doubled essence cost. I'm betting that if you buy 'ware and then immediately sell it, you'll pay the doubled cost and then have a 4x essence hole. Meanwhile, if you quit and reload first (causing the 'ware's cost to be loaded from default values), you'll have the expected 2x essence hole.

Recommend using `biocost *= 2;` to reflect that you're doubling the biocost.

luciensadi (commented):
Looking good! Is this complete to the point where I should review the whole thing against the requirements in the PR description, or is it still in progress?

src/newshop.cpp
TheVilest: Ah, gotcha, updated

src/newshop.cpp
TheVilest: Updated too prescribed changed, should be working fine now.

src/newshop.cpp
TheVilest: Since bioware essence value for the obj is doubled upon install for drakes, coupled it with a counterpart uninstall value setting of  x0.5 to handle the otherwise exponential bioware tumor growth issue

src/newshop.cpp
luciensadi: I'd remove this whole added block since you've written similar code at line 537.

src/newshop.cpp
```diff
@@ -503,6 +534,8 @@ bool install_ware_in_target_character(struct obj_data *ware, struct char_data *i
   // You must have the index to support it.
   else if (GET_OBJ_TYPE(ware) == ITEM_BIOWARE) {
     int esscost = GET_BIOWARE_ESSENCE_COST(ware);
+    if (GET_RACE(recipient) >= RACE_DRAKE_HUMAN && GET_RACE(recipient) <= RACE_DRAKE_TROLL)
+      esscost = GET_BIOWARE_ESSENCE_COST(ware) *= 2;
```
luciensadi, line 538: This isn't quite right-- what you've written is "double the essence cost of the bioware, set the bioware itself to take that doubled essence cost, and then take the result of that as our new `esscost` value". This will probably(?) work, but you're changing the base value of the item itself, so when you go to uninstall it later and double it there as well, you'll add back too much bioware essence.

Change to `esscost *= 2;` so you double the existing essence cost without touching the esscost of the 'ware.

luciensadi (changes requested):

src/newshop.cpp
TheVilest: 395 is for x2 bioware cost, 537 is for x2 cyberware cost, drakes have double loss for both cyber/bio, or am I not seeing something?

Also, if I change it to esscost *= 2; it only returns half the bioindex on uninstall

luciensadi: 537 is for bioware cost because the 535 ifcheck is for `ITEM_BIOWARE`. 395 _should_ have no effect because `biocost` is not used outside of your scope, but you're doubling the actual bioware object's essence in the process.

The fact that it's only returning half the bioindex on uninstall means that you're missing code to double the bioindex returned on uninstall. Try restoring your previous code, installing bioware, and then doing a copyover before uninstalling it-- you'll only get half back there as well, as your change to the bioware cost is not persisted across copyovers.

src/newshop.cpp
```diff
@@ -276,9 +278,16 @@ bool uninstall_ware_from_target_character(struct obj_data *obj, struct char_data
     obj_from_bioware(obj);
     GET_INDEX(victim) -= GET_CYBERWARE_ESSENCE_COST(obj);
```
luciensadi, line 279: This is where the character's bioware index is updated. Putting your essence cost change after this means that that change has no effect. Move your essence cost ifcheck before this instead.

Also, can you change this to `GET_BIOWARE_ESSENCE_COST` instead of cyberware? They mean the same thing, but this typo makes it harder to read.

src/newshop.cpp
TheVilest: Ok, one last shot, see if that does the trick

src/newshop.cpp
TheVilest: Swapped to CYBER, shoulda realized it sooner they're the same thing

PR #719, opened 2023-11-11
Fix rounding sets down too early

jdevnull:
See: https://discord.com/channels/564618629467996170/1172680716274901042

If instead of rounding the ratings of armor sets down early, we can multiply all non-set armor ratings by 100 when figuring out layering. We then divide by 100 at the end.

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -664,13 +664,14 @@ void affect_total(struct char_data * ch)
           else if (GET_WORN_BALLISTIC(worn_item) || GET_WORN_IMPACT(worn_item)) {
             int bal = 0, imp = 0;
 
-            // Remember that matched set items are 100x their expected value-- compensate for this.
+            // Remember that matched set items are 100x their expected value.
+            // Multiply non-set items here, divide the total later so we don't truncate too early.
             if (GET_WORN_MATCHED_SET(worn_item)) {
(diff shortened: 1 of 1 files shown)


luciensadi:
I'm in favor of correcting errors in the code, but this one will have a measurable effect on balance (namely, increasing most PCs armor by up to 1/1 while not changing NPC armor levels). Is the buff to PC armor levels needed under the current balancing paradigm?

jdevnull:
This doesn't affect PCs on the high end (qui 10+) since with the three big sets, it's better to replace both the body and under slots with non-set pieces, so the remaining set pieces total whole integers.

That said, it occurs to me that if we're keeping fractions, then there's a good argument that the qui requirement to avoid penalties should also not be rounded down. If Jexams' example is an elf, then they'd need to give up another set piece to avoid penalties.

Not sure if it'd be better to just close this PR or to also fix the qui requirement to match. Closing this leaves us at status quo, where the math is less intuitive for players trying to math out their outfits. Fixing the qui requirement to match would reduce the clothing options that lower qui characters can choose from while still avoiding bulkiness penalties.

luciensadi:
Given that the qui requirement fix would negatively impact lower qui characters and potentially break existing builds, I think we can leave this one at status quo. Appreciate the PR though!

PR #720, opened 2023-11-16
Miscellaneous Fixes

jdevnull:
1. MM pg 30 states that move-by-wire is not compatible with any other reaction or initiative enhancing cyberware or bioware. Also, there was previously no check preventing them from stacking with magic/adept reaction/initiative bonuses. This PR ensures that they don't stack in either case.

2. When leaving chargen, remaining unspent nuyen are dumped into a housing card in increments of 5k nuyen. This PR lets this happen when the character has exactly 5k nuyen remaining.

3. The absence of decrease (non-cybered) attribute spells from the chargen health spell instructor appears to be an oversight, given that they already teach decrease cybered attribute, increase (non-cybered) attribute, and increase cybered attribute spells. The decrease (non-cybered) attribute spells can be used as debuffs against NPC opponents and would allow healer totem shamanists to have an offensive spell option. If anything, it should be the decrease cybered versions that should be absent, since they are only really used to make subsequent increase cybered spells easier (where using ritualcast instead would be preferred anyway).

4. The previous attempt at preventing persona programs from using storage memory in store-bought decks failed because it was placed in a block requiring GET_OBJ_TYPE(cont) == ITEM_DECK_ACCESSORY. This PR moves the fix to the right place. This answers: https://discord.com/channels/564618629467996170/788953927269613608/1172638603009081464


luciensadi (approved):
LGTM, thanks!

PR #723, opened 2023-12-02
Ground Control to Major Tom

TheVilest:
- Added missing Autopilot vehicle mod supports
- Vehicle's Autopilot value gives additional bonus dice to flight tests equal to its rating
- Added negative rollback prevention to Sensor/Autopilot vehicle mods


src/zoomies.cpp
```diff
@@ -463,9 +464,9 @@ int flight_test(struct char_data *ch, struct veh_data *veh) {
       break;
   }
   
-  int dice = get_skill(ch, skill_num, tn, rbuf, sizeof(rbuf));
+  int dice = get_autopilot + get_skill(ch, skill_num, tn, rbuf, sizeof(rbuf));
   int successes = success_test(dice, tn, ch, "flight-test");
-  snprintf(ENDOF(rbuf), sizeof(rbuf) - strlen(rbuf), ". Dice %d, TN %d: %d success%s.", dice, tn, successes, successes == 1 ? "" : "es");
+  snprintf(ENDOF(rbuf), sizeof(rbuf) - strlen(rbuf), ". Dice %d, Autopilot Rating %d, TN %d: %d success%s.", dice, get_autopilot, tn, successes, successes == 1 ? "" : "es");
```
luciensadi, line 469: This will be confusing for players since 'dice' already contains the autopilot rating. Separate out the skill dice (int skill = x, int pilot = y, int successes = success_test(x + y, tn...)) and report on it that way to avoid confusion.

src/handler.cpp
```diff
@@ -303,7 +303,7 @@ void affect_veh(struct veh_data *veh, byte loc, sbyte mod)
       veh->armor += mod;
       break;
     case VAFF_SEN:
-      veh->sensor = mod;
+      veh->sensor += mod;
```
luciensadi, line 306: Is += for mod intentional here? This will change vehicle sensor ratings for any vehicle with a sensor mod currently installed.

src/zoomies.cpp
TheVilest: Should be easier to digest

src/handler.cpp
TheVilest: Sensors are entirely NERP atm so can just drop it, currently if you install and uninstall a sensor mod without this + it'll roll back to a sensor value of 252 (Which, again, does nothing, so borked if you do, borked if you don't)

src/zoomies.cpp
luciensadi: This actually won't work, because 467's `get_skill(ch, skill_num + get_autopilot, ...)` means that the skill rolled will vary based on what your autopilot rating is. For instance, if you're driving a truck (`SKILL_PILOT_TRUCK (63)`) with autopilot 5, you will get the dice for your skill in `63 + 5` aka `SKILL_DANCING (68)`.

Do this instead:
```c++
int dice = get_skill(ch, skill_num, tn, rbuf, sizeof(rbuf));
int successes = success_test(dice + get_autopilot, tn, ch, "flight-test");
snprintf(ENDOF(rbuf), sizeof(rbuf) - strlen(rbuf), ". Skill %d, Autopilot Rating %d, TN %d: %d success%s.", dice, get_autopilot, tn, successes, successes == 1 ? "" : "es");
```

src/handler.cpp
luciensadi: Why do vehicles suck so much ;-;

src/zoomies.cpp
TheVilest: We didn't realize that Vehicle Code was the real End Game Boss all along. Innocence has been lost.

luciensadi (approved):

luciensadi:
Looks good to me, thanks!

PR #724, opened 2023-12-17
Cyberware/bioware incompatibilities

jdevnull:
Add guard for metabolic arrester incompatibilities. According to MM pg 67, the metabolic arrester is not compatible with the adrenal pump or suprathyroid gland.

Remove guard preventing datajack + eye datajack. Multiple datajacks is not unusual in canon (SR3 pg 298), and the `get_datajack` function is able to handle multiple datajacks.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -2054,10 +2054,18 @@ bool biocyber_compatibility(struct obj_data *obj1, struct obj_data *obj2, struct
   else if (bio1 && bio2) {
     switch (GET_BIOWARE_TYPE(bio2)) {
       case BIO_CALCITONIN:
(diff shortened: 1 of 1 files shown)


luciensadi:
Sorry, just noticed this PR. I think the datajack code is worth keeping, because there's no coded reason to _have_ multiple datajacks, so it will just risk costing essence when someone doesn't need to spend it. I don't want people picking up essence holes because they did their operations out of order.

jdevnull:
Technically, different datajacks are blocked by different types of clothing (e.g., helmets, gloves), so there can be a very minor benefit in ease of jacking in when switching out equipment configurations, in addition to RP support.

Though it looks like normal datajacks will still be incompatible with induction datajacks since they're both type cyb_datajack. I'll go ahead and revert that bit since it's incomplete anyway.

luciensadi (approved):
LGTM, thanks!

PR #725, opened 2023-12-27
MacOS instructions for Boost

danhealy:
Adds instructions for installing boost and modifies the Makefile `LIBS` to use it, for Mac


luciensadi:
Thanks for the PR!

PR #726, opened 2024-01-02
Upload and Download commands to transfer data to headware memory

danhealy:
A set of commands, `do_upload_headware` and `do_download_headware` meant to be used in meat space as `upload <file>` and `download <file number>`.

Both commands require the character to have:
- headware memory
- datajack
- a working cyberdeck carried or worn (the helper `find_cyberdeck` is created for this check)

`upload <file>` will take any `ITEM_DECK_ACCESSORY` file and insert it into your headware memory, if there is enough available space.  This is intended to be used with data files or photos, not skillsofts.

`download <file number>` reverses the process, it removes the file from your headware and places it in your inventory.

Example:
```
l
Eye Modification Cyberware (Peaceful)
   Eye modification is a booming business in 2064, anything from cosmetic
modification to retinal duplicators to image magnification. If eye
modifications are bought in a package, essence reductions apply, so it is
usually best to do it this way. However, not always can you find your desired
modification in a package and will have to buy it separately. All vision
modifiers given by cyberware is treated as natural vision, and as such, mages
with electronic magnification cannot cast any spells that require line of
sight without projecting or perceiving. 
Obvious exits:
South - [60517] Cyberware Central
Up    - [60629] Eye Modification Alphaware
Doc Bell stands here, poking at a human eyeball in a jar. 
...he has a few things for sale. Use the LIST command to see them.

<10P 10M> 
photo
You take a photo.

<10P 10M> 
memory
Headware Memory Contents (73/75):
  1) a photo of Eye Modification Cyberware    (1 MP) 
  2) a photo of Cyberware Central             (1 MP) 

<10P 10M> 
view 1
You need headware memory and an image link to do this.

<10P 10M> 
download 1
You transfer a photo of Eye Modification Cyberware through a Fairlight Executive into your hands.

<10P 10M> 
memory
Headware Memory Contents (74/75):
  1) a photo of Cyberware Central             (1 MP) 

<10P 10M> 
i
You are carrying:
a photo of Eye Modification Cyberware
a Fairlight Executive
a spyware disk
a large plastiboard box
an Athletics 5 ActiveSoft
a radio
a Neophyte Guild housing subsidy card
a map of Seattle

You're not using any equipment.

You have no ammo in your pockets.

<10P 10M> 
exa photo
Eye Modification Cyberware
   Eye modification is a booming business in 2064, anything from cosmetic
modification to retinal duplicators to image magnification. If eye
modifications are bought in a package, essence reductions apply, so it is
usually best to do it this way. However, not always can you find your desired
modification in a package and will have to buy it separately. All vision
modifiers given by cyberware is treated as natural vision, and as such, mages
with electronic magnification cannot cast any spells that require line of
sight without projecting or perceiving. 
Doc Bell stands here, poking at a human eyeball in a jar. 

You have no training in Police Procedures, but you guess that a photo of Eye Modification Cyberware
is completely legal.
A photo of Eye Modification Cyberware is in excellent condition.

<10P 10M> 
upload photo
You transfer a photo of Eye Modification Cyberware into your headware memory, through a Fairlight
Executive.

<10P 10M> 
memory
Headware Memory Contents (73/75):
  1) a photo of Eye Modification Cyberware    (1 MP) 
  2) a photo of Cyberware Central             (1 MP) 
```

### DIFF
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -4168,19 +4168,18 @@ ACMD(do_memory)
 // Looks for any working cyberdeck in the character's inventory or worn equipment.
 struct obj_data *find_cyberdeck(struct char_data *ch)
 {
-  struct obj_data *deck;
   if (!ch) {
     mudlog("SYSERR: Got NULL ch to find_cyberdeck()!", ch, LOG_SYSLOG, TRUE);
     return NULL;
   }
 
-  for (deck = ch->carrying; deck; deck = deck->next_content) {
+  for (struct obj_data *deck = ch->carrying; deck; deck = deck->next_content) {
     if (OBJ_IS_VALID_CYBERDECK(deck)) {
       return deck;
     }
   }
 
-  for (int wear_idx = 0; !deck && wear_idx < NUM_WEARS; wear_idx++) {
+  for (int wear_idx = 0; wear_idx < NUM_WEARS; wear_idx++) {
     if (OBJ_IS_VALID_CYBERDECK(GET_EQ(ch, wear_idx))) {
       return GET_EQ(ch, wear_idx);
     }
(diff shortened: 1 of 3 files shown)


src/act.other.cpp
```diff
@@ -4165,6 +4165,128 @@ ACMD(do_memory)
   }
 }
 
+// Looks for any working cyberdeck in the character's inventory or worn equipment.
+struct obj_data *find_cyberdeck(struct char_data *ch)
+{
+  struct obj_data *deck;
```
luciensadi, line 4171: Nit: Unless you're going to use this outside of its for-loop scope, this is best defined as part of the for-loop instantiation (e.g. `for (struct obj_data *deck = ch->carrying...`). Keeping vars to the scope where they're used makes maintaining the code easier in the long term.

src/act.other.cpp
```diff
@@ -4165,6 +4165,128 @@ ACMD(do_memory)
...
+{
+  struct obj_data *deck, *jack, *memory;
+  skip_spaces(&argument);
+  if (!*argument) {
+    send_to_char("Which file number do you want to transfer out of headware memory?\r\n", ch);
+    return;
+  }
+
+  for (memory = ch->cyberware; memory; memory = memory->next_content) {
+    if (GET_OBJ_VAL(memory, 0) == CYB_MEMORY) {
```
luciensadi, line 4202: Nit: This is correct, but for readability please use `GET_CYBERWARE_TYPE()` instead of a direct objval reference. I'm slowly replacing those old references as I see them in the code, but haven't gotten to them all yet.

src/act.other.cpp
```diff
@@ -4165,6 +4165,128 @@ ACMD(do_memory)
...
+
+  if (!(deck = find_cyberdeck(ch))) {
+    send_to_char(ch, "You need to have a working cyberdeck to transfer data out of headware memory.\r\n");
+    return;
+  }
+
+  int x = atoi(argument);
+  struct obj_data *obj = memory->contains;
+  for (; obj; obj = obj->next_content) {
+    if (!--x)
```
luciensadi, line 4224: If a user gives you `download -1`, you'll have some pretty nasty behavior here. Recommend an error case after the `atoi()` call, something like `FAILURE_CASE(x <= 0, "You must provide a positive integer.")`

You may also need to switch to `x--` rather than `--x`, unless you're listing a zero-indexed list to the player with a +1 index. I think the current state is that `download 1` would get file 0, and `download 0` would run into the negatives.

src/act.other.cpp
```diff
@@ -4165,6 +4165,128 @@ ACMD(do_memory)
...
+  }
+
+  obj = get_obj_in_list_vis(ch, argument, ch->carrying);
+  if (!obj) {
+    send_to_char(ch, "You aren't carrying that file.\r\n");
+    return;
+  } else if (GET_OBJ_TYPE(obj) != ITEM_DECK_ACCESSORY) {
+    send_to_char(ch, "You can only upload photos and data.\r\n");
+    return;
+  }
```
luciensadi, line 4276: Recommend adding further guards here since `ITEM_DECK_ACCESSORY` covers a lot of ground (cookers, computers, etc). Here's the list from `awake.hpp`:

```c++
#define TYPE_FILE            0
#define TYPE_UPGRADE         1
#define TYPE_COMPUTER        2
#define TYPE_PARTS           3
#define TYPE_COOKER          4
```

You probably want to restrict it only to `TYPE_FILE`.

src/act.other.cpp
```diff
@@ -4165,6 +4165,128 @@ ACMD(do_memory)
...
+
+  if (GET_CHIP_SIZE(obj) > GET_CYBERWARE_MEMORY_FREE(memory)) {
```
luciensadi, line 4278: It's been a while since I looked at the headware memory code, so I don't recall the specifics of how compression works, but you checked the compression factor when uninstalling it and don't check it here. Is this intended? Does that clear when it's uninstalled somehow?

luciensadi (commented):
Thanks for the PR! Had a few thoughts on it, with some functionality questions / suggestions as well as some minor formatting changes (marked 'Nit').

src/act.other.cpp
danhealy: 👍 Ah yeah in this case it's referenced in both for loops, but actually it doesn't need to be

src/act.other.cpp
danhealy: This snippet was pulled from the code in `do_delete`, it seems to work as intended but I agree it's not perfect, I'll try rewriting it

src/act.other.cpp
danhealy: I'll have to double check this.  I assumed that compression is only possible in place, that is, that it reduces the used memory after the full amount has already occupied the space.  So when removing the file you need to adjust the used memory based on compression, but when adding it, the compression is not considered

danhealy: `help compress` says `Files that are compressed must be decompressed before they are used.` so I'm going to remove the compression math and just ensure that compressed files can't be downloaded

src/act.other.cpp
danhealy: I think adding the `FAILURE_CASE` is sufficient, the list is zero indexed but displayed starting with 1.

```c
  int x = atoi(argument);
  FAILURE_CASE(x <= 0, "You must provide a positive file index number.")
  struct obj_data *obj = memory->contains;
  for (; obj; obj = obj->next_content) {
    if (!--x)
      break;
  }
  ```

luciensadi (approved):
Looks good to me, thanks!

luciensadi:
@danhealy how do you want to be credited in the patch notes? Please supply a value for X: `(thanks X)` e.g. `(thanks Khai)`

PR #727, opened 2024-01-03
Minor fixes

jdevnull:
According to the FLEE helpfile:
> The FLEE command takes half as long as attempting to walk away in a specific direction.

This PR changes the wait_state in do_flee to match.


luciensadi:
LGTM, thanks!

PR #729, opened 2024-01-09
Fix spirit confusion force

jdevnull:
From newmagic.cpp line 4718, it looks like a spirit's confusion power should provide a modifier equal to 1/3 of the spirit's force. This PR changes the modifier in modify_target_rbuf_raw to match.


luciensadi (approved):
LGTM, thanks!

PR #734, opened 2024-03-03
If you can't see ch, you don't see their movement

jdevnull:
Previously, `should_tch_see_chs_movement_message` didn't check for invisibility. This meant that guards could become alerted even though they couldn't see/attack the character. The character would also see messaging as if the guard did see through their invis. This PR adds a `can_see_through_invis` check to `should_tch_see_chs_movement_message`.

This addresses: [https://discord.com/channels/564618629467996170/788953927269613608/1211114620618805313](url)

Alternatively, if we want guards to be able to become alerted through contextual clues despite not being able to see the char ("hey, did somebody just come through that door/bump the trash can/kick a pebble"), we might want to fix the messaging, and also make it a roll instead of being automatic.

### DIFF
--- a/src/act.movement.cpp
+++ b/src/act.movement.cpp
@@ -205,6 +205,11 @@ bool should_tch_see_chs_movement_message(struct char_data *tch, struct char_data
     return FALSE;
   }
 
+  // Check for invisibility.
+  if (!can_see_through_invis(tch, ch)) {
+    return FALSE;
+  }
+
   // Check for stealth and other person-to-person modifiers.
   if (we_care_about_sneaking && IS_AFFECTED(ch, AFF_SNEAK)) {
     int dummy_tn = 2;
--- a/src/act.movement.cpp
+++ b/src/act.movement.cpp
@@ -245,8 +245,11 @@ bool should_tch_see_chs_movement_message(struct char_data *tch, struct char_data
     // Don't allow negative results.
     open_test_result = MAX(open_test_result, 0);
 
-    // Vision modifiers, wound penalties, and other distractions.
-    int tn_modifiers = modify_target_rbuf_raw(tch, rbuf, sizeof(rbuf), get_vision_penalty(tch, in_room, rbuf, sizeof(rbuf)), FALSE);
+    // Vision penalties.
+    int vis_tn_modifier = calculate_vision_penalty(tch, ch);
+
+    // Other penalties.
+    int phys_tn_modifier = modify_target_rbuf_raw(tch, rbuf, sizeof(rbuf), vis_tn_modifier, FALSE);
 
     // House rule: Stealth/silence spells add a TN penalty to the spotter, up to 4.
     int stealth_spell_tn = get_spell_affected_successes(ch, SPELL_STEALTH);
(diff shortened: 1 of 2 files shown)


luciensadi:
I'm not sure a blanket immunity to being spotted is the right move here. Someone who isn't sneaking will make noise on entry, and someone who is sneaking should still have the potential to be heard during movement, just at a max visibility TN.

jdevnull:
What should the mechanics of this look like? If they have a stealth spell active, would that be enough? If they don't have a stealth spell active and are not sneaking, then what percentage odds should we aim for?

jdevnull:
Max visibility TN modifier is +8. Assuming no other contributions and rolling int for perception, we're looking at 45% chance at 4 dice, 70% chance at 8 dice. I believe elite guards have 8 dice.

jdevnull:
Revised `should_tch_see_chs_movement_message` so that there is now always a check, regardless of sneaking or invis. Specifically, invis without sneaking will no longer be an automatic detection, and invis + sneaking will no longer be an automatic non-detection. In addition, conditions (darkness spell, wound penalties, etc) can now make noticing movement non-trivial even when the moving character is neither invisible nor sneaking.

Also improved messaging for the case where tch notices ch despite tch not actually being able to see ch.

A concern is that there's a lot of movement that happens in the mud, so this could be a significant increase in processing. Added an escape case should both tch and ch be NPCs, though there's a question as to whether this case should result in always TRUE or FALSE.

jdevnull:
Added spirit concealment power and fixed how I was handling invis vs other penalties.

src/act.movement.cpp
```diff
@@ -196,81 +197,99 @@ bool should_tch_see_chs_movement_message(struct char_data *tch, struct char_data
...
 
-    // We rolled at least one hit above our open test TN? NOW we check if we can see them.
-    if (spotted_movement && CAN_SEE(tch, ch)) {
+    // Failed to notice them.
+    if (!spotted_movement) {
+      return FALSE;
+    }
+
+    // They were sneaking but we noticed.
+    if (IS_AFFECTED(ch, AFF_SNEAK) && spotted_movement) {
```
luciensadi, line 281: Let's make this check a `spotted_movement && (IS_AFFECTED(ch, AFF_SNEAK) || !CAN_SEE(ch))` since hearing an invis moving is also sus.

src/act.movement.cpp
```diff
@@ -196,81 +197,99 @@ bool should_tch_see_chs_movement_message(struct char_data *tch, struct char_data
...
+  }
 
   // Failed to see from vehicle.
   if (tch->in_veh && (tch->in_veh->cspeed > SPEED_IDLE && success_test(GET_INT(tch), 4) <= 0)) {
     return FALSE;
   }
 
   // Check for stealth and other person-to-person modifiers.
-  if (we_care_about_sneaking && IS_AFFECTED(ch, AFF_SNEAK)) {
+  if (we_care_about_sneaking) {
```
luciensadi, line 216: GitHub PR review isn't great for long changes. Am I correct in assuming that having the sneak check removed here means you can fail to spot someone you can see who's entering the room even if they're not sneaking? That doesn't seem like a desired outcome.

src/act.movement.cpp
jdevnull: The "can't see but we noticed case" looks like its handled from the result of the `should_tch_see_chs_movement_message` call on line 440, which in turn calls `process_spotted_invis`. Should that be moved from that part of the code to inside `should_tch_see_chs_movement_message` itself?

src/act.movement.cpp
jdevnull: That is correct. If ch is affected by invis, spirit concealment, or stealth/silence spells, and/or if tch is wounded, sustaining spells, non-invis vision penalties, or is otherwise distracted, then my thinking is that it should be possible for tch to fail to notice ch, even when ch is not actively sneaking.

If none of those apply, then 4 perception vs TN 2 is a 99.9% chance of success.

src/act.movement.cpp
```diff
@@ -196,81 +197,99 @@ bool should_tch_see_chs_movement_message(struct char_data *tch, struct char_data
   if (tch == ch || PRF_FLAGGED(tch, PRF_MOVEGAG) || !AWAKE(tch) || IS_IGNORING(tch, is_blocking_ic_interaction_from, ch))
     return FALSE;
 
+  // If both tch and ch are NPCs, we don't care.
+  if (IS_NPC(tch) && IS_NPC(ch)) {
```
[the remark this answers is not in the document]
jdevnull: Ah, you're correct. IS_PROJECT actually requires that IS_NPC is true.

src/act.movement.cpp
luciensadi: Unfortunately I'm gonna get bug reports from both the hidden person and the non-hidden person from this behavior. "I got complained at by X for sneaking around when I wasn't" isn't a fun outcome for staff or players. Let's only have it where sneaking characters have a chance to go unseen when moving.

jdevnull: There's another check below, at line 287 that comes into play. The "we_care_about_sneaking" flag is admittedly not well named, since it's actually "should we roll for perception". If NPC tch succeeds at a perception roll, there are 3 possible results depending on whether ch is doing something suspicious:
1. Ch is sneaking, then NPC tch becomes alert (as part of `should_tch_see_chs_movement_message`).
2, Ch is invis, then we do `process_spotted_invis` (as called by `do_simple_move`... should this happen inside `should_tch_see_chs_movement_message` instead?).
3. Ch is neither sneaking nor invis, so NPC doesn't react. This should more properly be "Ch is not suspicious", which may or may not include other spells and effects like spirit concealment and poltergeist, but that's a different discussion.

luciensadi (approved):

PR #736, opened 2024-03-16
A few initiative things

jdevnull:
1. Fix for wound modifiers applying twice to initiative
2. Add messaging for the case where a character rolls zero or negative initiative (only possible through a low roll + wound penalties).

These two address: https://discord.com/channels/564618629467996170/788953927269613608/1218251123249647756

3. With the now available delta grade move-by-wires cyberware, they are now feasible for character builds, but they were missing a key feature. MBW3 and MBW4, by the book, are supposed to gain extra actions in the first one or two passes through an initiative turn. We can simulate that by adding 10/20 to their initiative roll, though this is mechanically weaker because the extra actions will occur at the end of an initiative turn instead of near the beginning.

### DIFF
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -5705,8 +5705,9 @@ void roll_individual_initiative(struct char_data *ch)
 {
   if (AWAKE(ch))
   {
+    // this also applies wound modifiers
     GET_INIT_ROLL(ch) = roll_default_initiative(ch);
-    GET_INIT_ROLL(ch) -= damage_modifier(ch, buf, sizeof(buf));
+
     if (AFF_FLAGGED(ch, AFF_ACTION)) {
       GET_INIT_ROLL(ch) -= 10;
       AFF_FLAGS(ch).RemoveBit(AFF_ACTION);
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -5730,6 +5730,12 @@ void roll_individual_initiative(struct char_data *ch)
       else
         GET_INIT_ROLL(ch) += 20;
     }
+
+    // If you rolled zero or negative initiative, let it be known
+    if (GET_INIT_ROLL(ch) <= 0) {
+      act("You struggle to act as your wounds limit your actions! [OOC: You rolled an initiative of zero or less]", FALSE, ch, 0, 0, TO_CHAR);
+    }
+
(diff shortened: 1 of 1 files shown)


src/fight.cpp
```diff
@@ -5705,18 +5705,37 @@ void roll_individual_initiative(struct char_data *ch)
...
       GET_INIT_ROLL(ch) -= 10;
       AFF_FLAGS(ch).RemoveBit(AFF_ACTION);
     }
+
+    // By the book, MBW 3 adds an extra action to their first pass and 4 adds an additional action to their second.
+    // This implementation is mechanically weaker since chars will get the extras at the end of the turn.
+    for (obj_data *cyber = ch->cyberware; cyber; cyber = cyber->next_content) {
+      if (GET_CYBERWARE_TYPE(cyber) == CYB_MOVEBYWIRE) {
+        if (GET_CYBERWARE_RATING(cyber) == 3)
+          GET_INIT_ROLL(ch) += 10;
```
luciensadi, line 5721: +10 is a full extra pass, how about +5 for 3 and +10 for 4?

src/fight.cpp
```diff
@@ -5728,7 +5747,7 @@ void roll_individual_initiative(struct char_data *ch)
         && !MOB_FLAGGED(FIGHTING(ch), MOB_INANIMATE)
         && GET_MOBALERT(FIGHTING(ch)) == MALERT_CALM
         && success_test(GET_REA(ch), 4) > success_test(GET_REA(FIGHTING(ch)), 4)) {
-      GET_INIT_ROLL(FIGHTING(ch)) = 0;
+      GET_INIT_ROLL(FIGHTING(ch)) <= 0;
```
luciensadi, line 5750: Typo? This changed from an assignment to a no-op comparison.

jdevnull: Oops, yeah, not sure how that happened.

src/fight.cpp
jdevnull: Yeah, a full pass / 2 full passes are intentional, to actually allow the additional attack / 2 attacks as seems to be intended by the book. Part of the more generous reading is to try to get players to actually consider this 'ware as a feasible option.

Though maybe the potential for 2 more rounds of "out of initiative" messages outweighs that, so I'll tone that down.

luciensadi (approved):
LGTM, thanks!

PR #737, opened 2024-04-01
Minor fixes

jdevnull:
Repair costs at Jim's FixIt were scaling by current condition percentage (i.e., the more damaged the item, the cheaper the repair cost). PR changes this to scale by the actual damage percentage.

For vehicle repairs, `get_skill` was being called before determining the TN from vehicle damage and other modifiers, meaning that a character could always default regardless of the TN. PR swaps the order of operations so that repairing wrecks require actually having the appropriate skill.

### DIFF
--- a/src/act.drive.cpp
+++ b/src/act.drive.cpp
@@ -1100,9 +1100,9 @@ ACMD(do_repair)
     return;
   }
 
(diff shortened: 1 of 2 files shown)


luciensadi:
LGTM, thanks!

PR #738, opened 2024-04-10
Attempting to make autokill toggle more consistent (fingers crossed)

inb4jex:
All I've done so far is change the value that `GET_PHYSICAL(victim)` checks against from 0 to 1 to hopefully address a single deadly wound (that morts a mob) not properly triggering `stop_fighting(ch)`.

I do kind of want to explore the necessity of `(GET_MENTAL(victim) < 100` but I haven't quite figured out what the implications of changing/removing that would be.

### DIFF
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -3562,7 +3562,7 @@ bool raw_damage(struct char_data *ch, struct char_data *victim, int dam, int att
       break;
   }
 
-  if (GET_MENTAL(victim) < 100 || GET_PHYSICAL(victim) < 0)
+  if (GET_MENTAL(victim) < 100 || GET_PHYSICAL(victim) < 1)
     if (FIGHTING(ch) == victim)
       if (GET_POS(victim) == POS_DEAD || !IS_NPC(victim) || PRF_FLAGGED(ch, PRF_NOAUTOKILL)) {
         stop_fighting(ch);
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -3562,7 +3562,7 @@ bool raw_damage(struct char_data *ch, struct char_data *victim, int dam, int att
       break;
   }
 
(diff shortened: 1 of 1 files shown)


src/fight.cpp
```diff
@@ -3562,7 +3562,7 @@ bool raw_damage(struct char_data *ch, struct char_data *victim, int dam, int att
       break;
   }
 
-  if (GET_MENTAL(victim) < 100 || GET_PHYSICAL(victim) < 0)
+  if (GET_MENTAL(victim) < 100 || GET_PHYSICAL(victim) < 1)
```
luciensadi, line 3565: GET_MENTAL and GET_PHYSICAL are both multiplied by 100 to simulate two digits of precision, so this change takes it from a physical value of 0.00 to a physical of 0.01. It'd probably work best if it was < 100 like mental, since the game should put the victim in a mortally wounded state at that point.

inb4jex: But then wouldn't any instance of damage trigger the if statement? Because < 100 would also trigger after 1 box of damage right? Or is `if (FIGHTING(ch) == victim)` only true if the target is morted anyways?

jdevnull: `(int) (GET_PHYSICAL(victim)/100)` is the character's current number of physical boxes (that is, it starts at 1000 for an unwounded character with max 10 physical boxes and is reduced as they take damage).

`if (FIGHTING(ch) == victim)` checks to make sure that `victim` is `ch`'s current target.

inb4jex: Oh wait I completely misread what Lucien said. I think `GET_PHYSICAL(victim) < 100` would work then since it'd trigger when they go below 1 box of physical. Thanks for clearing stuff up!

luciensadi:
Looks good to me, thanks!

PR #741, opened 2024-05-02
Fix persona program storage space usage

jdevnull:
In `auto_repair_obj`, persona programs in store-bought decks were being added to `GET_CYBERDECK_USED_STORAGE`. Only utility programs and files that can be uploaded/downloaded should use storage memory (SR3 pg 207). This PR fixes this for store-bought decks (in custom decks, persona programs are handled as hardware, not software).

This check already exists for loading/unloading persona programs from store-bought decks.

### DIFF
--- a/src/newdb.cpp
+++ b/src/newdb.cpp
@@ -2433,7 +2433,8 @@ void auto_repair_obj(struct obj_data *obj, idnum_t owner) {
                 break;
             }
           }
-          if (GET_OBJ_TYPE(installed) == ITEM_PROGRAM) {
+          // Personas don't take up storage in store-bought decks (personas are not programs in custom decks)
+          if ((GET_OBJ_TYPE(installed) == ITEM_PROGRAM) && !((GET_OBJ_TYPE(obj) == ITEM_CYBERDECK) && (GET_PROGRAM_TYPE(installed) <= SOFT_SENSOR))) {
             GET_CYBERDECK_USED_STORAGE(obj) += GET_PROGRAM_SIZE(installed);
           }
         }


src/newdb.cpp
```diff
@@ -2433,7 +2433,8 @@ void auto_repair_obj(struct obj_data *obj, idnum_t owner) {
                 break;
             }
           }
-          if (GET_OBJ_TYPE(installed) == ITEM_PROGRAM) {
+          // Personas don't take up storage in store-bought decks (personas are not programs in custom decks)
+          if ((GET_OBJ_TYPE(installed) == ITEM_PROGRAM) && !((GET_OBJ_TYPE(obj) == ITEM_CYBERDECK) && (GET_PROGRAM_TYPE(installed) <= SOFT_SENSOR))) {
```
luciensadi, line 2437: How did this go when you tested it? If I'm reading it right, the logic is:
- If it's a program, and
- If it's neither installed in a store-bought deck nor a persona-type program
I think the intended logic is "if it's a program and either is in a custom deck or is a non-persona program"

jdevnull: If it's in a store-bought deck and is a persona-type program, then it should not take up storage memory. This is how it's written in `perform_put_cyberdeck` and `perform_get_from_container`, so I used the same pattern for consistency.

luciensadi: Hm, looks like I misread the logic, but the negated and is a bit rough to read in this case. If that's how it's written in the other files we can leave it be, and I'll consider coming back through to update logic for readability later.

luciensadi:
Thanks for the PR!

PR #743, opened 2024-06-08
Minor fixes

jdevnull:
1. Better messaging will let players know that excess cash-in-hand is the reason for increased docwagon retrieval costs. For consistency (and for better in-universe reasoning), cost of retrieval is paid from cash-in-hand first, and then from the bank only if cash-in-hand is insufficient.

2. Max physical/mental boxes and target mods should be reset _before_ various effects (like drugs) are applied. This solves the missing TN penalties when undergoing drug withdrawal: https://discord.com/channels/564618629467996170/788953927269613608/1228962851381575791

3. (reverted)

4. When dragging veh A into veh B, check the weight of veh A instead of veh B.

### DIFF
--- a/src/interpreter.cpp
+++ b/src/interpreter.cpp
@@ -624,7 +624,7 @@ struct command_info cmd_info[] =
     { "domain"     , POS_LYING   , do_domain   , 0, 0, BLOCKS_IDLE_REWARD },
     { "download"   , POS_RESTING , do_download_headware, 0, 0, BLOCKS_IDLE_REWARD },
     { "donate"     , POS_RESTING , do_drop     , 0, SCMD_DONATE, BLOCKS_IDLE_REWARD },
-    { "drag"       , POS_STANDING, do_drag     , 0, 0, BLOCKS_IDLE_REWARD },
+    { "drag"       , POS_SITTING , do_drag     , 0, 0, BLOCKS_IDLE_REWARD },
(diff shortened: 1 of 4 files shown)


src/interpreter.cpp
```diff
@@ -626,7 +626,7 @@ struct command_info cmd_info[] =
     { "domain"     , POS_LYING   , do_domain   , 0, 0, BLOCKS_IDLE_REWARD },
     { "download"   , POS_RESTING , do_download_headware, 0, 0, BLOCKS_IDLE_REWARD },
     { "donate"     , POS_RESTING , do_drop     , 0, SCMD_DONATE, BLOCKS_IDLE_REWARD },
-    { "drag"       , POS_STANDING, do_drag     , 0, 0, BLOCKS_IDLE_REWARD },
+    { "drag"       , POS_SITTING , do_drag     , 0, 0, BLOCKS_IDLE_REWARD },
```
luciensadi, line 629: Just to confirm, this worked when you tested it? Making a command suddenly work in vehicles is a leading cause of crashes due to unguarded references of ch->in_room.

jdevnull: Ah, a bit more complicated than I thought. Might actually be better to put the functionality into do_pull instead. Anyway, reverted this bit.

luciensadi:
LGTM, thanks!

PR #744, opened 2024-06-14
Hermes tweaks

jdevnull:
Specify that Hermes is looking for _highly skilled_ awakened. This solves the edge case where a lower skilled awakened may be directed here but isn't able to figure out why they can't pick up the quest.

Also, allow conjuring skill as an alternative to sorcery/aura reading (e.g., for summoning focused or aspected characters).

This addresses:
https://discord.com/channels/564618629467996170/788953927269613608/1251271414405664808
https://discord.com/channels/564618629467996170/788953927269613608/1251271642458492928


luciensadi:
LGTM, thanks!

PR #745, opened 2024-06-18
Fix for hedit 'x' crash, v.2

jdevnull:
1. When host changes are saved, we call `collate_host_entrances` to rebuild entrance lists. Thus, when deleting host A, we don't need to search other hosts for entrances pointing to host A. This solves the issue of trying to find an entrance in another host that may not (yet) exist.

2. It makes sense to clear host A's own list of entrances as part of `free_host` instead of in `hedit`.

3. When making a copy of an existing host for editing in `olc`, we don't want the copy to point to the existing host's list of entrances. This solves the issue where we can end up with a pointer directing us to freed memory. We also don't need to make a copy of the list because `collate_host_entrances` will rebuild the list if/when we save, so we can just set it to null.

### DIFF
--- a/src/hedit.cpp
+++ b/src/hedit.cpp
@@ -301,7 +301,10 @@ void hedit_parse(struct descriptor_data *d, const char *arg)
         send_to_char("Writing host to disk.\r\n", d->character);
         write_host_to_disk(d->character->player_specials->saved.zonenum);
         send_to_char("Saved.\r\n", CH);
-        Mem->DeleteHost(d->edit_host);
+
+        // Use of clear_host here instead of DeleteHost lets us keep the same linked lists
+        clear_host(d->edit_host);
+        delete d->edit_host;
         d->edit_host = NULL;
         PLR_FLAGS(d->character).RemoveBit(PLR_EDITING);
         STATE(d) = CON_PLAYING;
--- a/src/hedit.cpp
+++ b/src/hedit.cpp
@@ -213,12 +213,6 @@ void hedit_parse(struct descriptor_data *d, const char *arg)
     case 'N':
       STATE(d) = CON_PLAYING;
       if (d->edit_host) {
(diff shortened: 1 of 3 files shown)


src/hedit.cpp
```diff
@@ -307,8 +301,7 @@ void hedit_parse(struct descriptor_data *d, const char *arg)
         send_to_char("Writing host to disk.\r\n", d->character);
         write_host_to_disk(d->character->player_specials->saved.zonenum);
         send_to_char("Saved.\r\n", CH);
-        clear_host(d->edit_host);
-        delete d->edit_host;
+        Mem->DeleteHost(d->edit_host);
```
luciensadi, line 304: What's the difference between clear_host() and DeleteHost? Are we sure that all strings, trigger steps, etc have been copied and that deletion of d->edit_host and all pointed-to memory won't cause issues?

luciensadi:
Looks like the crashes were caused by the `entrance` linked list not being cloned on hedit. This caused the actual host's entrances to be blown away when the edited host was extracted, which broke memory constraints.

src/hedit.cpp
jdevnull: Ahh, I see. So the choice of clear_host was deliberately made instead of an oversight? I was thinking using clear_host would have been a source of memory leaks since things were being left behind.

jdevnull:
Yeah, this was what I was trying to describe in #3 of the PR description. That said, we don't actually need to clone the entrance list, since they're going to be rebuilt anyway - we just need the copy to not point to the actual host's linked list, so we can just assign it to null.

jdevnull:
I don't believe this is needed, so closing the PR.

luciensadi:
I tried a fix on my own while we were talking about it, but I wasn't sure if the change I'd made had actually stabilized things, so I'd left the PR open just in case. Appreciate you submitting this!

PR #747, opened 2024-08-25
Metavariant social penalty bits (add newbie tip, no penalties w/ vending machines)

jdevnull:
Added checks so that metavariant penalties aren't applied when a character is interacting with some kind of inanimate, since vending machines shouldn't care what you look like. Other inanimates (robots, etc) _probably_ also shouldn't care.

### DIFF
--- a/src/newshop.cpp
+++ b/src/newshop.cpp
@@ -1841,7 +1841,7 @@ void shop_list(char *arg, struct char_data *ch, struct char_data *keeper, vnum_t
   if (has_availtns) {
     send_to_char(ch, "This shop uses %s for difficult purchases.\r\n", skills[shop_table[shop_nr].etiquette].name);
     if (SHOULD_SEE_TIPS(ch) && get_metavariant_penalty(ch, keeper))
-      send_to_char(ch, "NOTE: %s ^w%s^n will suffer penalties to etiquette/negotiation tests here.\r\n",
+      send_to_char(ch, "NOTE: %s ^W%s^n will suffer penalties to etiquette/negotiation tests here.\r\n",
                    AN(pc_race_types_decap[(int)GET_RACE(ch)]), pc_race_types_decap[(int)GET_RACE(ch)]);
   }
 }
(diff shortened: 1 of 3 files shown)


src/newshop.cpp
```diff
@@ -1838,8 +1838,11 @@ void shop_list(char *arg, struct char_data *ch, struct char_data *keeper, vnum_t
   if (SHOULD_SEE_TIPS(ch))
     send_to_char("\r\nUse ^WPROBE^n for more details.\r\n", ch);
 
-  if (has_availtns)
+  if (has_availtns) {
     send_to_char(ch, "This shop uses %s for difficult purchases.\r\n", skills[shop_table[shop_nr].etiquette].name);
+    if (SHOULD_SEE_TIPS(ch) && get_metavariant_penalty)
+      send_to_char("\r\n^wMetavariants^n may suffer etiquette/negotiation penalties.\r\n", ch);
```
luciensadi, line 1844: Good idea! Let's change this to something like `<character's race pluralized> have a -4 TN penalty here.`

jdevnull: I'm not sure about the singular/plural form of fomori/oni, so I'm now using:
> NOTE: <a/an> <race> will suffer penalties to etiquette/negotiation tests here.

jdevnull:
Moved the same-race and inanimate target checks into `get_metavariant_penalty(ch, tch)` instead of doing the same code in 3 places. I'm having a little difficulty with mysql at the moment, so I can't really test it on my machine, but I think it should be good.

src/newshop.cpp
luciensadi: IMO they're like sheep, so their singular and plural are the same. That's also written in the helpfiles, like "Oni display" and "Fomori possess", so I think it's a reasonable take on it.

I can write that change if you want to be done with this PR!

jdevnull: Yeah, probably better if you take over, since I'm also waffling between making another array like `pc_race_types_decap` for the plural forms, or using a define like `AN`.

luciensadi: I'll just make an array for it.

luciensadi:
Thanks for the PR!

PR #749, opened 2024-08-27
On NPC death, leave up to one round in their firearm

jdevnull:
When ammo dropped in NPC firearms, players could learn what type of ammo was being fired at them with an in-character mechanism, helping them understand why some enemies are more dangerous than others. Allowing up to one round to remain in dropped NPC firearms restores this mechanism.

Technically, this restores some ability to farm ammo, but at no more than one round per firearm, it would be... inefficient.

### DIFF
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -686,7 +686,7 @@ void make_corpse(struct char_data * ch)
       else {
         // Zero out magazines from this and everything it contains.
         // Leave a round in NPC firearms so characters get some indication of what ammo was being used against them
-        zero_out_magazine_counts(o, IS_NPC(ch));
+        zero_out_magazine_counts(o, IS_NPC(ch) ? 1 : 0);
 
         // Put the item in the corpse.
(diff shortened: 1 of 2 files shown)


src/fight.cpp
```diff
@@ -685,7 +685,8 @@ void make_corpse(struct char_data * ch)
         extract_obj(o);
       else {
         // Zero out magazines from this and everything it contains.
-        zero_out_magazine_counts(o);
+        // Leave a round in NPC firearms so characters get some indication of what ammo was being used against them
+        zero_out_magazine_counts(o, IS_NPC(ch));
```
luciensadi, line 689: This works but isn't the easiest to read. Recommend `IS_NPC(ch) ? 1 : 0` so it's clear that the bool is being coerced to an int in the process.

luciensadi (changes requested):
Makes sense to me. Requesting one change for maintainability down the line.

luciensadi:
Thanks for the PR!

PR #754, opened 2024-09-29
Toxin extractors apply to drug onset damage as well as comedown

jdevnull:
Previously, toxin extractors reduced the power of drug comedown damage. Man & Machine, pg 71, doesn't appear to have an exception for drug onset damage, so this PR applies toxin extractors to onset damage tests.


luciensadi:
LGTM, thanks!

PR #756, opened 2024-10-13
Fix adept distance strike & penetrating strike interactions

jdevnull:
Previously, having both penetrating strike and distance strike active at the same time meant that neither would work. Also, an adept using distance strike would still get scorched by flame aura, despite being up to (magic) meters away from their target. This PR does the following:

1. If both powers are active, default to distance strike (that is, only disable penetrating strike)
2. Distance strike skips flame aura (similar to as if the adept was using a weapon).
3. For that matter, let's block power activation if an incompatible power is already active.
4. Added a comment on why we don't prevent activation of adept improved reflexes when other tech/magic reaction/init is active (we can interpret it as "does not stack" and so the character gets whichever bonus is higher, which is how it's handled in `handler.cpp: affect_total`.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -31,7 +31,7 @@ BOOST_FLAGS = -lboost_filesystem
 
 # You probably want to comment this out: This is the makefile specification for the GitHub runner.
 LIBS = -lcrypt -L/usr/local/lib $(COMMON_FLAGS) $(BOOST_FLAGS) -lnsl -lcurl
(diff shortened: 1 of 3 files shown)


jdevnull:
Hmm. CodeQL doesn't like part of the previously existing code that I didn't touch.
```utils.cpp:3818:8: error: 'strlcpy' is missing exception specification 'throw()' ```

luciensadi (approved):
I'll go ahead and merge this, then debug the CodeQL issue on my end.

PR #759, opened 2024-10-16
Implement VCR + reflex trigger matrix rules

jdevnull:
Previously, the only part of the rigger VCR matrix penalties (Matrix, pg 28) that had been implemented was subtracting the VCR's rating from the hacking pool. This PR implements the rest of the section: the +1 TN penalty (not dependent on VCR rating) and the use of a reflex trigger to disable their VCR and thus negate the penalties. For convenience, let's assume that if a rigger has a trigger, it applies to their VCR. Furthermore, let's assume that if a rigger has a trigger, that they will always disable their VCR when connecting to the matrix and enable their VCR after disconnecting.

This PR does not actually implement the ability to disable VCR bonuses to rigging, since there's no non-nerp reason to be able to do so.

Also, moved the guard to ensure the hacking pool can't fall below 0 to after all the pool contributions are calculated. This fixes an edge case where a rigger with less bonus hacking pool dice from 'ware (encephalon, math spu) than their VCR rating could end up with a higher hacking pool than expected.

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -955,10 +955,10 @@ void affect_total(struct char_data * ch)
     // also adds to initiative dice
     rigger_init_dice += has_rig;
 
-    // but reduces hacking pool
-    GET_HACKING(ch) -= has_rig;
-    if (GET_HACKING(ch) < 0)
-      GET_HACKING(ch) = 0;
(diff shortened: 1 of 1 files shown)


luciensadi (approved):
LGTM, thanks for the PR!

PR #763, opened 2024-11-27
Platelet factories & crash tests

jdevnull:
1. Synthacardium adds its rating in dice to resist cardiac and circulatory things (M&M pg 70), such as the platelet factories' risk of cardiac arrest when not taking anticoagulants for too long (M&M pg 69). This addresses https://discord.com/channels/564618629467996170/788953927269613608/1311143404419285105

2. DocWagon now gives aspirin to patients with platelet factories. This addresses https://discord.com/channels/564618629467996170/788953927269613608/1317680664346103829

3. Losing consciousness now stops a driver's vehicle (with a crash test using autonav dice) if it was moving. This occurs when damage causes POS_MORTALLYW, POS_STUNNED, or POS_DEAD, and is also checked in `raw_kill` and `docwagon_retrieve`. This addresses https://discord.com/channels/564618629467996170/788953927269613608/1317679405568364545

4. Crash tests now properly clears flags / stops the vehicle if the driver was ejected or the vehicle was destroyed.

### DIFF
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -96,7 +96,7 @@ extern struct zone_data *zone_table;
 extern void perform_tell(struct char_data *, struct char_data *, char *);
 extern int can_wield_both(struct char_data *, struct obj_data *, struct obj_data *);
 extern void find_and_draw_weapon(struct char_data *);
-extern void crash_test(struct char_data *ch, bool force_zero_successes);
+extern void crash_test(struct char_data *ch, bool no_driver);
 extern int get_vehicle_modifier(struct veh_data *veh, bool include_weather=TRUE);
 extern bool mob_magic(struct char_data *ch);
 extern void cast_spell(struct char_data *ch, int spell, int sub, int force, char *arg);
--- a/src/fight.cpp
+++ b/src/fight.cpp
@@ -106,6 +106,7 @@ extern bool check_sentinel_snap_back(struct char_data *ch);
 extern void end_quest(struct char_data *ch, bool succeeded);
 extern bool npc_vs_vehicle_blocked_by_quest_protection(idnum_t quest_id, struct veh_data *veh);
 extern bool ch_is_in_viewers_visual_range(struct char_data *ch, struct char_data *viewer);
+extern void stop_driving(struct char_data *ch, bool is_involuntary);
 
 // Corpse saving externs.
 extern bool Storage_get_filename(vnum_t vnum, char *filename, int filename_size);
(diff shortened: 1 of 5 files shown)


src/fight.cpp
```diff
@@ -2353,6 +2358,18 @@ void docwagon_retrieve(struct char_data *ch) {
   if (CH_IN_COMBAT(ch))
     stop_fighting(ch);
 
+  // Stop their vehicle
```
luciensadi, line 2361: This should be a check for AFF_PILOT or AFF_RIG. PLR_REMOTE is only used when you're sitting in a room and controlling a vehicle from afar.

luciensadi: Well, one line down, but same comment. Otherwise, this looks good to me!

jdevnull: I think when AFF_RIG is set, AFF_PILOT is also set, so we don't need to check both.

While I didn't think through the PLR_REMOTE check (I copy-pasted from raw_kill), I'm thinking that if a rigger sitting in a room gets picked up by docwagon, any drone they're controlling at the time should also stop (though it probably doesn't matter since they can still regain control remotely even if the drone was left still moving).

Thinking about it, maybe this should cause a crash test instead of just SPEED_OFF, rolling the autonav alone to check if the vehicle safely stops after the sudden loss of a driver. Not all that sure about this though... drones with 0 autonav and vehicles with higher handling ratings would have good odds of serious damage.

luciensadi: As an option, I've added `is_involuntary` to `stop_driving(struct char_data *, bool is_involuntary)`, which will force a crash test at 0 skill. I'd recommend using stop_driving to get someone from behind the wheel since it's designed around the edge cases etc already.

jdevnull: Nice. Got rid of the vehicle stopping code in both raw_kill and docwagon_retrieve in favor of calling stop_driving in raw_damage if stunned, morted, or dead.

I don't think is_involuntary should force zero successes on the crash test, since the autonav dice should still apply. Also not sure the is_involuntary code works as intended... I think it should apply to riggers dropping unconscious as well, and crash_test might not actually do anything since it happens after stopping the vehicle.

luciensadi: > Nice. Got rid of the vehicle stopping code in both raw_kill and docwagon_retrieve in favor of calling stop_driving in raw_damage if stunned, morted, or dead.

Was there a specific behavior you were trying to fix here? I'm concerned that this will mess with things like the staff kill command, and more generally concerned with changing things that aren't broken since doing so introduces the possibility of bugs.

> I don't think is_involuntary should force zero successes on the crash test, since the autonav dice should still apply.

Updating the success test to roll dice equal to the autonav rating instead could make sense.

> Also not sure the is_involuntary code works as intended... I think it should apply to riggers dropping unconscious as well, and crash_test might not actually do anything since it happens after stopping the vehicle.

I'm fine with both of these changes, I just threw something together for the change. I was planning on testing it but got caught up in work :(

jdevnull: > Was there a specific behavior you were trying to fix here? I'm concerned that this will mess with things like the staff kill command, and more generally concerned with changing things that aren't broken since doing so introduces the possibility of bugs.

The original reported problem was that a docwagon pick-up of a driver (that was morted by a platelet factories heart attack) would leave the vehicle moving, so they couldn't get back into the vehicle on returning (there was already code in place dealing with the case of a driver completely dying in raw_kill, which is what I copy-pasted into docwagon_retrieve). Moving this to the POS_MORTALLYW from damage step would help with the hypothetical situation where a player doc is trying to respond before docwagon pickup or death (the POS_STUNNED and POS_DEAD parts are just to keep things logically consistent).

luciensadi: Got it, let's call it in both places then. There's no harm in the overlap (it's got guards for being called on passive passengers), and that will help with the raw_kill edge cases.

jdevnull: Done

jdevnull:
Changes to crash_test:
- removed "You don't have the skills to be driving like this!" message when you succeed
- fixed edge case where dying when being ejected from a bike/motorboat doesn't clear driving/rigging flags
- stop vehicle if it's destroyed

luciensadi:
Appreciate the code! I'll get this loaded onto the buildport so we can test it.

PR #764, opened 2024-12-09
Major Feature: Otaku Prestige Class

biscuitWizard:
# Otaku
Added Otaku per Matrix 3 to AwakeMUD, CE. 

## Forms
Complex Forms have been added:
```
Syntax:    forms               -> Lists all memorized complex forms
           forms list          -> As above
           forms forget <form> -> Forgets a learned form
           forms learn <form>  -> Practices a form

Forms is the command used to manage and develop otaku Complex Forms. Using the 
'create complex form' syntax it is possible to create a complex form that mimics a utility. 
The rating and program type is decided upon using the 'create' syntax.

After a complex form is created, it must be learned to be accessible while in the matrix. 
Use 'forms learn <form name>' to finish learning the form. Forms cannot be upgraded,
so creating new forms is the only way to upgrade forms.
```

## Grade
Submersion grade is like initiation for mages. Paying the karma and associated nuyen costs for submersion allows an otaku to learn echoes. Echoes offer mechanical benefits that make an otaku more potent.

The following echoes are implemented:
- [x] Improved I/O Speed - +100mp to living deck's I/O speed
- [x] Improved Hardening - +1 to living deck's hardening
- [x] Improved MPCP - +1 to living deck's MPCP score
- [x] Improved Persona Bod - +1 to living persona's Bod
- [x] Improved Persona Evasion - +1 to living persona's Evasion
- [x] Improved Persona Masking - +1 to living persona's masking
- [x] Improved Persona Sensor - +1 to living persona's sensor
- [x] Improved Reaction - +1 to living deck's reaction
- [x] Ghosting - Adds +1 to otaku's Detection Factor
- [x] Neurofilter - -2 TN to WIL test when jacking out; bonuses to absorbing IC damage
- [x] Overclock - Adds 1 dice to otaku's Initiative

## Living Decks
Living Persona/Decks have implemented. An otaku during chargen is automatically given a datajack and an ASIST converter. Otakus can connect to the matrix directly without a deck. Their persona and deck is created 'virtually' and they're able to use those living entities to explore the matrix.

Persona damage goes straight to the otaku's stun track. Complex forms are loaded as programs. And the Deck's attributes are derived directly from the otaku's score.

When executing the command ``connect with cyberdeck`` or ``connect cyberdeck`` while possessing a cyberdeck, otaku can use their 'proxy cyberdeck' to store files and paydata while decking, while using their living persona.

## Houserules:
- [x] Damage to an otaku's MPCP deals physical damage to the otaku, regardless of neurofilter.
- [x] MPCP and personas are capped at 12
- [x] Otaku do not get a medic program (no use) 
- [x] 'connect with cyberdeck' allows otaku to use a cyberdeck as a file storage medium when downloading files 
- [x] Since bod/will ICCM rolls are standard here, neurofilter gives a -2 to TN on WIL saves for IC damage 

### Design Notes
Per Khai:
```
Proposed houserules:
No sprites. No tribes. Magic doesn't affect the living persona. MPCP and personas are capped at 12. Persona attributes don't face permanent damage (just log out and back in to recover). Damage to an otaku's MPCP deals physical damage to the otaku, regardless of neurofilter.

Otaku details:
Computer and channel skills follow skill limits and so max out at 12. Hardening is capped by MPCP, and complex forms are capped by both mpcp and computer skill. Those on the cyberadept path get +1 to the effective rating when using complex forms. Those on the technoshaman path get -1 TN when using their channels (i.e., system tests). Otaku also get an extra +1 to detection factor (DF), so taking rounding into account, the cyberadept (12 masking, 12+1 sleaze complex form, round up) can reach DF 14. Total persona ratings are not capped to 3x MPCP, so they don't dump evasion like most deckers do.

Otaku vs IC:
Where IC would deal damage to a cyberdeck's condition monitor, it instead deals damage to the otaku's stun track (which I understand to mean that the medic program has no effect).

Khai's commentary, part 1:
The damage to the otaku's stun track from standard damage is maybe a bigger deal than it initially appears. Where a decker would just run medic once or twice and then move on, an otaku may need to log out entirely to recover. That is to say, taking any hits at all will substantially increase the difficulty of some runs (spoiler: there's a GDQ run where you need to get through 9 consecutive red/black hosts, or more if you don't know exactly where you're going). The higher DF or lower system test TNs do help to reduce the amount of IC that need to be faced, but I'm not sure they'll have a large enough effect to be considered balanced. The houserule to apply physical damage from damage to the otaku's MPCP would have a similar effect.

Not dumping evasion does allow the otaku to use parry and evade detection maneuvers, but the way they're implemented (you need to be in combat with the IC before you can attempt a maneuver) means that the IC may often get an attack in before the otaku can maneuver.

Host Maximums:
From SR3 pg 205, hosts have max security rating = 12 and max subsystem ratings = 18. When a passive alert is triggered, all subsystem ratings are raised by 2.

Success rates vs host maximums:
With a maximum subsystem rating of 20, a decker with r12 programs roll system tests against a max TN of 8, whereas a technoshaman can reduce it to TN 7. If we assume mpcp 12, int 8, encephalon 2, and math spu 3, that's a maximum hacking pool of 11 with a default allocation of 3. So, we can expect top end deckers (and otaku) to roll 12 computer skill + 3 int task pool + 3 hacking pool =  18 dice. The host resists by rolling their security rating against the decker's DF.

  Decker: 86% success rate, +0.33 tally per system test
  Cyberadept: 87% success rate, +0.28 tally per system test
  Technoshaman: 91% success rate, +0.33 tally per system test

Khai's commentary, part 2:
Not surprisingly, the otaku do better on individual tests than the decker, but are they enough better to offset the stun boxes from standard IC attacks?

Physical combat:
Otaku get -1 to physical attributes and +1 to mental attributes. The net result compared to an equivalent decker is -1 soak, -0.5 max worn armor, -1 melee power, +0.5 cpool.

Chargen:
Per Matrix pg 136-137, otaku always start with 5k nuyen in resources, so they can't start with cerebral boosters, encephalon, or math spu (note: mud implementation allows int task pool for matrix tests, so a cold asist decker does get 3 dice out of these). They can start with computer 8 and are able to use hacking pool dice immediately, so they're a bit ahead on dice for matrix tests. However, they can only have one channel at rating 6, one at 5, and the rest at 4. So compared to an array of r6 utility programs, a technoshaman is net even on system test TNs (of the three relevant categories, better in one and worse in one), and a cyberadept is mostly worse.

In terms of mpcp + hardening (to avoid mpcp damage), the wakyambi (8 mpcp, 3 hardening) and gnome (7 mpcp, 4 hardening) are comparable to a CMT Avatar, but any other metavariant is worse.

In terms of DF, otaku (assuming maxed out charisma and willpower) tend to be equivalent to the decker (note: there's no difference between DF 6 and 7), though cyberadepts that are elf metavariants or gnome can be a point better (DF 8).

Khai's commentary, part 3:
I think it's reasonable to claim that otaku without sprites are not better than mundane deckers in either the early or late game. It's possible that they are noticeably worse depending on how much of an impact their weakness to standard IC attacks may actually have. In addition, it takes a great deal more time and effort for an otaku to progress: compare 250 TKE and 800k nuyen for a decker to complete an mpcp 12 deck + r12 programs, against 3400 TKE and 14.3m nuyen for a living persona to match. I would not be inclined towards making any further nerfs (beyond the proposed house rules at the top) at this time.

I support making this a prestige role. As Lucien described, "not better, just different and usually harder".

```

Khai:
```
I would say that an intelligent sprite implementation (where you could order them to intercept IC instead of risking taking wound penalties that might prevent you from going any deeper into a network) would solve the otaku's biggest weakness and allow the otaku's strengths to be more meaningful.
```


biscuitWizard:
todo list:
- Limit complex form rank by stats
- mpcp damage overrides
- gray ic damage overrides
- submersion & echos
- otaku channel trainer

biscuitWizard:
Progress update: All of the echoes are implemented and have mechanics. Submersion needs testing but should work exactly like initiation.

Complex forms should be designable but I forgot to declare the command so it's not possible to actually learn them, though that should just be 1 line of code. And then a whole lot of bugs. As I'm pretty sure I used the design ticks vars instead of the design prog ticks vars. Probably lots of bugs.

Connecting to the matrix should work and have all the right MPCP scores. Channels should work on any system test. Tarbabies can't affect channels as per the book. Complex forms are loaded as programs onto the deck. So assuming you can make a complex form, the form should work as appropriate.

The channel skills have all been added. Need an NPC and room / quest gate to learn the actual skills. The limits for chargen channels needs to be added (6, 5, 4, 3, 3)

Chargen priority system is working with otaku. Resources are always capped at 5000. The UI has a bug where it says you get 20000 but you only get 5000. I probably missed a ternary somewhere.

Chargen prestige system works with otaku, it's a toggle under the races menu. It asks for a prestige character as it should. all works. SQL migrations done for adding otaku to the game. You can save/quit/load correctly.

Chargen points system does not work at all with otaku. TODO.

Remaining major action items:
MPCP damage mechanics
Gray IC stun track mechanics
Learning channel skills
Review and debug complex forms

biscuitWizard:
'forms' is added.

It's now possible to list all your complex forms. You can also 'forms forget <form>' to forget a complex form, and 'forms learn <form>' to B/R a form. Complex forms now appropriately show up as programs in the matrix. Complex forms are essentially complete.

Submersion appears to work as expected.

MPCP damage has been implemented but likely needs some additional balancing.

I think I'd like to make it so when using 'connect' it checks if you have a deck, and asks if you're connecting through your deck, your persona, or your persona through your deck so that it can use your deck as a 'storage medium' for paydirt.

src/utils.cpp
```diff
@@ -2087,6 +2087,22 @@ int get_num_of_cyber_replacements(struct char_data *ch) {
 
 int get_attr_max(struct char_data *ch, int attr) {
   int racial_max = racial_limits[(int) GET_RACE(ch)][RACIAL_LIMITS_NORMAL][attr];
+
+  if (IS_OTAKU(ch)) {
```
luciensadi, line 2091: Can you add a comment with a page number here just for future reference?

src/utils.cpp
```diff
@@ -5819,6 +5835,65 @@ struct obj_data *make_new_finished_program(int part_type, int mpcp, int rating=0
   return prog;
 }
 
+struct obj_data *make_otaku_deck(struct char_data *ch) {
+  struct obj_data *asist = find_cyberware(ch, CYB_ASIST);
+  struct obj_data *new_deck = read_object(OBJ_CUSTOM_CYBERDECK_SHELL, VIRTUAL, OBJ_LOAD_REASON_OTAKU_BS);
+
```
luciensadi, line 5841: Please add safeguard checks for `if (!ch || !asist | !new_deck) { log something, return }` here. I never think I'll hit them when coding, but every time one of those safeguard log messages pop, I know they prevented a crash.

luciensadi (commented):
Partial review, more to come as I have time this week/weekend.

src/utils.cpp
luciensadi: You can generally assume `read_object()` will never fail, so you can omit that if you want.

src/utils.cpp
```diff
@@ -5819,6 +5835,65 @@ struct obj_data *make_new_finished_program(int part_type, int mpcp, int rating=0
...
+  obj_to_obj(make_new_finished_part(PART_BOD, mpcp, GET_REAL_WIL(ch)), new_deck);
+  obj_to_obj(make_new_finished_part(PART_EVASION, mpcp, GET_REAL_INT(ch)), new_deck);
+  obj_to_obj(make_new_finished_part(PART_SENSOR, mpcp, GET_REAL_INT(ch)), new_deck);
+  obj_to_obj(make_new_finished_part(PART_MASKING, mpcp, (GET_REAL_WIL(ch) + GET_REAL_CHA(ch) + 1) / 2), new_deck);
+  obj_to_obj(make_new_finished_part(PART_ASIST_HOT, mpcp), new_deck);
+  obj_to_obj(make_new_finished_part(PART_RAS_OVERRIDE, mpcp), new_deck);
+
+  GET_CYBERDECK_MPCP(new_deck) = MIN(12, mpcp);
+  GET_CYBERDECK_HARDENING(new_deck) = MIN(GET_REAL_WIL(ch), GET_REAL_WIL(ch) / 2 + GET_ECHO(ch, ECHO_IMPROVED_HARD));
+  GET_CYBERDECK_ACTIVE_MEMORY(new_deck) = 999999;
```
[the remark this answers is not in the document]
biscuitWizard: It's not used. Otaku don't have active memory. I'll add comments and so on.

src/spec_procs.cpp
```diff
@@ -1052,7 +1078,21 @@ SPECIAL(teacher)
...
   }
-  if (GET_SKILL_POINTS(ch) > 0)
+  if (PLR_FLAGGED(ch, PLR_NEWBIE) && skills[skill_num].requires_resonance && channel_points > 0) {
+    // Check this wouldn't put us above our newbie limits.
+    if (REAL_SKILL(ch, skill_num) + 1 > 3) {
+      // Check we don't have any other channel skills at the same value.
+      for (int ci=154; i < 159;i++) {
+        if (REAL_SKILL(ch, ci) == REAL_SKILL(ch, skill_num) + 1) {
+          // Break.
+          send_to_char(ch, "When buying channel skills with freebies you can only have one channel skill at %d, and %s is already at that value.",
```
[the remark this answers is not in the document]
biscuitWizard: Should only be called in chargen

src/spec_procs.cpp
```diff
@@ -908,8 +909,16 @@ SPECIAL(teacher)
     return TRUE;
   }
 
+  if (IS_OTAKU(ch)) {
+    // Otaku get free skill points to spend on JUST channels based on the avg of these stats, round up
```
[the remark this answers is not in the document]
biscuitWizard: Fixed!

src/spec_procs.cpp
```diff
@@ -1052,10 +1071,34 @@ SPECIAL(teacher)
...
+            send_to_char(ch, "When buying channel skills with freebies you can only have one channel skill at %d, and %s is already at that value.",
+              REAL_SKILL(ch, skill_num) + 1, skills[ci].name);
+              return FALSE;
+          }
+        }
+        GET_CHANNEL_POINTS(ch)--;
+      }
+    } 
+    else
+      GET_KARMA(ch) -= get_skill_price(ch, skill_num) * 100;
```
luciensadi, line 1095: Embedding this else-statement here means folks can't use skill points on channels. Recommend changing the overall logic to:
```c++
if (skills[skill_num].requires_resonance && GET_CHANNEL_POINTS(ch) > 0) {
  //pay with chanp
} else if (GET_SKILL_POINTS(ch) > 0) {
  //pay with skillp
} else {
  //pay with karma
}

File: src/spec_procs.cpp [hunk 7]
```diff
@@ -1153,6 +1196,10 @@ void train_attribute(struct char_data *ch, struct char_data *trainer, int ind, i
   // Apply the change.
   GET_REAL_ATT(ch, attr) += 1;
 
+  // Update channel points if in chargen
+  if (trainers[ind].is_newbie && IS_OTAKU(ch))
```
luciensadi, line 1200: You're probably going to have to change the logic so that you can't train/untrain at the newbie stat trainer if you have any channel skills set. Otherwise, someone could go spend all their points, then walk back to the trainer, UNTRAIN INT, TRAIN INT, and have max points again.

src/spec_procs.cpp
```diff
@@ -5545,7 +5592,10 @@ SPECIAL(chargen_unpractice_skill)
     }
 
     // Success. Lower the skill by one point.
-    GET_SKILL_POINTS(ch)++;
+    if (skills[skill_num].requires_resonance)
+      GET_CHANNEL_POINTS(ch)++;
```
luciensadi, line 5596: Logically, you probably want to only increase their channel points up to their attribute max, then put any extra points back into the general skill points pool. Otherwise you could have a situation where someone spends chanp + skillp on channels, then tries to untrain to pick up a different skill only to be locked into channeling and forced to remake.

src/chargen.cpp
```diff
@@ -825,6 +844,11 @@ int parse_prestige_race(struct descriptor_data *d, const char *arg)
   case 'd':
   case 'D':
     return RACE_FEATHERED_SERPENT;
+  case 'o':
+  case 'O':
+    d->ccr.is_otaku = !d->ccr.is_otaku;
```
luciensadi, line 849: Are any races forbidden from being otaku (e.g. dragons, because they cannot have cyberware implanted)? If so, you'll want to check here for that race already being set, and check in the race toggle and refuse to proceed if otaku is already enabled.

src/chargen.cpp
```diff
@@ -957,7 +988,7 @@ int parse_assign(struct descriptor_data *d, const char *arg)
     case PR_MAGIC:
       break;
     case PR_RESOURCE:
-      GET_NUYEN_RAW(d->character) = nuyen_vals[d->ccr.temp];
+      GET_NUYEN_RAW(d->character) = MIN(nuyen_vals[d->ccr.temp], d->ccr.is_otaku ? 5000 : nuyen_vals[0]);
```
luciensadi, line 991: What's the reason for / effect of this line? It seems to cap your nuyen to 5k if you're an otaku no matter what. If Otaku can never start with more than 5k, please disable setting the nuyen resources priority for otaku on any but the lowest slot to avoid creating a newbie trap.

src/chargen.cpp
```diff
@@ -1084,6 +1118,37 @@ static void start_game(descriptor_data *d, const char *origin)
...
+  // Otaku get some starting gear, so assign that here
+  if (d->ccr.is_otaku) {
+    // Equip cyberware (deduct essence and modify stats as appropriate)
+    obj_data *temp_obj;
+    for (int cyb = 0; cyb < OTAKU_CYBERWARE_NUM; cyb++) {
+      if (otaku_cyberware[cyb]) {
+        if (!(temp_obj = read_object(otaku_cyberware[cyb], VIRTUAL, OBJ_LOAD_REASON_ARCHETYPE))) {
+          snprintf(buf, sizeof(buf), "SYSERR: Invalid cyberware item %ld specified for otaku class.",
+                  otaku_cyberware[cyb]);
+          mudlog(buf, CH, LOG_SYSLOG, TRUE);
```
luciensadi, line 1130: FYI, this is the old style of logging. You don't need to change this, but we now would skip pre-composing the buf and just use:

`mudlog_vfprintf(CH, LOG_SYSLOG, "fmt", vars);`

src/chargen.cpp
```diff
@@ -2037,6 +2131,33 @@ void create_parse(struct descriptor_data *d, const char *arg)
       priority_menu(d);
     break;
 
+  case CCR_OTAKU_PATH:
+  #define PATH_HELP_STRING "\r\nCyberadepts express their connection to the matrix in a formulaic and rational way. Their complex forms are more potent." \
```
luciensadi, line 2135: What is the gameplay impact of more potent forms vs easier channel/command? Are cyberadepts better at matrix combat than technoshamans or something?

src/chargen.cpp
```diff
@@ -2037,6 +2131,33 @@ void create_parse(struct descriptor_data *d, const char *arg)
...
     break;
 
+  case CCR_OTAKU_PATH:
+  #define PATH_HELP_STRING "\r\nCyberadepts express their connection to the matrix in a formulaic and rational way. Their complex forms are more potent." \
+            "\r\n" \
+            "\r\nTechnoshamans commune with the matrix in a more magical way. The natural way in which they interact with the matrix makes it easier to channel and command." \
+            "\r\n" \
+            "\r\nEnter C to select Cyberadept, or T to select Technoshaman: "
+    if (isalpha(*arg) && isalpha(*(arg+1))) {
+      snprintf(buf, sizeof(buf), "\r\nARG: '%s'\r\n", arg);
```
luciensadi, line 2141: Leftover debug code?

src/chargen.cpp
biscuitWizard: Correct, otaku can never start with more than 5k per the book. But it has to be settable as higher otherwise there aren't enough options to spend all priority points / etc.

src/chargen.cpp
biscuitWizard: ACIFS is a technoshaman's specialty, and operational utilities are a cyberadept's specialty.

src/chargen.cpp
biscuitWizard: For the record I also made this not a newbie trap in that no matter what you select it will show 5k in the interface.

src/newdb.cpp
```diff
@@ -563,7 +563,7 @@ bool load_char(const char *name, char_data *ch, bool logon, int pc_load_origin)
       GET_ARCHETYPAL_MODE(ch) = (bool) atoi(row[5]);
       GET_ARCHETYPAL_TYPE(ch) = atoi(row[6]);
       GET_PRESTIGE_ALT_ID(ch) = atol(row[7]);
-      
+      GET_CHANNEL_POINTS(ch) = atoi(row[9]);
```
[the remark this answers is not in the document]
biscuitWizard: Yep! Will fix.

src/chargen.cpp
biscuitWizard: From discord: we're going to lock out all but the common core metatypes

src/chargen.cpp
biscuitWizard: image

src/utils.cpp
```diff
@@ -5819,6 +5836,74 @@ struct obj_data *make_new_finished_program(int part_type, int mpcp, int rating=0
   return prog;
 }
 
+struct obj_data *make_otaku_deck(struct char_data *ch) {
+  if (!ch) {
+    mudlog_vfprintf(ch, LOG_FUCKUPLOG, "Tried to create an otaku deck, but provided invalid character.");
```
luciensadi, line 5841: `LOG_SYSLOG` please, the fuckuplog is exclusively for players writing invalid commands that I need to add to the `fix_common_fuckups()` function later.

src/newmatrix.cpp
```diff
@@ -1798,6 +1855,10 @@ ACMD(do_logoff)
         send_to_char("Your hitcher has disconnected.\r\n", ch);
         temp->persona->decker->hitcher = NULL;
       }
+     // Clear the deck if this is an otaku
+    if (PERSONA->decker && PERSONA->decker->deck && PERSONA->decker->deck->obj_flags.extra_flags.IsSet(ITEM_EXTRA_OTAKU_RESONANCE)) {
```
luciensadi, line 1859: This is being called from the case `if (!PERSONA)`, so operating on `PERSONA->decker` will cause undefined behavior and/or a segfault.

src/newmatrix.cpp
```diff
@@ -1830,11 +1891,15 @@ ACMD(do_logoff)
   send_to_host(PERSONA->in_host, buf, PERSONA, FALSE);
 
   // Cleanup of uploads, downloads, etc is handled in icon_from_host, which is called in extract_icon.
-
   extract_icon(PERSONA);
   PERSONA = NULL;
   PLR_FLAGS(ch).RemoveBit(PLR_MATRIX);
 
+  // Clear the deck if this is an otaku
+  if (PERSONA->decker && PERSONA->decker->deck && PERSONA->decker->deck->obj_flags.extra_flags.IsSet(ITEM_EXTRA_OTAKU_RESONANCE)) {
```
luciensadi, line 1899: This is invoking after `PERSONA` is extracted and nulled, so this will cause undefined behavior and/or a segfault.

src/otaku.cpp
```diff
@@ -0,0 +1,197 @@
...
+      break;
+    case 3:
+      tke = 200;
+      break;
+    default:
+      tke = 200 + ((GET_GRADE(ch)-2)*200);
+      break;
+  }
+  if (tke > GET_TKE(ch)) {
+    send_to_char(ch, "You do not have high enough TKE to initiate. You need %ld TKE.\r\n", tke);
```
luciensadi, line 68: Is this function a copy/paste of the initiation cost code? If so, are the changes significant enough that code duplication is warranted, or can this be merged into the initiation cost code with a mode flag to switch between awakened and otaku initiations?

src/otaku.cpp
```diff
@@ -0,0 +1,197 @@
...
+    disp_submersion_menu(ch->desc);
+    return;
+  }
+}
+
+bool can_select_echo(struct char_data *ch, int i)
+{
+  // Count our grades vs selected echos so we can't learn more than we have grades
+  int available_echoes = GET_GRADE(ch);
+  for (int i = 1; i < ECHO_MAX; i++) {
```
luciensadi, line 116: Are you intending to shadow `i` here? This makes reading the code a bit confusing, so if we can swap this out to `idx` or something for clarity that'd be appreciated.

src/otaku.cpp
```diff
@@ -0,0 +1,197 @@
...
+  if (GET_ECHO(ch, i))
+    return FALSE;
+  return TRUE;
+}
+
+void disp_echo_menu(struct descriptor_data *d)
+{
+  CLS(CH);
+  for (int i = 1; i < ECHO_MAX; i++) {
+    if (PRF_FLAGGED(CH, PRF_SCREENREADER)) {
```
luciensadi, line 134: no notes, just a 👍 for thinking about screenreader usability here

src/cf_create.cpp
```diff
@@ -0,0 +1,331 @@
...
+  SOFT_VALIDATE,
+  SOFT_SWERVE,
+  SOFT_CLOAK,
+  SOFT_LOCKON
+};
+
+void cfedit_disp_menu(struct descriptor_data *d)
+{
+  CLS(CH);
+  send_to_char(CH, "1) Name: ^c%s^n\r\n", FORM->restring);
```
luciensadi, line 49: Is the form guaranteed to always exist and have a restring? If so this is fine, but otherwise this will crash. `GET_OBJ_NAME(FORM)` handles these edge cases and is generally safer.

src/chargen.cpp
```diff
@@ -916,6 +949,13 @@ int parse_assign(struct descriptor_data *d, const char *arg)
   if (d->ccr.pr[lowest_kosher_magic_slot] == PR_RACE)
     lowest_kosher_magic_slot--;
 
+  // Except for otaku who are always mundane lol
+  if (*arg == 2 && d->ccr.is_otaku && d->ccr.temp <= 1) {
```
luciensadi, line 953: What's the function of `d->ccr.temp` here? Is there ever a situation in which an otaku would be able to set magic?

src/chargen.cpp
```diff
@@ -1084,6 +1127,37 @@ static void start_game(descriptor_data *d, const char *origin)
...
+  // Otaku get some starting gear, so assign that here
+  if (d->ccr.is_otaku) {
+    // Equip cyberware (deduct essence and modify stats as appropriate)
+    obj_data *temp_obj;
+    for (int cyb = 0; cyb < OTAKU_CYBERWARE_NUM; cyb++) {
+      if (otaku_cyberware[cyb]) {
+        if (!(temp_obj = read_object(otaku_cyberware[cyb], VIRTUAL, OBJ_LOAD_REASON_ARCHETYPE))) {
+          mudlog_vfprintf(CH, LOG_SYSLOG, "SYSERR: Invalid cyberware item %ld specified for otaku class.",
+            otaku_cyberware[cyb]);
+          mudlog(buf, CH, LOG_SYSLOG, TRUE);
```
luciensadi, line 1139: Is there something specific in `buf` you're logging here, or is this a remnant of the old logging line?

src/chargen.cpp
luciensadi: Got it, is that worth writing in here somehow? I'm sure we're going to have folks who want to play "magic matrix characters" based on flavor without actually reading the otaku rules first.

luciensadi (commented):
27/27 files viewed, we're in the home stretch!

src/newmatrix.cpp
```diff
@@ -1897,13 +1962,36 @@ ACMD(do_connect)
   for (int i = 0; !cyberdeck && i < NUM_WEARS; i++)
     if (GET_EQ(ch, i) && (GET_OBJ_TYPE(GET_EQ(ch,i )) == ITEM_CYBERDECK || GET_OBJ_TYPE(GET_EQ(ch,i )) == ITEM_CUSTOM_DECK))
       cyberdeck = GET_EQ(ch, i);
+
+  if (!IS_SENATOR(ch) && *argument) {
```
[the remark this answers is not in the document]
biscuitWizard: Because of the hitchers change I need to actually make this more robust. Will incorporate these changes

src/cf_create.cpp
biscuitWizard: By the time it reaches this time in the execution area, yes. But I'll change it anyways as it shouldn't matter.

src/chargen.cpp
biscuitWizard: Artifact of copy+pasting code. Deleted!

src/chargen.cpp
biscuitWizard: Artifact. Fixed.

src/chargen.cpp
biscuitWizard: Added it to the help file!

PR #765, opened 2024-12-12
fix has_trigger handling for VCR decking penalties

jdevnull:
Previous implementation assumed that, in handler.cpp affect_total, has_trigger == 0 when there's no reflex trigger. This PR fixes has_trigger handling so that it properly results in matrix penalties when there's no reflex trigger (uses a conditional instead of re-assigning has_trigger when determining has_wires, and checks for has_trigger == -1).

This answers: https://discord.com/channels/564618629467996170/788953927269613608/1316639248421752872


luciensadi:
LGTM, thanks!

PR #767, opened 2025-02-08
Fix mental/physical recovery pos fail

jdevnull:
Mortally wounded characters cannot recover without medical attention, but unconscious characters can (SR3 pg 126). This PR removes the IS_NPC check since this should apply to all characters. Also and changes the `mental_gain` check from `POS_STUNNED` to `POS_MORTALLYW`, and the `physical_gain` check to also cover the possibility of `POS_DEAD`.

This answers the report: https://discord.com/channels/564618629467996170/788953927269613608/1337602787973795841

### DIFF
--- a/src/limits.cpp
+++ b/src/limits.cpp
@@ -65,7 +65,7 @@ void mental_gain(struct char_data * ch)
   }
 
   // Can't regenerate? Skip.
-  if (IS_NPC(ch) && GET_DEFAULT_POS(ch) <= POS_STUNNED) {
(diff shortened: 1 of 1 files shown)


jdevnull:
Oops, misunderstood what this code was supposed to do. This PR doesn't actually fix the bug report, since the mob's default state on spawning for the quest is mortally wounded. The question is, why does there need to be a check for the default state?

jdevnull:
Closing, will discuss in discord

PR #769, opened 2025-02-24
Made it so attribute maximums are calculated.

biscuitWizard:
This just moves the code that pulls from racial_limits into a util function for calculating the attribute maximums for a character. These changes are used for otaku, but spooled out into their own change for convenience.


luciensadi (approved):
LGTM

PR #770, opened 2025-02-26
Fixed Hitchikers

biscuitWizard:
Patched the hitcher code for matrix. List of changes:

- [x] Players can no longer move while connected to a hitcher
- [x] The syntax has been changed to 'connect <driver>' to connect as a hitcher to a driver (the person on the matrix)
- [x] say has been fixed for hitchers and now the messages are sent to the driver
- [x] hitchers now have a curated command list to secure what they can do, preventing erroneous states.  This can be expanded as desired.
- [x] score now shows if you have a hitcher connected or not

Example image of new UI:
image

### DIFF
--- a/src/act.comm.cpp
+++ b/src/act.comm.cpp
@@ -109,7 +109,7 @@ ACMD(do_say)
           ) {
           continue;
         }
-        if (!IS_IGNORING(targ, is_blocking_ic_interaction_from, ch)) // Gag check for ignored characters
+        if (!IS_IGNORING(targ, is_blocking_ic_interaction_from, ch)) {// Gag check for ignored characters
           // We found our hitcher
           snprintf(buf, sizeof(buf), "Your hitcher says, \"%s^n\"\r\n", capitalize(arg_known_size));
           send_to_char(buf, targ);
--- a/src/act.comm.cpp
+++ b/src/act.comm.cpp
@@ -100,17 +100,25 @@ ACMD(do_say)
         }
       }
     } else {
-      for (struct char_data *targ = get_ch_in_room(ch)->people; targ; targ = targ->next_in_room)
-        if (targ != ch && PLR_FLAGGED(targ, PLR_MATRIX) && !IS_IGNORING(targ, is_blocking_ic_interaction_from, ch)) {
-          // Send and store.
-          snprintf(buf, sizeof(buf), "Your hitcher says, \"%s^n\"\r\n", capitalize(arg_known_size));
-          send_to_char(buf, targ);
-          store_message_to_history(targ->desc, COMM_CHANNEL_SAYS, buf);
+      for (struct char_data *targ = get_ch_in_room(ch)->people; targ; targ = targ->next_in_room) {
+        if (targ == ch
+            || !PLR_FLAGGED(targ, PLR_MATRIX)
+            || !targ->persona
+            || !targ->persona->decker
+            || targ->persona->decker->hitcher != ch
+          ) {
+          continue;
         }
-      // Send and store.
-      snprintf(buf, sizeof(buf), "You send, down the line, \"%s^n\"\r\n", capitalize(arg_known_size));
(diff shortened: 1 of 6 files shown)


src/act.comm.cpp
```diff
@@ -100,17 +100,25 @@ ACMD(do_say)
...
+            || targ->persona->decker->hitcher != ch
+          ) {
+          continue;
         }
-      // Send and store.
-      snprintf(buf, sizeof(buf), "You send, down the line, \"%s^n\"\r\n", capitalize(arg_known_size));
-      send_to_char(buf, ch);
-      store_message_to_history(ch->desc, COMM_CHANNEL_SAYS, buf);
+        // We found our hitcher
+        snprintf(buf, sizeof(buf), "Your hitcher says, \"%s^n\"\r\n", capitalize(arg_known_size));
```
luciensadi, line 113: Please add the `!IS_IGNORING(targ, is_blocking_ic_interaction_from, ch)` gag back to this. The hitcher should still think their message went through properly, but it should not print to the decker's screen.

Edit: I see that you shifted the ignore bit below to being handled when the hitcher connects, but there's also a case where someone starts ignoring a hitcher who is already connected.

src/interpreter.cpp
```diff
@@ -1293,6 +1293,52 @@ ACMD_DECLARE(do_talk);
...
 ACMD_DECLARE(do_fry_self);
 
+struct command_info mtx_htr_info[] =
+  {
+    { "RESERVED", 0, 0, 0, 0
+    , BLOCKS_IDLE_REWARD },
+    { "afk", 0, do_afk, 0, 0, ALLOWS_IDLE_REWARD },
+    { "disconnect", 0, do_logoff, 0, 1, BLOCKS_IDLE_REWARD },
+    { "logoff", 0, do_logoff, 0, 0, BLOCKS_IDLE_REWARD },
+    { "emote", 0, do_echo, 0, SCMD_EMOTE , BLOCKS_IDLE_REWARD },
```
luciensadi, line 1303: What's the effect of using the emote command while hitching? It just prints to the room you're in, right? Is this for people trying to interact with a hitching character from meatspace, and if so, is the hitcher expected to be able to see/interact with those PCs?

src/newmatrix.cpp
```diff
@@ -1864,36 +1880,41 @@ ACMD(do_connect)
...
-        }
-      }
+  // hitcher code, new syntax is connect <dude>
+  if (*argument && !PLR_FLAGGED(ch, PLR_MATRIX)) {
+    skip_spaces(&argument);
+    
+    temp = get_char_room_vis(ch, argument);
+    if (IS_IGNORING(temp, is_blocking_ic_interaction_from, ch)) temp = NULL;
+    if (!temp) {
+      send_to_char(ch, "You don't see anyone named '%s' here.\r\n", argument);
```
luciensadi, line 1890: The goal is for the ignored person to not know they're being ignored, so when the ignore bit is set, please use an error like "It doesn't look like you can hitch a ride with %s" below. Otherwise they'll know that they're being blocked, because the person they're trying to connect to is standing in front of them.

luciensadi (commented):
Looking good, thanks for putting this together so quickly!

src/interpreter.cpp
biscuitWizard: It actually doesn't let them emote. If you try to emote  (I tested every command) it just says you can't do that while hitching, because there was an exception already installed in do_emote. Since that code was already written I figured I wouldn't remove it.

src/act.comm.cpp
biscuitWizard: I didn't consider that and will do that.

src/newmatrix.cpp
biscuitWizard: Will fix!

luciensadi:
Thanks for the code!

PR #773, opened 2025-03-04
Fixed uploads not processing from otaku proxy decks

biscuitWizard:
When uploading from a proxy deck the upload does not progress. This patch fixes that issue by redirecting the upload to the proxy deck since persona->decker->deck on otaku will never (in the current code) have storage memory.

If living persona ever get memory this will need to be modified.

### DIFF
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -3041,7 +3041,7 @@ void process_upload(struct matrix_icon *persona)
           // Make it seen by them.
           GET_DECK_ACCESSORY_FILE_FOUND_BY(soft) = GET_IDNUM(persona->decker->ch);
           // Remove it from their deck's used storage.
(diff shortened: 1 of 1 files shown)


src/newmatrix.cpp
```diff
@@ -2983,7 +2983,11 @@ void process_upload(struct matrix_icon *persona)
 {
   // Note: We only upload one file at a time. Find the first valid one and process it, then stop.
   if (persona && persona->decker && persona->decker->deck) {
-    for (struct obj_data *soft = persona->decker->deck->contains, *next_obj; soft; soft = next_obj) {
+    obj_data *deck = persona->decker->deck;
+    if (persona->decker->proxy_deck)
+      deck = persona->decker->proxy_deck;
```
luciensadi, line 2988: If you're making this change, you should propagate it to the rest of this block. There's a call to change storage memory below that still hits the original deck.

biscuitWizard: I even read the code and still didn't see that. I am just getting old and my eyesight is going, obviously.

PR #775, opened 2025-03-04
Otaku fixes 4

biscuitWizard:
Yet another round of otaku fixes.

This one fixes the detection factor problem vile noticed.

It also finally persists echoes and grades to db, so when you logoff you don't lose your echoes or submersion grade.


biscuitWizard:
image
fancy UI change, proxy decks now appear at bottom. Living persona no longer have a storage line (to prevent confusion)

PR #776, opened 2025-03-04
Added drug handling to otaku

biscuitWizard:
got bored; decided it would be fun if otaku could enjoy drugs too.

thought it would be simple.

it was not simple.

### DIFF
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -32,6 +32,7 @@
 #include "deck_build.hpp"
 #include "vehicles.hpp"
 #include "player_exdescs.hpp"
+#include "otaku.hpp"
 
 /* external functions */
 extern void stop_fighting(struct char_data * ch);
@@ -863,6 +864,11 @@ void affect_total(struct char_data * ch)
   rigger_init_dice -= GET_INIT_DICE(ch);
 
   apply_drug_modifiers_to_ch(ch);
+
+  // Part of the otaku simulation; modifies otaku attributes depending on drugs, cyber, etc.
+  if (ch->persona && ch->persona->type == ICON_LIVING_PERSONA) {
+    update_otaku_deck(ch, ch->persona->decker->deck);
+  }
 
   rigger_rea += GET_REA(ch);
   rigger_init_dice += GET_INIT_DICE(ch);
--- a/src/handler.cpp
+++ b/src/handler.cpp
@@ -866,7 +866,10 @@ void affect_total(struct char_data * ch)
   apply_drug_modifiers_to_ch(ch);
 
   // Part of the otaku simulation; modifies otaku attributes depending on drugs, cyber, etc.
-  if (ch->persona && ch->persona->type == ICON_LIVING_PERSONA) {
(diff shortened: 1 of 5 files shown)


src/otaku.cpp
```diff
@@ -129,6 +183,54 @@ struct obj_data *make_otaku_deck(struct char_data *ch) {
...
 
+void update_otaku_deck(struct char_data *ch, struct obj_data *cyberdeck) {
+  int mpcp = GET_OTAKU_MPCP(ch);
+
+  GET_CYBERDECK_MPCP(cyberdeck) = MIN(12, mpcp);
+  GET_CYBERDECK_HARDENING(cyberdeck) = MIN(get_otaku_wil(ch), get_otaku_wil(ch) / 2 + GET_ECHO(ch, ECHO_IMPROVED_HARD));
+  GET_CYBERDECK_ACTIVE_MEMORY(cyberdeck) = 0; // Otaku do not have active memory.
+  GET_CYBERDECK_TOTAL_STORAGE(cyberdeck) = 0;
+  GET_CYBERDECK_RESPONSE_INCREASE(cyberdeck) = MIN(mpcp * 1.5, get_otaku_rea(ch) + GET_ECHO(ch, ECHO_IMPROVED_REA));
+  GET_CYBERDECK_IO_RATING(cyberdeck) = MIN(get_otaku_int(ch) * 200, (get_otaku_int(ch) * 100) + (GET_ECHO(ch, ECHO_IMPROVED_IO) * 100));
```
luciensadi, line 194: Remember that `MIN` and `MAX` are hacky macros that execute their contents between 1-2 times per branch. This expands to the equivalent of this function:

```c++
if (get_otaku_int(ch) * 200 <= (get_otaku_int(ch) * 100) + (GET_ECHO(ch, ECHO_IMPROVED_IO) * 100)) {
  return get_otaku_int(ch) * 200;
} else {
  return (get_otaku_int(ch) * 100) + (GET_ECHO(ch, ECHO_IMPROVED_IO) * 100);
}
```

These are relatively lightweight functions and are idempotent, but if update_otaku_deck() gets called a lot in a single tick, it'll contribute to the tick lag. Best practice for a single-threaded game like this is to save the output of reused idempotent functions to a local variable where possible.

src/handler.cpp
```diff
@@ -864,6 +865,11 @@ void affect_total(struct char_data * ch)
 
   apply_drug_modifiers_to_ch(ch);
 
+  // Part of the otaku simulation; modifies otaku attributes depending on drugs, cyber, etc.
+  if (ch->persona && ch->persona->type == ICON_LIVING_PERSONA) {
+    update_otaku_deck(ch, ch->persona->decker->deck);
```
luciensadi, line 870: Please check for `persona->decker` and `persona->decker->deck`. An ugly error message is always better than a crash!

src/otaku.cpp
luciensadi: Not a required change, just a point for future dev work.

PR #778, opened 2025-03-05
Otaku Fixes Round 5

biscuitWizard:
Commits contain individual file fixes

### DIFF
--- a/src/otaku.cpp
+++ b/src/otaku.cpp
@@ -25,6 +25,11 @@ extern struct otaku_echo echoes[];
 int get_otaku_wil(struct char_data *ch) {
   int wil_stat = GET_REAL_WIL(ch);
 
+  /* Handling Cyberware/Bioware */
+  struct obj_data *pain_editor = find_bioware(ch, BIO_PAINEDITOR);
+  if (pain_editor && GET_BIOWARE_RATING(pain_editor))
+    wil_stat++;
+
   /* Handling Drugs */
   int detox_force = affected_by_spell(ch, SPELL_DETOX);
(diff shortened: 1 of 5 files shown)


src/otaku.cpp
```diff
@@ -25,6 +25,11 @@ extern struct otaku_echo echoes[];
 int get_otaku_wil(struct char_data *ch) {
   int wil_stat = GET_REAL_WIL(ch);
 
+  /* Handling Cyberware/Bioware */
+  struct obj_data *pain_editor = find_bioware(ch, BIO_PAINEDITOR);
+  if (pain_editor && GET_BIOWARE_RATING(pain_editor))
```
luciensadi, line 30: Is PE activation tracked in rating, or is it on bioware flags instead?

src/spec_procs.cpp
```diff
@@ -5680,6 +5680,12 @@ SPECIAL(chargen_skill_annex) {
       return TRUE;
     }
 
+    if (IS_OTAKU(ch) && GET_SKILL(ch, SKILL_COMPUTER) < 6) {
+      send_to_char(ch, "Otakus must have at least a rating 6 in Computers, but your proficiency is only %d. You should ^WPRACTICE^n your Computers before proceeding.\r\n",
```
luciensadi, line 5684: If this is a hard requirement, can we make them start with 6 computers and -6 skill points by default, and refuse to let them create with any less than 6 skill points?

luciensadi:
Merging this for testing, but please consider the feedback for another PR.

PR #780, opened 2025-03-06
Otaku fixes 6

biscuitWizard:
Change notes are in the commit messages

### DIFF
--- a/src/cf_create.cpp
+++ b/src/cf_create.cpp
@@ -174,7 +174,7 @@ void cfedit_parse(struct descriptor_data *d, const char *arg)
       send_to_char(CH, "You can't create a complex form of a higher rating than your living persona's MPCP rating.\r\n"
                    "Enter Rating: ");
     } else if (option_n <= 0) {
-      send_to_char(CH, "You can't create a complex form less than 0, you silly bastard.\r\n"
+      send_to_char(CH, "You can't create a complex form with a rating less than 0.\r\n"
                    "Enter Rating: ");
     } else {
       GET_DESIGN_RATING(d->edit_obj) = option_n;
--- a/src/cf_create.cpp
+++ b/src/cf_create.cpp
@@ -168,10 +168,13 @@ void cfedit_parse(struct descriptor_data *d, const char *arg)
     break;
   case CFEDIT_RATING:
     if (option_n > GET_SKILL(CH, SKILL_COMPUTER)) {
-      send_to_char(CH, "You can't create a program of a higher rating than your computer skill.\r\n"
+      send_to_char(CH, "You can't create a complex form of a higher rating than your computer skill.\r\n"
                    "Enter Rating: ");
     } else if (option_n > GET_OTAKU_MPCP(CH)) {
-      send_to_char(CH, "You can't create a program of a higher rating than your living persona's MPCP rating.\r\n"
+      send_to_char(CH, "You can't create a complex form of a higher rating than your living persona's MPCP rating.\r\n"
+                   "Enter Rating: ");
+    } else if (option_n <= 0) {
+      send_to_char(CH, "You can't create a complex form less than 0, you silly bastard.\r\n"
                    "Enter Rating: ");
     } else {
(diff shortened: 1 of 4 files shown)


src/cf_create.cpp
```diff
@@ -168,10 +168,13 @@ void cfedit_parse(struct descriptor_data *d, const char *arg)
...
     if (option_n > GET_SKILL(CH, SKILL_COMPUTER)) {
-      send_to_char(CH, "You can't create a program of a higher rating than your computer skill.\r\n"
+      send_to_char(CH, "You can't create a complex form of a higher rating than your computer skill.\r\n"
                    "Enter Rating: ");
     } else if (option_n > GET_OTAKU_MPCP(CH)) {
-      send_to_char(CH, "You can't create a program of a higher rating than your living persona's MPCP rating.\r\n"
+      send_to_char(CH, "You can't create a complex form of a higher rating than your living persona's MPCP rating.\r\n"
+                   "Enter Rating: ");
+    } else if (option_n <= 0) {
+      send_to_char(CH, "You can't create a complex form less than 0, you silly bastard.\r\n"
```
luciensadi, line 177: Funny, but please drop the ad-hom and give it a dispassionate `You can't create a complex form less than 0` for both of these. Game culture can be really influenced by the code, so if the game starts knocking people...

src/chargen.cpp
```diff
@@ -1126,8 +1126,12 @@ static void start_game(descriptor_data *d, const char *origin)
     GET_LOADROOM(d->character) = RM_CHARGEN_START_ROOM;
   }
 
-  // Otaku get some starting gear, so assign that here
+  // Otaku get some starting gear and skills, so assign that here
   if (d->ccr.is_otaku) {
+    // Assign starting computers to be helpful
+    GET_SKILL_POINTS(d->character) -= 6;
```
luciensadi, line 1132: What if someone has less than 6 skill points at creation? Do you enforce the 6-point minimum?

src/spec_procs.cpp
```diff
@@ -5680,24 +5680,33 @@ SPECIAL(chargen_skill_annex) {
...
-    for (int ci=SKILL_CHANNEL_ACCESS; ci <= SKILL_CHANNEL_SLAVE;ci++) {
-      if (REAL_SKILL(ch, ci) <= 3) continue;
-      for (int csi=SKILL_CHANNEL_ACCESS; csi <= SKILL_CHANNEL_SLAVE;csi++) {
-        if (csi == ci) continue; // skip checking the parent skill against itself.
-        if (REAL_SKILL(ch, csi) == REAL_SKILL(ch, ci)) {
-          send_to_char(ch, "You cannot have more than one channel skill at %d for otaku at character generation. Please reduce the value of %s.\r\n",
-            REAL_SKILL(ch, ci), skills[ci].name);
+      for (int i = SKILL_CORPORATE_ETIQUETTE; i <= SKILL_ELF_ETIQUETTE; i++) {
+        if (GET_SKILL(ch, i) > 4) {
+          send_to_char(ch, "Socially maladjusted decker children cannot have an etiquette skill higher than 4. You should ^WUNPRACTICE^n your %s before proceeding.\r\n", skills[i].name);
```
luciensadi, line 5692: I know the lore has otakus as children, but we don't allow characters under 18 in the game, so I've been envisioning these as adults. Does that align with what you've been designing?

src/chargen.cpp
biscuitWizard: Have to. :/ That's the rules. The south exit on chargen won't let you proceed with negative skillpoints too.

src/spec_procs.cpp
biscuitWizard: Yes. I was just being silly.

src/chargen.cpp
biscuitWizard: I believe they can sell english skill points though to afford it.

luciensadi: Can we force folks to have at least 6 skill points when moving from the creation menus to the game? I don't want folks cheesing it by unlearning English to meet their minimum.

PR #781, opened 2025-03-07
Otaku fixes 7

biscuitWizard:
Fixes present in commit messages

### DIFF
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -17,7 +17,6 @@
 #include "moderation.hpp"
 #include "pets.hpp"
 #include "otaku.hpp"
-#include "matrix_storage.hpp"
 
 #define PERSONA ch->persona
 #define PERSONA_CONDITION ch->persona->condition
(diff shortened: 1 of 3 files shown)


src/newmatrix.cpp
```diff
@@ -834,10 +837,25 @@ void matrix_fight(struct matrix_icon *icon, struct matrix_icon *targ)
         bod = targ->decker->masking;
         break;
       }
+      if (targ->type == ICON_LIVING_PERSONA)
+        bod += targ->decker->hardening; // Otaku get to add their hardening to this check.
```
luciensadi, line 841: They do? This feels OP, can't they get to hardening-12 for a completely unreachable bod number?

src/newmatrix.cpp
```diff
@@ -834,10 +837,25 @@ void matrix_fight(struct matrix_icon *icon, struct matrix_icon *targ)
...
       success = success_test(matrix[targ->in_host].security, bod);
       int resist = success_test(bod + MIN(GET_MAX_HACKING(targ->decker->ch), GET_REM_HACKING(targ->decker->ch)), iconrating);
       GET_REM_HACKING(targ->decker->ch) = MAX(0, GET_REM_HACKING(targ->decker->ch) - GET_MAX_HACKING(targ->decker->ch));
       success -= resist;
+      if (targ->type == ICON_LIVING_PERSONA) {
+        int damage_total = success / 2;
+        if (!damage_total) {
```
luciensadi, line 848: Using a negation check here allows a negative damage through. `if (damage_total <= 0)` is the better choice. Same comment for `!damage_total` below (which I wouldn't expect to ever go negative out of convert_damage(), but may as well update it)

PR #782, opened 2025-03-08
Otaku fixes 8

biscuitWizard:
Thanks for submitting code! It's highly recommended that you discuss changes / new features with Lucien before merging, so if you haven't done that already, please reach out to him in Discord.

Please include in this PR a description of what your code changes (e.g. "Colors all magazines bright purple for easier locating in inventory lists") or what it's intended to do. The more descriptive you can be, the easier it will be for us to review your code!


src/cf_create.cpp
```diff
@@ -26,22 +26,17 @@ extern int get_program_skill(char_data *ch, obj_data *prog, int target);
...
-  SOFT_ARMOR,
-  SOFT_CAMO,
-  SOFT_CRASH,
-  SOFT_DEFUSE,
-  SOFT_EVALUATE,
-  SOFT_VALIDATE,
-  SOFT_SWERVE,
-  SOFT_CLOAK,
-  SOFT_LOCKON
+  SOFT_SHIELD
```
luciensadi, line 39: There are only 11 entries here, but 16 defined above in COMPLEX_FORM_TYPES. Does this need to be reduced?

src/newmatrix.cpp
```diff
@@ -719,6 +719,29 @@ ACMD(do_matrix_position)
...
+        ICON_IS_IC(icon) ? matrix[icon->in_host].security : GET_SKILL(icon->decker->ch, SKILL_COMPUTER));
+
+      if (shield_test > 0) {
+        success -= shield_test;
+        send_to_icon(targ, "You raise your shield program and deflect some of the attack.\r\n");
+      }
+      GET_PROGRAM_RATING(soft)--;
+      if (GET_PROGRAM_RATING(soft) <= 0) {
+        send_to_icon(targ, "Your shield program crashes as the rating is depleted.\r\n");
+        extract_obj(soft);
```
luciensadi, line 737: Does extracting it in the matrix like this permanently remove it, or does it just remove the loaded copy?

Also, is the rating decrease permanent, or does it only effect what's loaded at the time?

src/newmatrix.cpp
```diff
@@ -2398,24 +2427,24 @@ ACMD(do_connect)
...
         !proxy_deck && cyberdeck->obj_flags.extra_flags.IsSet(ITEM_EXTRA_OTAKU_RESONANCE) ? "the jackpoint" : "$s cyberdeck");
     }
   } else {
-    snprintf(buf, sizeof(buf), "$n plugs the leads of $s 'trode net into $s cyberdeck.");
+    snprintf(buf, sizeof(buf), "$n plugs the leads of $s 'trode net into $s cyberdeck.\r\n");
   }
   act(buf, FALSE, ch, 0, 0, TO_ROOM);
   if (cyberdeck->obj_flags.extra_flags.IsSet(ITEM_EXTRA_OTAKU_RESONANCE)) 
-    act("You jack into the matrix to commune with the resonance.", FALSE, ch, 0, 0, TO_CHAR);
+    act("You jack into the matrix to commune with the resonance.\r\n", FALSE, ch, 0, 0, TO_CHAR);
```
luciensadi, line 2445: `act()` auto-appends `\r\n`. You only need to add it for `send_to_char()` calls.

edit: it looks like all of the messages you added \r\n to here are acts?

src/newmatrix.cpp
biscuitWizard: It only removes the loaded copy, and it only affects the loaded copy.

src/newmatrix.cpp
biscuitWizard: I was encountering a problem despite that with room names appending on connection. I don't know if that's because of shared buffers or what, but the only fix was adding newlines?

src/cf_create.cpp
biscuitWizard: Good catch!

src/newmatrix.cpp
biscuitWizard: Nevermind I figured out why. Fixing

PR #785, opened 2025-03-13
[Cyberware] Medial Cortex Bridge

biscuitWizard:
Just a simple PR adding the Homebrew Medial Cortex Bridge from Shadowrun: Denver to here.

```
This network of cultured neurons bridges the medial prefrontal cortex and the hypothalamus to give these parts of the brain more control over the amygdala. Additionally, modified adrenal medullae alter the hormones produced during panic situations. The character is less prone to panic attacks, stress, or other mental manipulations. Also referred to as the MCT Bridge.
```

It's like a Cerebral Booster, but for WIL.

Didn't discuss before writing it; reject PR if not a good fit for the game.

### DIFF
--- a/lib/world/obj/5.obj
+++ b/lib/world/obj/5.obj
@@ -2690,4 +2690,66 @@ Material:	metal
 	Location:	Strength
 	Modifier:	2
 BREAK
+#593
+Keywords:	medial cortex bridge mct
+Name:	a medial cortex bridge I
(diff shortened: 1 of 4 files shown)


luciensadi:
Thanks for the code, but this doesn't fit with our current balance. Denver has a much more resources-restricted environment where the tradeoffs of using something like this would make sense.

PR #789, opened 2025-03-16
Health recovery rebalancing

jdevnull:
This PR tones down the maximum possible mental/physical health recovery by making similar bonuses additive instead of multiplicative:
- symbiotes add with adept rapid healing
- room bonuses add with other room bonuses

Characters who aren't stacking everything should see no or little difference in their recovery rates.

### DIFF
--- a/src/limits.cpp
+++ b/src/limits.cpp
@@ -106,40 +106,53 @@ void mental_gain(struct char_data * ch)
   } else {
     // For PCs, we have to calculate.
 
-    // Augmentations
-    for (struct obj_data *bio = ch->bioware; bio; bio = bio->next_content) {
-      if (GET_BIOWARE_TYPE(bio) == BIO_SYMBIOTES) {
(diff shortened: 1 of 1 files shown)


luciensadi:
Thanks for the code! I'm very much on the fence about this one-- it's cheesy af to stack healing factors, but it doesn't really hurt anything to do so. Is there a specific impact you're looking to mitigate with this?

jdevnull:
There's not really a specific thing that I want to solve with this, just a general balance thing to tone down the upper end by around a factor of 2 (that is, healing from 1 to 10 will take closer to 4 minutes instead of 2).

luciensadi:
Got it! I've thought this over, and while I'm on board with not letting power creep too much, fast healing feels relatively harmless. I'm inclined to close this out for now and revisit if fast healing becomes problematic in the future; what do you think?

jdevnull:
I don't have strong feelings about it either way, I'll go ahead and close.

PR #790, opened 2025-03-16
Various matrix things

jdevnull:
~~1. Previously, crippler/ripper damage to otaku was very unlikely to happen, except when a race condition occurs (see: 4 below). In the scenario of an early-game otaku with 7 in the relevant attribute, hacking pool set to 3, and 12 soak dice (body + cpool), that is in a Red host with rating 6 security (existing Red hosts at this rating are difficult but accessible by early game deckers), and is successfully hit by a rating 6 crippler/ripper, they would only have a 0.15% chance of taking any damage at all. This PR removes the second resistance test, making it a 6.5% chance. Conceptually, this would work more like powerbolt (where you only have a spell resistance test and no soak test).~~

~~It's unclear what the desired probability of damage is, but we can increase it to 21% in the given scenario by removing the /2 a few lines before the deleted portion.~~

2. Previously, otaku system tests were being capped at min 2 before other penalties were applied (like host alert and wound penalties). This is removed here so that otaku can actually reach the min TN 2 while taking penalties.

3. Some cleanup in system_test - no need to search for decker utility programs with living personae since they use channels instead, and no reason to search for sleaze when `get_detection_factor(ch)` already does that.

4. When hit by a crippler/ripper IC, living persona would have their attribute lowered, but then reset by the next `affect_total` -> `update_otaku_deck`. If the crippler/ripper IC hits while the attribute is lowered, number of successes rolled by the IC can rise sharply. Per https://discord.com/channels/564618629467996170/797657781507194880/1350710279071203338 , crippler/ripper IC should only deal damage, not reduce persona attributes.

### DIFF
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -881,8 +881,7 @@ void matrix_fight(struct matrix_icon *icon, struct matrix_icon *targ)
         send_to_icon(targ, "It tears into your living persona!\r\n");
         do_damage_persona(targ, damage_total);
         return;
-      }
-      if (success >= 2) {
+      } else if (success >= 2) {
         send_to_icon(targ, "It tears into the program!\r\n");
         while (success >= 2) {
           success -= 2;
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -875,10 +875,6 @@ void matrix_fight(struct matrix_icon *icon, struct matrix_icon *targ)
         if (damage_total <= 0) {
           return send_to_icon(targ, "It fails to damage your living persona.\r\n");
         }
-        damage_total = convert_damage(stage((2 - success_test(GET_BOD(targ->decker->ch) + GET_BODY_POOL(targ->decker->ch), MIN(matrix[icon->in_host].color + 1, DEADLY))), damage_total));
-        if (damage_total <= 0) {
-          return send_to_icon(targ, "You power through it and nullify the damage.\r\n");
-        }
         send_to_icon(targ, "It tears into your living persona!\r\n");
         do_damage_persona(targ, damage_total);
         return;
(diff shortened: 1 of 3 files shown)


src/newmatrix.cpp
```diff
@@ -443,7 +445,17 @@ int system_test(rnum_t host, struct char_data *ch, int type, int software, int m
         channel_rating += GET_SKILL(ch, SKILL_CHANNEL_SLAVE);
         break;
     }
-    target = MAX(2, target - channel_rating);
+    target -= channel_rating;
+  } else {
```
biscuitWizard, line 449: Good change for clarity

src/newmatrix.cpp
```diff
@@ -443,7 +445,17 @@ int system_test(rnum_t host, struct char_data *ch, int type, int software, int m
...
+    // ... or by the appropriate decker utility program
+    for (struct obj_data *soft = DECKER->software; soft; soft = soft->next_content) {
+      if (GET_PROGRAM_TYPE(soft) == software) {
```
biscuitWizard, line 452: Nit: Inverse this check to reduce nesting. So ```if (GET_PROGRAM_TYPE(soft) != software) continue;```

jdevnull:
5. Some minor refactoring of matrix_fight.
- Removed duplication and potential issues with using the "dam" variable for three different things.
- Standardized matrix IC damage by host color (looks like we were accidentally using the dumpshock damage levels (L, M, S, D, D) instead of the ones used by IC (M, M, S, S, +2S), SR3 pg 226, 228-230.
  - Houseruled to (L, M, S, S, +2S) to avoid increasing the difficulty of blue hosts.
- Re-implemented crippler/ripper vs living personae similar to Killer IC (discussed at https://discord.com/channels/564618629467996170/564621949842751508/1350811801905922081 )
- A bunch of additional comments

6. Neurofilters provide ICCM functionality, so use the ICCM from otaku deck (but ICCM isn't in the otaku deck yet, oops!)

PR #793, opened 2025-03-19
Added GMCP Support with Protocols

biscuitWizard:
This adds support for the GMCP. Essentially it outputs JSON data on certain game events to enable enhanced clients. Basically events are emitted to the client with different module names (IE: Core.Hello) and then a JSON blob.

## Client Calls:
Many if not all the modules support being called with '.Get' at the end in order to manually trigger them via the client.

The supported/added modules are listed below:
## Core.Supports
```json
{
  "Core": ["Supports"],
  "Room": ["Info", "Exits"],
  "Char": ["Status", "Info", "Pools"],
  "Matrix": ["Info", "Deck"]
}
```

## Room.Info
```
{
  coords = {
    x = 20,
    y = 20,
    z = 2.5
  },
  description = "   The pavement and sidewalks of Cambridge Drive are relatively clean\r\n(compared 
to many other roads you have seen), and numerous street lamps line\r\nthe sides to illumine it 
during the night.  No cars are allowed onto\r\nCambridge, due to its narrow width. \r\n",
  exits = [ {
      direction = "n",
      state = "OPEN"
    }, {
      direction = "se",
      state = "OPEN"
    }, {
      direction = "s",
      state = "CLOSED"
    } ],
  name = "Cambridge Drive",
  vnum = 2087,
  zone = {
    name = "West Tacoma",
    number = 20
  }
}
```

## Room.Exits
```json
{
  "room_name": "blar",
  "room_vnum": "123",
  "exits": [
    {"direction": "east", "state": "OPEN"},
    {"direction": "north", "state": "INACCESSIBLE"}
  ]
}
```

## Char.Vitals
```
{
  ammo = 0,
  armor = {
    ballistic = 0,
    impact = 0
  },
  carrying_weight = 4.5299997329712,
  domain = "N/A",
  karma = 0,
  max_ammo = 0,
  max_carry_weight = 530,
  mental = 1000,
  mental_max = 1000,
  nuyen = 0,
  physical = 1000,
  physical_max = 1000
}
```

## Char.Info
```
{
  height = 1.75,
  metatype = "Human",
  name = "Nyx",
  noto = 0,
  pronouns = "Female",
  rep = 0,
  syspoints = 5320,
  tke = 0,
  weight = 59
}
```

## Char.Pools
```
{
  pools = {
    combat = {
      body = 0,
      dodge = 8,
      max = 75,
      offense = 67
    },
    hacking = 19
  },
  task_pools = {
    Body = 0,
    Charisma = 0,
    Intelligence = 0,
    Quickness = 0,
    Reaction = 0,
    Strength = 0,
    Willpower = 0
  }
}
```

## Matrix.Info
Sends information about current host
```json
{
  "name": "Seattle LTG",
  "vnum": 2934,
  "color": 1,
  "icons": [
    "Black ice drone"
  ]
}
```

## Matrix.Deck
Sends information about current deck/persona
```json
{
  "deck": {
    "name": "a CMT avatar deck",
    "mpcp": 6,
    "bod": 6,
    "masking": 6,
    "evasion": 6,
    "sensor": 6,
    "io": 100,
    "programs": [
      {"name": "attack program", "rating": 6, "type": 12},
      ....
    ]
  },
  "persona": {
    "name": "a living avatar",
    "condition": 1000
  }
}
```

### DIFF
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -939,7 +939,11 @@ void game_loop(int mother_desc)
       // Send GMCP Vitals
       for (d = descriptor_list; d; d = next_d) {
         next_d = d->next;
-        SendGMCPCharVitals(d->character);
+        if (!d->prompt_mode) continue;
+        if (!d->character) continue;
+
+        if (d->character->in_room)
+          SendGMCPCharVitals(d->character);
       }
     }
 
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -942,8 +942,10 @@ void game_loop(int mother_desc)
         if (!d->prompt_mode) continue;
         if (!d->character) continue;
 
-        if (d->character->in_room)
+        if (d->character->in_room) {
           SendGMCPCharVitals(d->character);
+          SendGMCPCharPools(d->character);
+        }
       }
     }
 
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -942,10 +942,8 @@ void game_loop(int mother_desc)
         if (!d->prompt_mode) continue;
         if (!d->character) continue;
 
-        if (d->character->in_room) {
-          SendGMCPCharVitals(d->character);
-          SendGMCPCharPools(d->character);
-        }
+        SendGMCPCharVitals(d->character);
+        SendGMCPCharPools(d->character);
       }
     }
 
--- a/src/comm.cpp
+++ b/src/comm.cpp
@@ -75,6 +75,7 @@
 #include "newhouse.hpp"
 #include "factions.hpp"
 #include "player_exdescs.hpp"
+#include "gmcp.hpp"
 
 
 const unsigned perfmon::kPulsePerSecond = PASSES_PER_SEC;
@@ -934,6 +935,14 @@ void game_loop(int mother_desc)
       }
     }
 
+    {
+      // Send GMCP Vitals
+      for (d = descriptor_list; d; d = next_d) {
+        next_d = d->next;
+        SendGMCPCharVitals(d->character);
+      }
+    }
+
     /* handle heartbeat stuff */
     /* Note: pulse now changes every 0.10 seconds  */
 
@@ -1359,6 +1368,7 @@ int make_prompt(struct descriptor_data * d)
       prompt = GET_PROMPT(d->character);
       ch = d->character;
     }
+    
     if (!prompt || !*prompt)
       data = "> ";
     else if (!strchr(prompt, '@')) {
--- a/src/gmcp.cpp
+++ b/src/gmcp.cpp
@@ -81,35 +81,41 @@ void SendGMCPCharInfo( struct char_data * ch )
   SendGMCP(ch->desc, "Char.Info", payload.c_str());
 }
 
+json SerializeRoomExits( struct room_data *room )
+{
+  json j;
+  
+  for (int door = 0; door < NUM_OF_DIRS; door++) {
+    if (!room->dir_option[door]) continue;
+    if (!room->dir_option[door]->to_room) continue;
+    if (room->dir_option[door]->to_room == &world[0]) continue;
+    if (IS_SET(room->dir_option[door]->exit_info, EX_HIDDEN)) continue;
+    
+    json exit = json::object();
+    exit["direction"] = exitdirs[door];
+    if (IS_SET(room->dir_option[door]->exit_info, EX_LOCKED)) {
+      exit["state"] = "LOCKED";
+    } else if (IS_SET(room->dir_option[door]->exit_info, EX_CLOSED)) {
+      exit["state"] = "CLOSED";
+    } else {
+      exit["state"] = "OPEN";
+    }
+
+    j.push_back(exit);
+  }
+
+  return j;
+}
+
 void SendGMCPExitsInfo( struct char_data *ch ) 
 {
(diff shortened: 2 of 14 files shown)


src/gmcp.cpp
```diff
@@ -0,0 +1,189 @@
...
+void SendGMCPCharStatus( struct char_data * ch )
+{
+  if (!ch || !ch->desc || !ch->desc->pProtocol->bGMCP) return;
+  json j;
+
+  j["physical"] = GET_PHYSICAL(ch);
+  j["mental"] = GET_MENTAL(ch);
+  j["physical_max"] = GET_MAX_PHYSICAL(ch);
+  j["mental_max"] = GET_MAX_MENTAL(ch);
+  j["pools"] = {"body", GET_BODY_POOL(ch), "magic", GET_MAGIC_POOL(ch), "combat", GET_COMBAT_POOL(ch), "hacking", GET_HACKING(ch)};
```
luciensadi, line 166: Folks probably want to see this split into the broken-down pools (combat pool is Dodge, Body, Offense, etc)

src/gmcp.cpp
```diff
@@ -0,0 +1,189 @@
...
+  j["task_pools"] = json::object();
+  for (int x = 0; x < 7; x++)
+    j["task_pools"][attributes[x]] = GET_TASK_POOL(ch, x);  
+
+  // Dump the json to a string and send it.
+  std::string payload = j.dump();
+  SendGMCP(ch->desc, "Char.Status", payload.c_str());
+}
+
+void SendGMCPCharPools( struct char_data * ch )
```
luciensadi, line 176: This seems like duplication of the code from above. Should the pools JSON composition logic be moved to a function that's called from both of these?

src/handler.cpp
```diff
@@ -1488,6 +1491,8 @@ void char_to_room(struct char_data * ch, struct room_data *room)
   _char_with_spell_to_room(ch, SPELL_SHADOW, ch->in_room->shadow);
   _char_with_spell_to_room(ch, SPELL_LIGHT, ch->in_room->light);
   _char_with_spell_to_room(ch, SPELL_POLTERGEIST, ch->in_room->poltergeist);
+  
+  SendGMCPRoomInfo(ch, ch->in_room);
```
luciensadi, line 1495: This works, but folks are going to see odd results. We sometimes change character room locations during processing of other code before restoring them to where they were. This is invisible to players, but will become visible with this change.

luciensadi (commented):
Thanks, I've been interested in adding GMCP support for a while!

src/handler.cpp
biscuitWizard: Maybe we should only add it on the do_move code? Hmm

src/comm.cpp
```diff
@@ -934,6 +935,20 @@ void game_loop(int mother_desc)
...
     }
 
+    {
+      // Send GMCP Vitals
+      for (d = descriptor_list; d; d = next_d) {
+        next_d = d->next;
+        if (!d->prompt_mode) continue;
+        if (!d->character) continue;
+
+        if (d->character->in_room) {
```
luciensadi, line 945: Why only if they're in a room? Are you invoking room-specific things in vitals/pools?

src/gmcp.cpp
```diff
@@ -0,0 +1,404 @@
...
+  SendGMCP(ch->desc, "Char.Pools", payload.c_str());
+}
+
+
+void ExecuteGMCPMessage(descriptor_t *apDescriptor, const char *module, const json &payload) {
+  if (!strncmp(module, "Core.Supports.Get", strlen(module)))
+    SendGMCPCoreSupports(apDescriptor);
+  else if (!strncmp(module, "Room.Info.Get", strlen(module)))
+    if (!apDescriptor->character)
+      mudlog_vfprintf(NULL, LOG_SYSLOG, "Unable to Handle GMCP Room.Info.Get Call: No Valid Character");
```
luciensadi, line 337: All of these calls should be `apDescriptor->character` instead of NULL. The code knows how to handle null character pointers.

src/gmcp.cpp
```diff
@@ -0,0 +1,404 @@
...
+    std::string payload = j.dump();
+    SendGMCP(ch->desc, "Matrix.Info", payload.c_str());
+    return;
+  }
+
+  struct host_data *cur_host = &matrix[persona->in_host];
+  j["name"] = cur_host->name;
+  j["description"] = cur_host->desc;
+  j["vnum"] = cur_host->vnum;
+  j["color"] = cur_host->color;
```
luciensadi, line 80: I think you usually have to run a program to see this, which gives a risk of being spotted by IC.

src/gmcp.cpp
```diff
@@ -0,0 +1,404 @@
...
+void SendGMCPRoomInfo( struct char_data *ch, struct room_data *room ) 
+{
+  if (!ch || !ch->desc || !ch->desc->pProtocol->bGMCP) return;
+  json j;
+
+  j["vnum"] = GET_ROOM_VNUM(room);
+  j["name"] = GET_ROOM_NAME(room);
+
+  // Zone info
+  struct zone_data *z = &zone_table[room->zone];
```
luciensadi, line 210: Zone info is not available to players; let's not send this along. The names are semi-random like "Warboss's third zone", etc.

PR #794, opened 2025-03-20
when victim dies/morted/stunned, stop victim driving instead of attacker

jdevnull:
This PR fixes an error where the attacker would be forced to stop driving instead of the victim, when the victim becomes morted, stunned, or dead from raw_damage.

Answers: https://discord.com/channels/564618629467996170/788953927269613608/1351792093227651082


luciensadi:
Good catch, thanks!

PR #796, opened 2025-03-20
Added radio program

biscuitWizard:
It is now possible to use the radio from the Matrix.

* Moved radio link from NERPS to Mechanical. 
* Implemented radio link program with multiplier one. 
* Adapted radio code to enable matrix use.
* Max Encryption Rating is based on Program Level
* It is not possible to use radio link software with a rating exceeding the hardware
* Otaku can use the radio in the matrix if they have a cyber radio and the complex form


src/utils.hpp
```diff
@@ -1049,6 +1049,8 @@ bool CAN_SEE_ROOM_SPECIFIED(struct char_data *subj, struct char_data *obj, struc
 #define GET_PART_PART_COST(part)               (GET_OBJ_VAL((part), 8))
 #define GET_PART_CHIP_COST(part)               (GET_OBJ_VAL((part), 9))
 #define GET_PART_INITIAL_BUILD_TICKS(part)     (GET_OBJ_VAL((part), 10))
+#define GET_PART_RADIO_FREQ(part)              (GET_OBJ_VAL((part), 11)) // Settable by player
+#define GET_PART_RADIO_CRYPT(part)             (GET_OBJ_VAL((part), 12)) // Settable by player
```
luciensadi, line 1053: Crypt is obsolete and shouldn't be used unless `ENABLE_RADIO_CRYPT` is defined.

src/act.comm.cpp
```diff
@@ -690,39 +690,102 @@ ACMD(do_page)
...
+
+ACMD(do_radio)
+{
+  struct obj_data *obj, *radio = NULL;
+  char one[MAX_INPUT_LENGTH], two[MAX_INPUT_LENGTH];
+  int i;
+  bool cyberware = FALSE, vehicle = FALSE, matrix = FALSE;
+
+  radio = find_radio(ch, &cyberware, &vehicle, &matrix);
+  if (PLR_FLAGGED(ch, PLR_MATRIX) && !matrix) {
```
luciensadi, line 785: This will block otaku from using their cyberware radios.

src/act.comm.cpp
```diff
@@ -1128,10 +1170,12 @@ ACMD(do_broadcast)
           else
             strlcat(radio_string, "^n", sizeof(radio_string));
 
-          store_message_to_history(d, COMM_CHANNEL_RADIO, act(radio_string, FALSE, ch, 0, d->character, TO_VICT));
+          store_message_to_history(d, COMM_CHANNEL_RADIO, act(radio_string, FALSE, ch, 0, d->character, 
+            d->character->persona && d->character->persona->decker && d->character->persona->decker->deck ? TO_DECK : TO_VICT));
```
luciensadi, line 1174: You're using a much more involved check here (persona, decker, and deck) vs above (just persona). Which do you want to standardize on?

biscuitWizard: persona

PR #797, opened 2025-03-23
Add cha mods to otaku masking, cha+wil mods to otaku mpcp

jdevnull:
This PR updates otaku masking calculation to include charisma modifiers (wil mods were already included), and also otaku MPCP calculation to include cha and wil mods (int mods were already included).

### DIFF
--- a/src/otaku.hpp
+++ b/src/otaku.hpp
@@ -7,6 +7,7 @@ struct otaku_echo {
   bool nerps;
 };
 
+int     get_otaku_cha(struct char_data *ch);
 int     get_otaku_wil(struct char_data *ch);
 int     get_otaku_int(struct char_data *ch);
 int     get_otaku_qui(struct char_data *ch);
@@ -24,7 +25,7 @@ extern int complex_form_programs[COMPLEX_FORM_TYPES];
 
 #define GET_OTAKU_MPCP(ch)           \
   (({ \
-    int mpcp = (get_otaku_int(ch) + GET_REAL_WIL(ch) + GET_REAL_CHA(ch) + 2) / 3; \
+    int mpcp = (get_otaku_int(ch) + get_otaku_wil(ch) + get_otaku_cha(ch) + 2) / 3; \
     if (GET_ECHO(ch, ECHO_IMPROVED_MPCP)) { \
       mpcp = MIN(get_otaku_int(ch) * 2, mpcp + GET_ECHO(ch, ECHO_IMPROVED_MPCP)); \
     } \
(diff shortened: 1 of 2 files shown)


src/otaku.hpp
```diff
@@ -24,7 +25,7 @@ extern int complex_form_programs[COMPLEX_FORM_TYPES];
 
 #define GET_OTAKU_MPCP(ch)           \
   (({ \
-    int mpcp = (get_otaku_int(ch) + GET_REAL_WIL(ch) + GET_REAL_CHA(ch) + 2) / 3; \
+    int mpcp = (get_otaku_int(ch) + get_otaku_wil(ch) + get_otaku_cha(ch) + 2) / 3; \
```
luciensadi, line 28: These are all very heavily drug-based functions. What stops someone from slamming every drug available, connecting to get a high MPCP, and then having detox or something cast on them to alleviate the penalties of comedown? Does their MPCP and program ratings need to update on a regular tick?

jdevnull: If I understand everything correctly, higher otaku MPCP only allows the use of higher rating complex forms (sleaze, attack, and defense), but doesn't affect channels (which directly modify the TNs for system tests to actually do anything in the matrix). The character gains +1 MPCP for every +3 mental attribute points. Masking is mainly relevant for detection factor to make it more difficult for the host to resist system tests. Because of the averaging, it's +1 DF for every +2.4 points (from the masking side, it's +1 DF for every +4 points, and from the sleaze side, it's +1 DF for every +6 points).

Living persona attributes are also capped at MPCP x 1.5, but I don't think this will come up much at all (would basically require otaku to submerge multiple times and make bad echo choices).

The applicable drugs are: kamikaze, nitro, zen (for wil), psyche (for int), and novacoke (for wil). I don't believe detox affects damage from drugs, only modifiers, so using nitro will probably get the otaku killed the moment it comes down. Zen's onset gives +1 TN to actions with its +1 wil, so is almost always a net negative.

The remaining 3 (kamikaze, psyche, and novacoke) combine for a total of +3 mental attributes (thus +1 MPCP and +1 DF), so I think we can reasonably stick with that when considering balance.

Psyche doesn't have any comedown effects but is more expensive (something like 500 nuyen per dose after max negotiation rolls). Novacoke's addiction tests are difficult (second only to Nitro) and has some very serious comedown effects (sets cha to 1 and cuts willpower in half). Kamikaze is strong overall, especially since it also provides 4 pain resistance), but has much shorter durations than the other two. Whether or not they use detox, I would say the combined benefits are not that strong, but hopefully enough that some fraction of otaku players will spend the nuyen for them.

As coded, `apply_drug_modifiers_to_ch` and `update_otaku_deck` are called by `affect_total`.

luciensadi: Thanks for the analysis!

PR #798, opened 2025-03-29
Various otaku/matrix things

jdevnull:
1) When connected with a living persona, use proxy deck values when showing storage memory in prompts.
https://discord.com/channels/564618629467996170/788953927269613608/1354737619493191741
https://discord.com/channels/564618629467996170/788953927269613608/1355008663164682290

2) Prevent living persona updates from overriding room I/O limits. Note that the way this was fixed means that I/O won't get updated if the otaku loses int, but I think typical room limits are low enough that it won't matter.
https://discord.com/channels/564618629467996170/788953927269613608/1355285007186530478

3) Add hints for decrypting SANs and learning complex forms.
https://discord.com/channels/564618629467996170/1219763142628999198/1354663103731859553
https://discord.com/channels/564618629467996170/788953927269613608/1354591108218491041

4) Fix matrix initiative: `GET_CYBERDECK_RESPONSE_INCREASE` is the otaku's matrix reaction, not response increase
https://discord.com/channels/564618629467996170/788953927269613608/1355299468643270808

5) Add otaku path to score.
https://discord.com/channels/564618629467996170/788953927269613608/1354095217044750357

### DIFF
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -555,12 +555,12 @@ bool do_damage_persona(struct matrix_icon *targ, int dmg)
 {
   if (targ->type == ICON_LIVING_PERSONA) {
     // It's an otaku! They get to suffer MENTAL DAMAGE!
-    // targ->condition seems to be 1-10 scale, while ch mental wounds seems to be 1-100. Multiply by ten.
-
     // damage() returns TRUE if the target has been deleted from memory, so when this function returns true, we must bail out of everything immediately.
     if (damage(targ->decker->ch, targ->decker->ch, dmg, TYPE_BLACKIC, MENTAL))
       return TRUE;
 
+    // targ->condition is 0-10 scale, while ch mental wounds is 0-100.
+    targ->condition = GET_MENTAL(targ->decker->ch) / 10;
     return FALSE;
   }
   targ->condition -= dmg;
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -2918,8 +2918,8 @@ ACMD(do_decrypt)
     // If there's nothing to decrypt, there's nothing to succeed at.
     if (!matrix[PERSONA->in_host].stats[mode][MTX_STAT_ENCRYPTED]) {
       send_to_icon(PERSONA, "The local %s subsystem doesn't seem to be encrypted.\r\n", mtx_subsystem_names[mode]);
-      if (mode == ACIFS_ACCESS && PRF_FLAGGED(ch, PRF_SEE_TIPS))
-        send_to_icon(PERSONA, "[OOC: To decrypt a remote host's SAN, try \"decrypt <host>\".]\r\n");
+      if (PRF_FLAGGED(ch, PRF_SEE_TIPS) && (mode == ACIFS_ACCESS))
+        send_to_icon(PERSONA, "[OOC: To decrypt a remote host's SAN, try ^WDECRYPT <HOST>^n.]\r\n");
       return;
     }
 
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -2917,7 +2917,9 @@ ACMD(do_decrypt)
 
     // If there's nothing to decrypt, there's nothing to succeed at.
     if (!matrix[PERSONA->in_host].stats[mode][MTX_STAT_ENCRYPTED]) {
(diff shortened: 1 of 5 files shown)


luciensadi:
FML, I tried to make it merge to a branch so I could keep a snapshot of what I've reviewed and allow you to do more on it without adding it to the game base, and it dropped the majority of the changes from the PR :/

src/pro_create.cpp
```diff
@@ -243,6 +243,9 @@ struct obj_data *can_program(struct char_data *ch)
       comp = NULL;
     if (!comp) {
       send_to_char(ch, "There is no computer here for you to use.\r\n");
+      // A computer isn't used to design a complex form, but we use a different command for that
+      if (PRF_FLAGGED(ch, PRF_SEE_TIPS) && IS_OTAKU(ch))
+        send_to_char("[OOC: To learn a complex form, try ^WFORMS LEARN <FORM>^n.]\r\n", ch);
```
luciensadi, line 248: Does this check need to be outside of the `!comp` check then?

src/otaku.cpp
```diff
@@ -240,7 +240,6 @@ void _update_living_persona(struct char_data *ch, struct obj_data *cyberdeck, in
   ch->persona->decker->mpcp = mpcp;
   ch->persona->decker->hardening = GET_CYBERDECK_HARDENING(cyberdeck);
   ch->persona->decker->response = GET_CYBERDECK_RESPONSE_INCREASE(cyberdeck);
-  ch->persona->decker->io = GET_CYBERDECK_IO_RATING(cyberdeck);
```
luciensadi, line 243: This should probably be set to _something_. What's the original value?

src/newmatrix.cpp
```diff
@@ -242,19 +242,21 @@ void roll_matrix_init(struct matrix_icon *icon)
...
   {
-    // Matrix pg 18 & 24, available bonuses are response increase, reality filter, and hot asist
-    init_dice += GET_INIT_DICE(icon->decker->ch) + icon->decker->response + (icon->decker->reality ? 1 : 0) + (icon->decker->asist[0] ? 1 : 0);
-
     if (icon->type == ICON_LIVING_PERSONA) {
-      init_dice = MIN(5, init_dice + GET_ECHO(icon->decker->ch, ECHO_OVERCLOCK));
+      // Matrix pg 138 & 145, 4 dice with a possible +1 from overclock
+      init_dice = 4 + (GET_ECHO(icon->decker->ch, ECHO_OVERCLOCK) ? 1 : 0);
+      // This is the living persona's matrix reaction, not response increase
+      icon->initiative = icon->decker->response;
```
luciensadi, line 249: Response is a flat value here, but doubled for non-otaku. Is that intended?

src/pro_create.cpp
jdevnull: It needs to be inside an error condition, and the ones I think an otaku would reach when attempting "design <form>" is either the absence of a computer here, or the absence of <form> in a present computer.

src/otaku.cpp
jdevnull: This value is set in `do_connect` (newmatrix.cpp), where it is also capped by the room they're connecting from. The problem was that every time `update_otaku_deck` here is called, it overrides that value and bypasses that limit. The math here is also wrong (it should actually be handled as /10), so otaku are currently downloading files somewhere between 10 and 30 times faster than intended.

This fix is not entirely correct because a change in int (e.g., due to drugs) can cause the living persona's I/O rating to change, but the room limit will pretty much always be the limiting factor. A more correct fix duplicates the code from do_connect.

src/newmatrix.cpp
jdevnull: Yes, this is intended. For a deck, each level of response gives +2 reaction and +1 init die, and is limited to 0-3. The code that sets `GET_CYBERDECK_RESPONSE_INCREASE` (which then gets passed to `DECKER->response` in `do_connect`) in `update_living_persona` is actually calculating the otaku's matrix reaction, not level of response increase. Otaku don't actually get response increase levels, so I'm letting the two different things use the same struct member.

For clarity, should I get rid of the `GET_CYBERDECK_RESPONSE_INCREASE` calculation in `update_living_persona` and instead calculate the otaku's matrix reaction here?

luciensadi: It'd probably be clearer, but I'm okay with it as-is. Thanks!

luciensadi:
Thanks for the code!

PR #802, opened 2025-04-06
More matrix fixes

jdevnull:
1. `decker->active` is the amount of free memory, so unloading a program should add to this value, not subtract.

2. Fix rolls output for system tests so that the displayed starting TN is actually the starting TN, and to remove the duplicate "called modifier" display when the modifier is not zero (and removed entirely when it is zero).
Addresses: https://discord.com/channels/564618629467996170/788953927269613608/1355447999177953402

3. Fix missing dumpshock damage.
Addresses: https://discord.com/channels/564618629467996170/788953927269613608/1358548420708794428

### DIFF
--- a/src/newmatrix.cpp
+++ b/src/newmatrix.cpp
@@ -339,9 +339,11 @@ bool dumpshock(struct matrix_icon *icon)
 {
   if (!icon) return FALSE;
 
-  if (icon->decker && icon->decker->ch)
+  struct char_data *ch = icon->decker->ch;
+
+  if (icon->decker && ch)
   {
-    send_to_char(icon->decker->ch, "You are dumped from the matrix!\r\n");
+    send_to_char(ch, "You are dumped from the matrix!\r\n");
     snprintf(buf, sizeof(buf), "%s^n depixelates and vanishes from the host.\r\n", CAP(icon->name));
     send_to_host(icon->in_host, buf, icon, FALSE);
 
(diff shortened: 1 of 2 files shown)


src/newmatrix.cpp
```diff
@@ -476,13 +475,12 @@ int system_test(rnum_t host, struct char_data *ch, int type, int software, int m
       }
     }
   }
-  snprintf(rollbuf, sizeof(rollbuf), "System test against %s with software %s: Starting TN %d", acifs_strings[type], programs[software].name, target);
 
   int skill = get_skill(ch, SKILL_COMPUTER, target) + MIN(GET_MAX_HACKING(ch), GET_REM_HACKING(ch));
   GET_REM_HACKING(ch) -= skill - get_skill(ch, SKILL_COMPUTER, detect);
   detect = 0;
 
-  snprintf(ENDOF(rollbuf), sizeof(rollbuf) - strlen(rollbuf), ", after get_skill %d, plus called modifier %d is %d", target, modifier, target + modifier);
```
luciensadi, line 485: Where is this modifier coming from?

jdevnull: It looks like it's used in `do_restrict` (which doesn't apply against IC, so appears to be NERP outside of decker PVP). I removed it here because it also gets added to rollbuf some 3 lines later under `if (modifier) ...`

jdevnull:
4. Psyche provides +1 int to Otaku stats, not +2.

luciensadi:
Looks good to me, thank you!

PR #806, opened 2025-09-20
Fixes to socials and load_single_veh (also credit finaltwist)

jdevnull:
Fixes resulting from discussion with FinalTwist (details start at: https://discord.com/channels/564618629467996170/564621949842751508/1418725432945021049 )

1. don't continue reading from socials file if there's no room remaining in the allocated array
2. add comment documenting socials file format to `boot_social_messages`
3. fix socials: scratch, swear, cheer
4. rename socials: crack, flip, roll, think to em_crack, em_flip, em_roll, em_think 
5. when attempting `load_single_veh` but the file doesn't exist, return immediately

### DIFF
--- a/lib/misc/socials
+++ b/lib/misc/socials
@@ -183,7 +183,7 @@ You collapse in a heap.
 $n collapses in a heap.
 #
 
-crack 0 0
+em_crack 0 0
 You crack your knuckles.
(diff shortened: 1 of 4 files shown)


luciensadi:
LGTM, thanks!

PR #838, opened 2026-03-02
add kinesics emotes for social adepts

lelagh:
needs a helpfile update, I don't know how those are done...

* two new_echo sub commands: KEMOTE and KECHO (they're identical, just a different way of typing it)
* only works if you are an adept with your kinesics currently activated, otherwise, it just tells you why you can't do it
* speech in quotation marks is not permitted and must_echo_with_name is enforced as it's all body language in TT
* only adepts with kinesics currently active will see it

### DIFF
--- a/src/interpreter.cpp
+++ b/src/interpreter.cpp
@@ -778,6 +778,8 @@ struct command_info cmd_info[] =
     { "jobs"       , POS_DEAD    , do_recap    , 0, 0, BLOCKS_IDLE_REWARD },
     { "junk"       , POS_RESTING , do_drop     , 0, SCMD_JUNK, BLOCKS_IDLE_REWARD },
 
+    { "kecho"      , POS_LYING   , do_new_echo , 0, SCMD_KEMOTE, BLOCKS_IDLE_REWARD },
+    { "kemote"     , POS_LYING   , do_new_echo , 0, SCMD_KEMOTE, BLOCKS_IDLE_REWARD },
     // { "kil"        , POS_FIGHTING, do_kil      , 0, 0, BLOCKS_IDLE_REWARD },
     { "kill"       , POS_FIGHTING, do_kill     , 0, SCMD_KILL, BLOCKS_IDLE_REWARD },
     { "keep"       , POS_LYING   , do_keep     , 0, 0, BLOCKS_IDLE_REWARD },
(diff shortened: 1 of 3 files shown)


luciensadi (changes requested):
This is an interesting thought! I have a few requests:
- Put a length limit on it, maybe 80 characters max. This is only good for "basic messages and emotional states", but folks may not realize that and start writing War and Peace with their kemotes.
- Add a tag to it, like `(kinesics) Lucien flexes awkwardly, and you understand he's saying he's hungry`, so people know which mode they should respond with. I'm flexible on how and where it's tagged as long as it's noticeable.
- Update the visibility check so staff can always see it, and questers/RPE can always see it as long as it's in a NERPcorpolis apartment.

scoutyoshimoto:
alrighty, all of those are in and tested on my janky server:

* 80 character limit
* prepends (Kinesics) at the beginning of the line
* staff will see them always
* questors see if the adept has the hired flag

and for the help file with KEMOTE and KECHO as aliases for KINESICS

```
Adepts within sight of each other can use KEMOTE or KECHO to non-verbally communicate basic messages and emotional states to each other using body language alone. Functionally, these are short EMOTEs or ECHOs that can only be used or seen by other adepts with their kinesics power activated.
```

src/interpreter.cpp
```diff
@@ -778,6 +778,8 @@ struct command_info cmd_info[] =
     { "jobs"       , POS_DEAD    , do_recap    , 0, 0, BLOCKS_IDLE_REWARD },
     { "junk"       , POS_RESTING , do_drop     , 0, SCMD_JUNK, BLOCKS_IDLE_REWARD },
 
+    { "kecho"      , POS_LYING   , do_new_echo , 0, SCMD_KEMOTE, BLOCKS_IDLE_REWARD },
+    { "kemote"     , POS_LYING   , do_new_echo , 0, SCMD_KEMOTE, BLOCKS_IDLE_REWARD },
```
luciensadi, line 782: If `kecho` and `kemote` are the same command, it's better to define one and then have the other be an alias (near the bottom of the file). Otherwise both show up in the `commands` list and cause confusion for players.

luciensadi (approved):
Only one change to request, and I can do that on my end. Thanks for the code!

PR #841, opened 2026-03-11
Proof of Concept: Scope hinting via prefixes

max-buzzkill:
Currently `const_generic_find` is used for a number operations, and numeric prefixes allow differentiating items on multiple keyword matches.  However, these numeric indices are calculated per-scope via `get_number`. This leads to issues if someone is for example 2 items that match a keyword (or the same item like a ring) and is trying to refer to another match/copy in their inventory: `1.keyword` ends up matching the first instance in their equipment, while `3.keyword` fails because there is no 3rd item in equipment, and the inventory scope is indexing that match at `1.keyword`. While this is mostly an inconvenience, there are cases, such as sustaining foci supporting a ritual-cast spell where there is an associated cost to removing/swapping items.

Proposing  adding scope hinting prefixes: `worn.`, `inv.`, `room.` which narrow the scope bitmap ans therefore search space to the relevant scope, if it is valid for that command. The above scenario can be resolved by using `inv.keyword` or `inv.1.keyword` to match the item in inventory. 

Additionally, experimenting with adding a logical scope: `exit.` to force exist/door interactions in a similar fashion, allowing the use of keywords beyond the cardinal directions. This might need some additional input/be expanded to other commands if it is thought to be useful, hoping for some feedback on this.

### DIFF
--- a/src/act.movement.cpp
+++ b/src/act.movement.cpp
@@ -1512,8 +1512,11 @@ ACMD(do_gen_door)
   two_arguments(argument, type, dir);
   RIG_VEH(ch, veh);
 
+  // Scope hint: "exit." prefix forces door/exit resolution, skipping object and vehicle searches.
+  bool force_exit_search = strip_exit_scope_prefix(type);
+
   // 'unlock 1' etc for subscribed vehs
-  if (*type && is_number(type)) {
+  if (!force_exit_search && *type && is_number(type)) {
     num = atoi(type);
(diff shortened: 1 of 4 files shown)


max-buzzkill:
Closing this for now, likely not worth considering until there's a unified item resolution strategy (and all the bespoke item finds are eliminated, otherwise this is likely to just cause confusion)

PR #847, opened 2026-04-02
hide concealed foci from cameras/other minor tweaks

lelagh:
as if there invisible to everyone except staff no matter what, a nikon shouldn't see them either

reword the text in COUNT to clarify that mails are fine now

### DIFF
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -3272,6 +3272,11 @@ ACMD(do_photo)
               continue;
             }
 
+            // Conceealed foci should not show in photos
+            if (GET_OBJ_TYPE(eq) == ITEM_FOCUS && IS_OBJ_STAT(eq, ITEM_EXTRA_CONCEALED_IN_EQ)) {
+              continue;
+            }
+
             snprintf(ENDOF(buf), sizeof(buf) - strlen(buf), "%s%s\r\n", where[j], GET_OBJ_NAME(GET_EQ(found_ch, j)));
           }
         found = TRUE;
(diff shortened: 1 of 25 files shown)


src/act.other.cpp
```diff
@@ -3272,6 +3272,11 @@ ACMD(do_photo)
               continue;
             }
 
+            // Conceealed foci should not show in photos
+            if (GET_OBJ_TYPE(eq) == ITEM_FOCUS && IS_OBJ_STAT(eq, ITEM_EXTRA_CONCEALED_IN_EQ)) {
```
luciensadi: is `eq` defined here? Prior code uses `GET_EQ(found_ch, j)`

lelagh: it totally does >.>

src/act.other.cpp
```diff
@@ -3276,6 +3276,11 @@ ACMD(do_photo)
               continue;
             }
 
+            // Conceealed foci should not show in photos
+            if (GET_OBJ_TYPE(found_eq) == ITEM_FOCUS && IS_OBJ_STAT(found_eq, ITEM_EXTRA_CONCEALED_IN_EQ)) {
```
luciensadi: Does this compile? I'm still having a hard time finding a `found_eq` ref in the preceding code. It looks like it needs to be `GET_EQ(found_ch, j)` instead of `found_eq`.

lelagh: it was compiling for some reason? but yeah nah, it should be

PR #854, opened 2026-05-21
typos and line breaks

lelagh:
* flipped the ternary on the CHECK orders list, cause "2 day" and "1 days" was the wrong way around
* fixed a typo in one of the luxury lifestyle strings
* added some line breaks to some tips
* and added some line breaks to some (OOC: ...) lines


luciensadi (approved):
LGTM, thanks!

PR #859, opened 2026-06-01
magic errata/typos

lelagh:
fix for some traffic message typos

fix for a liquid color typo

fix for a non-capitalized look string when you trauma patch someone

add a typo guard for one of the easily misunderstood Sperethiel words

+ fix for the pocket secretary misleading you about mails and item counts.

and in line with the two books:

* gator was missing the detection bonus

* rat's and jackal's -1 combat was in the wrong place, so not applied

* fish had a bonus to all spirits of waters, and then also one double bonus to one selected spirit on top

* moon had +2 detection, should be +1, and was missing the typical spread of "spirits of the waters"

* moon/sea was hit by the siren penalty, and siren had the wrong manipulation bonus (transformation, not control)

* adepts now allowed to learn the nerps divining metamagic as per TT 

* sun/leopard shamans are not supposed to have time based bonuses, but are still allowed as they don't ONLY have time based bonuses like owl

bonus notes:

* goose is noted as a combat penalty in the HELP file, should be a +1 bonus by the TT rules
* siren/seductress/lover and moon/sea just say +2 manipulation in the HELP files, should probably note it's control/transformation respectively as those are coded restrictions


luciensadi:
Thanks for the patch!

PR #861, opened 2026-06-09
silent mode for docwagon receivers

lelagh:
this one's just a "silent mode" to not send the alert echos to everyone in your room to cut down on to_room noise when several people are wearing one

### DIFF
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -1234,7 +1234,9 @@ const char *tog_messages[][2] = {
                             {"You will now automatically ready your holstered/sheathed weapons.\r\n",
                              "You will no longer automatically ready your holstered/sheathed weapons.\r\n"},
                             {"You will now see roundtime messages for your actions.\r\n",
-                             "You will no longer see roundtime messages for your actions.\r\n"}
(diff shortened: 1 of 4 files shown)


luciensadi (changes requested):
It looks like the initial ping when someone goes down is not included in the patch (e.g. playerdoc.cpp line 122 and others), so you'll still get the shrill beeps etc.

Also, I think the effect of this is reversed from how things usually run-- if a player does not want to see DW pings from others, that should be a toggle they set themselves, e.g. `toggle docpings` or something, which sets the appropriate PRF_ bit on the toggler. This way folks who are bothered by it can tune it out, and others who like it or don't care are able to keep listening for them. This prevents a situation where a player who doesn't enjoy the pings pressures another player into disabling them.

lelagh:
ohh good call actually, i'll flip it around :)

lelagh:
alright there we go shrill beeps covered and now it's a toggle on you (docs can also toggle it to see their own pings but not any bonus room pings too)

lelagh:
whoops I don't know how I did that

PR #862, opened 2026-06-09
apply cyber penalties/bonuses to fleeing

lelagh:
this one just some updates to bring the flee function in line with the fight code: the fight code adds the jack bonus before halving from foot anchors, so that order is here, plus the kid stealth bonus is a separate TN modifier, so that's handled separately

### DIFF
--- a/src/act.offensive.cpp
+++ b/src/act.offensive.cpp
@@ -631,6 +631,10 @@ struct char_data *find_a_character_that_blocks_fleeing_for_ch(struct char_data *
       break;
       }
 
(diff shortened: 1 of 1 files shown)


luciensadi:
I'm hesitant about this one, as it adds penalties and changes how fleeing works without a game design justification. The flee code is houseruled already to make it easier to flee (avoids wasting player time on roundtime); is there a need to further differentiate between folks with these various types of 'ware?

lelagh:
this mainly came out of a discussion about foot anchors and probably a bit skippable anyway

PR #863, opened 2026-06-13
Clarifies Radio Link / Interface info description

Naxodile:
Clarifies the text in `create part/program` to specify Radio Link is a required program for functionality, not an installation software. (Verified at: [act.comm.cpp:1068](https://github.com/luciensadi/AwakeMUD/blob/219de60026c6811d906cc103370fcc33f1d400cf/src/act.comm.cpp#L1073))


luciensadi:
Thanks for the code!

PR #867, opened 2026-06-26
Fix cpool allocation behavior

jdevnull:
Previously, unallocated cpool dice would claim to be allocated to soak but end up in offense and/or dodge instead. This is because the `affect_total` cpool logic was different than the `do_cpool` logic.

In addition to fixing the bug, this PR does the following:
1. Refactored so that do_cpool and affect_total both call set_combat_pools.
2. Unallocated cpool dice will be allocated to soak or dodge depending on where allocated dice already exist. If you have already allocated more dice to dodge than soak, then your unallocated dice will be added to dodge, otherwise it will be added to soak.
3. You can attempt to over-allocate cpool and that setting will be saved (though actual pools will be capped as normal). This allows your cpool to automatically adapt to changes while following the player's priorities (e.g., different weapon skills changing offense cap, temporary stat buffs changing cpool totals, etc).
4. Removed adding dodge to astral - dodge is not used in astral combat.

Example: assume 10 total cpool, with 2 ranks of brawling and 6 ranks of pistols.
`cpool 0 0 0` -> all 10 unallocated dice default to soak
`cpool 1 0 0` -> the 9 unallocated dice are added to the 1 that's already allocated to dodge
`cpool 0 0 15` -> with brawling, offense is capped at 2 due to skill, leaving 8 in soak. Draw the pistol, offense automatically changes to cap 6 and soak to 4.
`cpool 0 10 5` -> all 10 cpool go in soak, with 0 remaining for offense. Add a combat sense spell and those dice will go into offense first because of the saved over-allocation (up to the lower of the over-allocated 5, or the offense cap at weapon skill), and then any remaining dice are added to soak.

This PR addresses:
https://discord.com/channels/564618629467996170/788953927269613608/1518520055527374931
https://discord.com/channels/564618629467996170/788953927269613608/1475395618393423912
https://discord.com/channels/564618629467996170/788953927269613608/1376382525911466176
(and others)

- Khai

### DIFF
--- a/src/act.other.cpp
+++ b/src/act.other.cpp
@@ -4598,7 +4598,7 @@ ACMD(do_cpool)
   extern int get_skill_num_in_use_for_weapons(struct char_data *ch);
   extern int get_skill_dice_in_use_for_weapons(struct char_data *ch);
 
-  int dodge = 0, bod = 0, off = 0, total = GET_COMBAT_POOL(ch);
+  int dodge = 0, bod = 0, off = 0;
 
   if (!*argument) {
     do_pool(ch, argument, 0, 0);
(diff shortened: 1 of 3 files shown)


jdevnull:
Also, the code that wanted to add dice to dodge for astral was removed because dodge is not used in astral combat. No bug reports for it though.

luciensadi (approved):
Thanks for the patch!

PR #868, opened 2026-07-27
fix radio forms and matrix comma broadcasts

lelagh:
(did not realise this check would fail)

1: ~~broadcast as an otaku would fail on checking for the Radio Link as `do_radio` has a sub check for the complex form/cyber radio, but `do_broadcast` does not, so do the same check here~~ the test didn't exactly work as it should have, it just got skipped and then passed or failed on checking for any kind of radio at all at the end of the prechecks

2: also add the comma alias as a broadcast option in the matrix


luciensadi (approved):

luciensadi:
Thanks for the patch!

PR #869, opened 2026-09-05
Port DG Scripts from tbaMUD, with trigedit

Fiztba:
## What this is

This is the engine, the trigger checks, the `%variable%` substitution, the m/o/w command sets, the file format, the OLC integration, and `trigedit` / `tattach` / `tdetach` / `tstat` / `vdelete`.

`STAT` on a room, a mobile or an object also lists whatever triggers are attached to it, so a builder can ask what a thing runs without opening an editor on it.

Seventeen help entries come with it, in `SQL/dg_scripts_help.sql`, loaded by `gensql.sh` and by `docker/db/init-db.sh` alongside the other help SQL.

It's a faithful port rather than a reimagining, with the deliberate departures listed below. 

## File format

`.trg` files are VTable key/value, like the rest of Awake's world files, not Diku's positional format:

```
#60500
Name:	rift greeter
AttachType:	2
Types:	64
NArg:	100
Arg:
Commands:$
* the script goes here
wecho The rift shivers as it notices %actor.name%.
~
BREAK
END
```

They live in `lib/world/trg/` with an `index`, and are written by `trigedit` the same way `redit` writes rooms. `*` comment lines are indented by one space on the way out because `File::GetLine` drops a line starting with `*`; a literal `~` inside a script is written as `\7` the way `ReadString` expects. An empty index is allowed, so a game with no triggers still boots.

Attachments ride on a `Scripts:` field on the room, mobile or object — a space-separated list of trigger vnums. Deliberately not `[TRIGGERS]`, which `hedit` already uses for Matrix IC steps.

The zone `T` command works: `T <if_flag> <attach type> <trigger vnum> <room vnum>`. Its arg2 stays a vnum rather than being renumbered to an rnum, which sidesteps a long-standing tbaMUD bug where inserting a trigger silently repoints every zone command after it. `zedit` writes `T` lines back out but has no menu option for adding one, so a builder has to put it in the `.zon` by hand; the help entry says so. Adding that option is a functional change and not mine to make here.

## Shadowrun mapping

DG's variable names come from a D&D-shaped game, so they had to be pointed somewhere:

| DG field | Awake |
| --- | --- |
| `hitp` / `maxhitp` | physical condition monitor, in boxes |
| `mana` / `maxmana` | mental condition monitor, in boxes |
| `gold` | nuyen |
| `exp` | karma |
| `con` / `dex` / `wis` | `bod` / `qui` / `wil` (the Awake names also work) |
| `armor` | ballistic; impact is its own field |
| — | `mag`, `rea`, `ess` added |
| — | objects get `val0` through `val17` |

Dropped for want of anything to point at: `align`, `class`, `damroll`, `hitroll`, `prac`, `saving_*`, `stradd`, `is_killer`, `is_thief`, `drunk`, `hunger`, `thirst`, `move`/`maxmove`, `affect`, `questpoints`, `cost_per_day`.

## What isn't here, and why

- **`%cast%` and `%dg_affect%`** — Awake has no `call_magic(level, spellnum)` and no per-character timed-affect list to hang them on. Both would need real design decisions about how a script-cast spell interacts with Force, drain and the astral, and those are yours to make, not mine.
- **The zone `V` command** — `reset_com` has no string arguments, and `V` is already Vehicles.
- **`attach` and `detach`** are already player commands for weapon accessories, so the DG ones are `tattach` and `tdetach`. `rdelete` is `vdelete`, so nobody reads it as a way to delete rooms.
- **`struct event` is `struct dg_event`** — the bare name is too generic for this codebase. Every *function* keeps its tbaMUD name.
- **Room triggers can't see vehicle occupants**, who are on no room's people list. `%actor.room%` does answer with the enclosing room for someone in a vehicle rather than pointing at the void.
- **A redit save keeps a room's script variables**, where stock DG throws them away: `add_room()` calls `extract_script()`, which frees the triggers, the variables and the script together, and then rebuilds an empty one. That is a destructor being used to swap a trigger list. The trigger half is its own function here, so a save drops the triggers and keeps the rest - which matters most on the room holding the mud-wide globals, since it has no triggers of its own and losing its script leaves `%global%` with nowhere to be written. A save still ends any script the room is part-way through a `wait` on. Offered back upstream as tbamud/tbamud#356.

## Player script variables

`remote` and `rdelete` work on a player, so quest state can follow someone around, and those variables are saved with the pfile. A new `pfiles_scriptvars` table holds them; the schema is in `SQL/awakemud.sql` and in `SQL/Migrations/add_scriptvars.sql`, which `docker/db/init-db.sh` picks up on its own.

None of it reaches a player who never meets a script. The save hangs off the existing dirty-bit machinery, so the table is touched only for someone a script has actually written to, and reading variables back at login is explicitly not a change, so a player who loads them and touches no script leaves the rows exactly as they found them. I checked that by planting a value in the table by hand, playing a session that touched no script, and confirming it came back untouched.

Being pfile data, they need a ceiling: a `remote` inside a `while` could otherwise grow one player's save without limit. A player carries at most 128 variables. Overwriting one is always allowed - only a new name can lengthen the list - and refusing names the trigger, the player and the variable in the script log, so whoever wrote the loop can find it. Say the word if you would rather that number were different, or configurable.

`STAT` lists the script variables on a room, mobile, object or player. A quest that misbehaves is close to undiagnosable otherwise.

## Awake-specific hazards this had to deal with

- `extract_char` and `extract_obj` free immediately, and every trigger check holds exactly one saved `next` pointer while a script runs, so anything a script purges is freed out from under whichever list is being walked. `mpurge`, `opurge`, `wpurge` and `mjunk` all record the intent instead, and the heartbeat carries it out once per pulse. Until then the thing is still on every list it was on, so `SCRIPT_CHECK` and `valid_dg_target` both treat it as already gone. The general fix would be deferred extraction for the whole game, but that changes *when* things die, so DG keeps its own deferral and leaves the rest of the codebase alone.
- `do_mtransform` uses the `medit` struct-assignment plus `copy_over_necessary_info()` idiom rather than tbaMUD's `memcpy`, which is unsafe here because `char_data` owns `unordered_map`s.
- The six editor and clone commands each copied a prototype with a struct assignment and then hand-wrote the list of pointers that must not be shared - six lists to keep in step with `free_room()`, `free_obj()` and `free_char()`. There is one `clone_*_for_editing()` per type now, and the call sites keep only what they genuinely mean differently: which zone the copy belongs to, what is standing in it, and the identity a clone has to be issued fresh. What consolidating them turned up is the next section.
- `objList::UpdateObjs` and `copy_over_necessary_info()` preserve live scripts and script ids.
- `handler.hpp` declared `get_obj(char *)` and `get_obj_in_list(char *, obj_data *)` and nothing in the tree defined either, so the first caller outside the scripting code would have found out at link time. They are declared where they are defined now, and take a `const` name like every neighbouring lookup.
- `REQUIRE_ZONE_EDIT_ACCESS` moved from `olc.cpp` into `olc.hpp` so `trigedit` can use it, which exposed that only the `zone_data *` overload of `can_edit_zone` was declared there.

## Four OLC bugs found on the way past

Putting one clone function behind all six commands meant diffing the six hand-written lists against each other, and they did not agree. The fourth turned up while testing the room-variable change below. None of these are DG's; they are in OLC today.

**`rclone` shares strings with the room it copied.** It duplicated `name`, `description`, `address` and `flight_code` but not `night_desc`, and of an exit's five strings it duplicated two. Abandoning the clone runs `free_room()` on it, which deletes strings the source room still points at. I stashed the fix and reproduced it under AddressSanitizer rather than reason about it:

```
==4880==ERROR: AddressSanitizer: attempting double-free on 0x75943aeb7090
    #1 free_room(room_data*) db.cpp:5990          <- room->night_desc
    #2 DeleteRoom(room_data*) memory.cpp:161
    #3 free_editing_structs(descriptor_data*, int) comm.cpp:2535
    #4 close_socket(descriptor_data*) comm.cpp:2713
```

Three abandoned clones of a room that has a night description, and the third one double-frees. With the clone function in place the same sequence is clean.

**`free_room()` never freed `flight_code`.** Every room has had one since flight went in, redit hands the room a fresh copy on every save, and nothing ever released the old one. It does now, and redit releases the old string before overwriting it.

**`free_char()` compares a mobile's highlight colour against itself.** The guard is meant to skip strings a mobile is borrowing from its prototype:

```c
if (SETTABLE_CHAR_COLOR_HIGHLIGHT(ch) && SETTABLE_CHAR_COLOR_HIGHLIGHT(ch) != SETTABLE_CHAR_COLOR_HIGHLIGHT(ch))
```

The second operand should be `proto`. As written the condition can never be true, so a mobile that owns its own highlight string never gives it back.

**Saving the first room in the world puts a second copy of it in the world.** `real_room()` answers `NOWHERE` when there is nothing there and `0` when what it found is the first room, and redit's save tests that answer with `> 0`. Room 0 is where the mud-wide `%global%` variables live, so saving it left a second room with the same vnum shadowing the first, every rnum above the insertion point shifted by one, and `%global.<name>%` reading out of a room nothing had ever written to. `hedit` and `icedit` test `real_host()` and `real_ic()` the same way, against the same contract, and are fixed with it - though only redit's was reproduced. The same test appears in a dozen places outside OLC, which I have left alone.

Two things I deliberately did not touch, and would want your steer on. redit, medit and iedit save through three different mechanisms for the same job - `free_room()` then reassign, `copy_over_necessary_info()`, `objList::UpdateObjs()`. That asymmetry is real, but unifying it is a redesign rather than a safety pass. And a mobile's astral and matrix text stay shared with the prototype where its physical text is duplicated, because `read_mobile()` hands every instance the prototype's strings by pointer and that sharing is load-bearing.

## Help entries

`HELP TRIGGERS` is the way in, and indexes the rest: `TRIGEDIT`, `TATTACH`, `TSTAT`, one entry per attach type listing every trigger type with what its numeric argument means and which variables it starts with, the script language, the three field lists, the three command sets, `ZONE TRIGGERS` and `PLAYER VARIABLES`.

I wrote them out of this tree rather than carrying over the ones that ship elsewhere, because those describe a different game. Every trigger type's variable list came out of `dg_triggers.cpp`, every field out of `dg_variables.cpp`, every command's arity out of its own function. Nothing in them names a command, field or trigger type that does not exist here, and every `See Also` was resolved against the topic list, new and existing both.

## What was actually run

Built with g++ 15 and booted against a real database. Driven through a scripted telnet client, as an immortal and as a mortal:

- room enter, command and speech triggers, including `%arg%` and `if`/`else`/`end`
- a mobile greet trigger that speaks, `wait 2 sec`, echoes, then `mpurge self` - the wait event resumes and the deferred extraction lands without a crash
- object wear and remove triggers
- `remote` writing a counter onto a player, `%actor.tries%` reading it back, the value surviving a clean quit and a reboot, a script's `rdelete` and a staff `vdelete` both persisting, and a value planted in the table by hand surviving a session that touched no script
- `trigedit` creating a trigger, editing an existing one, copying one over another, and inserting the first trigger into an empty index
- `redit` attaching triggers, the `Scripts:` field appearing in the `.wld`, and the attachment surviving a reboot
- a zone `T` command attaching on reset, firing, and coming back out of the boot rewrite byte for byte
- `tattach`, `tdetach`, `tstat`, `stat` on a room, mobile, object and player
- every one of the seventeen help entries, read in game as a mortal, by each of its keywords, with the existing `ATTACH`, `TIME`, `SWITCH` and `COMMANDS` entries checked for shadowing
- staff exemption behaving exactly as stock DG does - `valid_dg_target` call sites match pristine tbaMUD one for one
- all six editor and clone commands, each opened and aborted, opened and saved, and had its link dropped mid-edit

All of it again under `-fsanitize=address,undefined`. That was clean as first written. Reading my own diff back afterwards was not, and the next section is what came out of it. UBSan is unchanged either way: two hits fire at boot, both signed overflow in matrix program cost maths in `db.cpp` from 2020, untouched here.

**Not exercised live:** the fight, hitprcnt, death, bribe, cast, time, timer, reset, login, act, door, leave and memory triggers; `drop`, `give`, `receive` and `consume` (Awake gates dropping and giving until a character is authorized, and I did not want to fake my way past that; what a refused drop does was worth knowing, and it fires nothing); `mtransform`; `mload` / `oload`; and `%purge%` on an object. They compile and follow the same paths as what was tested, but I would rather say so than imply otherwise.

Running it turned up seven bugs that compiling never would have. Four in this port: the editor copies aliasing the live scripts, an off-by-one on `top_of_trigt` that read past the trigger table once trigedit had inserted anything, `T` missing from the zone writer so a hand-written `T` command lasted exactly until the next reboot, and `trigedit` seeding an append-only string editor, which made every edit a duplication. Three already in OLC, described above. Each is its own commit with the reasoning.

## Reading it back

I went over this branch afterwards as though someone else had written it. Nine commits are what that turned up, each with its own reasoning; the two after them came out of a question about one of those findings, and are in the OLC and departures sections above. **Three of the findings were AddressSanitizer reports**, and the reproductions are the regression tests now:

- **A greet trigger doing `mpurge <the other mob in the room>`** freed that mob out from under the walk `greet_mtrigger` was in the middle of - `heap-use-after-free` at `dg_triggers.cpp:257`. The deferral was drawn around the script's own owner; it has to be drawn around everything a script destroys, because every trigger check holds exactly one saved `next` pointer. And something purged must not keep acting for the rest of the pulse it spends waiting on the lists, so `SCRIPT_CHECK` refuses it - tbaMUD says the same thing with `MOB_NOTDEADYET`.
- **`stat` on a script-loaded object read one past the end of `obj_load_reasons[]`** - `global-buffer-overflow`. The defines skip a number, so every name past that point has been off by one since load tracking went in; nothing had reached the end of the array until scripts got a load reason of their own.
- **A script variable name longer than the column it saves into took the whole game down** at that player's next save, because `prepare_quotes()` halts the MUD rather than truncate. `remote` refuses an over-long name on a player now and says so in the script log.

Six more that the sanitizer would never have caught:

- **`give`, `receive`, `get`, `drop`, `wear` and `consume` all fired at the top of their functions**, ahead of the checks that can still turn the action down. A receive trigger that paid out did so on gives that then failed for a dozen reasons, and the player kept the item as well - repeatable at will. Each runs at the last point before the state changes now, and give and receive run at each of the three places `perform_give()` actually hands the object over. Re-run live from a mortal client afterwards, because `remove_otrigger` excludes staff exactly as stock DG does: get, wear and remove each fire immediately before the game's own message, and a drop the game refuses fires nothing. Drop, give, receive and consume are still checked by reading only.
- **Load triggers never fired for anything a zone reset placed** - only `mload`, `oload` and `wload` called them. That is nearly everything, and the first thing a builder would have tried. `wizload` fires them too now, so one can be tested without waiting for a reset.
- **`act()`'s `vict_obj` is a `void *`** that `perform_act()` reads as whatever the format code asks for. The act-trigger hook cast it to `char_data *` regardless, so `put knife in bag` in a room with an act trigger on "puts" handed `char_script_id()` a container.
- **`idelete` dropped a live object's running script**, and with `script_id` back to zero the uid lookup table kept an entry pointing at memory about to be freed.
- **`dg_owner_purged` was a single global**, so a nested script purging its own owner made the outer script bail out.
- **`build_cmdlist()` split with `strtok()`**, which runs delimiters together, so every blank line a builder used to space a script out was lost on the next reload.

### Still true, and worth knowing

- **A script that kills still frees immediately.** `mdamage` goes through `damage()` and `raw_kill()` to `extract_char()`, so `mdamage %self% <enough>` frees the running script's owner mid-script. DG's deferral covers purge and junk; it cannot cover the game's own death path without changing when things die for everything else, and that is not a call I should make inside a port.
- **Two hooks still run ahead of a refusal**: a drop into a vehicle that turns out to be full, and removing hardened armour or a gyromount while fighting. Both leave the object where it was.

### Two more I would want your steer on

- **Greet, Greet-All and Enter do not turn a mover back on `return 0`.** The move has already happened by the time they run, so 0 only stops the triggers that would have run after. tbaMUD sends the character back where they came from. Doing that here means `char_from_room` / `char_to_room` from inside a trigger check, in a game with vehicles and elevators, so I would rather ask. The help entries describe the behaviour as it actually is.
- **`%actor.nuyen(n)%` and `%actor.karma(n)%` write the field directly.** `gain_nuyen()` and `gain_karma()` take arguments a script has no way to supply - a nuyen category, and whether reputation, limits and multipliers apply - so the port writes the raw field and nothing is logged. `physical` and `mental` are capped at the top but have no floor and do not call `update_pos()`, so a script can put someone below zero boxes without them falling over. What a script should be allowed to do to a character sheet is a policy question, not a porting one.

No world data is changed. The only non-`src/` files are an empty `lib/world/trg/index`, the three SQL files, and the two lines that load the new one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


luciensadi:
Thanks for this! It's going to take me some time to review properly, so this may sit for a bit, but I appreciate the code and am aiming to get it reviewed as quickly as I can.

